PR #8, opened 2018-12-23
Fixed stuff for targeting Android

ErikBjare:
I accidentally did my own migration to Rocket 0.4, then I fixed some things.

Explanation here: https://github.com/ActivityWatch/aw-android/issues/1#issuecomment-449656866


Cargo.toml
```diff
@@ -1,10 +1,13 @@
 [package]
 name = "aw_server"
 version = "0.1.0"
-authors = ["Johan Bjäreholt <johan@bjareho.lt>"]
+authors = ["Johan Bjäreholt <johan@bjareho.lt>", "Erik Bjäreholt <erik@bjareho.lt>"]
+
+[target.'cfg(target_os="android")'.dependencies]
```
johan-bjareholt: Put platform-specific stuff att bottom of Cargo.toml pls

src/dirs.rs
```diff
@@ -1,21 +1,48 @@
+#[cfg(not(target_os="android"))]
```
johan-bjareholt: It's possible to cfg the whole module, do that instead so you don't have to do return Err(());

src/lib.rs
```diff
@@ -26,3 +26,45 @@ pub mod datastore;
 pub mod query;
 pub mod endpoints;
```
johan-bjareholt: Would be cool if you could add a library feature named "http" or something so we can conditionally remove all rocket stuff at compile time.

Or maybe that will be hard since we use rocket_contrib Json,JsonValue on a lot of places? (i assume that rocket_contrib depends on rocket).

src/lib.rs
```diff
@@ -26,3 +26,45 @@ pub mod datastore;
...
 pub mod dirs;
+
+
+// Everything below from;
+//   https://mozilla.github.io/firefox-browser-architecture/experiments/2017-09-21-rust-on-android.html
+//
+// Should probably move it into module
```
johan-bjareholt: TODO

src/dirs.rs
ErikBjare: Yeah but then all functions that use dirs would break, and I believe some paths should be returned here on Android, I just don't know which yet.

Cargo.toml
ErikBjare: Fixed in https://github.com/ActivityWatch/aw-server-rust/pull/8/commits/c11fab3c22e51fc45358a19906ceb3359e389ecf

src/dirs.rs
```diff
@@ -1,21 +1,71 @@
...
+        dir.push("aw-server-rust");
+        return Ok(dir);
+    }
+
+    #[cfg(target_os="android")]
+    unsafe {
+         return match ANDROID_DATA_DIR {
+            Some(ref path) => Ok(path.to_path_buf()),
+            None => Err(())
+        }
```
ErikBjare: @johan-bjareholt Any better way to do this? The current way is having a static variable that should be set before calling it on Android. 

Maybe get the android_data_dir from Rust on every call? Not sure how though. (But could add a comment about it)

src/dirs.rs
```diff
@@ -1,21 +1,71 @@
...
+    {
+        let mut dir;
+        dir = appdirs::user_config_dir(Some("activitywatch"), None, false)?;
+        dir.push("aw-server-rust");
+        return Ok(dir);
+    }
+
+    #[cfg(target_os="android")]
+    {
+        return Err(());
```
ErikBjare: @johan-bjareholt Maybe I should return proper error messages?

src/lib.rs
ErikBjare: One day maybe, but right now I just want to get this done. Works fine regardless.

Cargo.toml
```diff
@@ -1,21 +1,23 @@
...
 mpsc_requests = "0.2"
 plex = "0.2.3"
 appdirs = "0.2.0"
 
-[dependencies.rocket_contrib]
-version = "*"
-default-features = false
-features = ["json"]
+[lib]
+crate-type = ["cdylib"]
```
johan-bjareholt: Should it really always be a cdylib? What are the consequences from this? What happens to the unittests?

src/dirs.rs
johan-bjareholt: I think a panic! would be fine here. It would be a programmer error to try and use this on android when it's not implemented for that platform.

src/dirs.rs
```diff
@@ -1,21 +1,71 @@
...
+static mut ANDROID_DATA_DIR: Option<PathBuf> = None;
+
 pub fn get_config_dir() -> Result<PathBuf, ()> {
-    let mut dir = appdirs::user_config_dir(Some("activitywatch"), None, false)?;
-    dir.push("aw-server-rust");
-    return Ok(dir);
+    #[cfg(not(target_os="android"))]
+    {
+        let mut dir;
+        dir = appdirs::user_config_dir(Some("activitywatch"), None, false)?;
```
johan-bjareholt: Cleaner with just straight up "mut dir = ...", but not a big deal

src/dirs.rs
johan-bjareholt: For now you could just hardcode it to "/data/data/org.activitywatch.activitywatch/".

The right thing to do I believe though is to figure out the android packagename and set it to "/data/data/packagename", but not sure how to actually do that from rust in a nice way but there seems to be possible in multiple ways, here's one.
https://stackoverflow.com/questions/42918762/how-to-get-app-package-name-or-applicationid-using-jni-android
But at that point, maybe it's easier to just pass it down from Java.

Also, appdirs-rs says "Android support should be coming soon" but that repo hasn't been updated since 2016.

Cargo.toml
ErikBjare: Honestly, I don't quite know, I just followed the Mozilla guide. Has no effect on the build of the executable nor tests afaik.

src/dirs.rs
ErikBjare: I'd rather not hardcore it, and can't be bothered to try and call the JNI from Rust right now, so this'll have to do.

That being said, I noticed they were calling the Android logger using the JNI, which I'd really like to be able to. Should look into that.

src/dirs.rs
ErikBjare: Ah, makes sense, will fix that.

ErikBjare:
Also added basic Travis, works nicely apart from having to comment away a broken test for `remove_bucket` in https://github.com/ActivityWatch/aw-server-rust/pull/8/commits/c8594f0083c307732bb8978cd1c643dfd8b69d7d.

ErikBjare:
Merging this now! :tada:

Issue #9, opened 2019-01-11
Flooding transform seems to be broken


ErikBjare:
I just took a glance at the flooding transform as @johan-bjareholt suggested in https://github.com/ActivityWatch/aw-core/pull/73 and I find several issues.

https://github.com/ActivityWatch/aw-server-rust/blob/b31fe9571f9f8c072ec0980879416ef9516e2928/src/transform/mod.rs#L61-L96

 - `gap` isn't calculated correctly (calculated as the difference between event starts, should be calculated as the difference between event1 end and event2 start)
 - Calculating gap correctly will lead to some events getting a negative gap, which in turn will lead to the same bug as the one fixed in https://github.com/ActivityWatch/aw-core/pull/73
 - Doesn't seem to prioritize flooding the larger event over the smaller event

No rush to fix, but the issue is there. Should be easier to rewrite now that the issues have been removed from aw-core-python.

johan-bjareholt:
> Doesn't seem to prioritize flooding the larger event over the smaller event

This seems like an odd assumption, wouldn't the best solution be to meet half-way?

Anyways, this would be impossible to do currently since the peeked e2 is not mutable, I need to solve that somehow.

ErikBjare:
It's not a big deal really, just how I formulated flooding when I made the spec. I made the assumption that the probability of the larger event actually occupying the gap is larger than the small event doing so.

Arguably just adds complexity with minimal value, but prevents a lot of small events from together becoming large after flooding.

johan-bjareholt:
Fixed in #10

PR #10, opened 2019-02-03
transform: Fix calculation of gap in flood

johan-bjareholt:
- [x] gap isn't calculated correctly (calculated as the difference between event starts, should be calculated as the difference between event1 end and event2 start)
- [x] Calculating gap correctly will lead to some events getting a negative gap, which in turn will lead to the same bug as the one fixed in ActivityWatch/aw-core#73
- [x] ~~Doesn't prioritize flooding the larger event over the smaller event~~ Now takes the middle between the events

Solves #9

### DIFF
--- a/src/transform/mod.rs
+++ b/src/transform/mod.rs
@@ -65,10 +65,19 @@ pub fn flood(events: Vec<Event>, pulsetime: chrono::Duration) -> Vec<Event> {
     let mut e1_iter = events_sorted.drain(..).peekable();
     let mut new_events = Vec::new();
     let mut drop_next = false;
+    let mut gap_prev : Option<chrono::Duration> = None;
     while let Some(mut e1) = e1_iter.next() {
         if drop_next {
             drop_next = false;
             continue;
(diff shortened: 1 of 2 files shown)


johan-bjareholt:
@ErikBjare Review please. Finished the last thing we talked about.

ErikBjare:
Taking the middle of two events isn't how we do it in aw-server-python? 

Taking the middle minimizes the absolute errors but letting the larger one flood the entire gap minimizes the relative errors. We should choose one and stick to it in both implementations.

Otherwise looks good, but I only glanced over it (we should really look into sharing an integration test suite with aw-server-python though, would make it a lot easier to ensure both implementations follow a common spec).

johan-bjareholt:
> Taking the middle minimizes the absolute errors but letting the larger one flood the entire gap minimizes the relative errors.

Personally I see less risk in taking the middle, but I have probably not thought of most use-cases. We need to discuss this and do a proper evaluation.

PR #18, opened 2019-02-17
Test with Travis on Android

ErikBjare:


johan-bjareholt:
Huh, that's pretty cool

ErikBjare:
@johan-bjareholt There's more to be done though, working on it.

ErikBjare:
It's not working (see the weird `-llog` issue) but I set the Android target as an allowed failure and did most of the work, so I'd suggest we merge this unless you know what might be up @johan-bjareholt.

johan-bjareholt:
`-llog` seems to be the logging library for the android NDK, so maybe we are somewhere missing to add a -L to the NDKs library dir or something?

johan-bjareholt:
A missing append to LD_LIBRARY_PATH in compile-android.sh?

ErikBjare:
@johan-bjareholt Good point, just realized I forgot to tell cargo where to find the NDK linker and ar. Removed the config locally, which reproduced the issue. Added the config file to the repo now, so hopefully that'll be all! :tada:

PR #19, opened 2019-02-17
Attempt at sharing database object between server and library interfaces

ErikBjare:


src/android/mod.rs
```diff
@@ -25,16 +25,19 @@ pub mod android {
...
     use datastore::Datastore;
     use models::{Event, Bucket};
+    use endpoints;
 
-    fn openDatastore() -> Datastore {
-        let db_dir = dirs::db_path().to_str().unwrap().to_string();
-        Datastore::new(db_dir)
-    }
+
+    static mut DATASTORE: Option<&Datastore> = None;
```
johan-bjareholt: I've seen this way of making Singletons with Options a few times before in a couple of gtk-rs projects, not a perfect solution but it's fine.

johan-bjareholt: Just realized though that those projects have them inside a atomic reference counter and a mutex which is not possible here.

johan-bjareholt: The best solution would probably be so DatabaseWorker is a singleton as the actual Database object is just a wrapper with a Requester from mpsc_requests which could be cloned without any issues.

johan-bjareholt:
I now have a prototype where the Database object is clone-able.
Would this solve the issue for you?

ErikBjare:
@johan-bjareholt That would work very nicely, yes!

johan-bjareholt:
Here's the PR #20

ErikBjare:
Closing, solved in #20

PR #20, opened 2019-02-20
Externalize datastore lock

johan-bjareholt:
Makes it possible to clone the Datastore producer object.
This will make it possible for the android code to both have the Web API running at the same time as we do other requests to the datastore. Will also probably make it easier to implement sync later on as that would probably run in its own thread and also need access to the datastore.

An example would be:
```
let datastore = datastore::Datastore::new_in_memory();
let state = endpoints::ServerState {
    datastore: Mutex::new(datastore.clone()),
    asset_path: PathBuf::from("aw-webui/dist"),
};
endpoints::rocket(state, None)
// Do other stuff with the original datastore object
```

### DIFF
--- a/src/android/mod.rs
+++ b/src/android/mod.rs
@@ -3,6 +3,7 @@
 
 use std::os::raw::{c_char};
 use std::ffi::{CString, CStr};
+use std::sync::Mutex;
 use dirs;
 
 mod logcat;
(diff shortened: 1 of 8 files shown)


johan-bjareholt:
These changes obviously break the android builds until fixed

ErikBjare:
Working perfectly, thanks!

PR #21, opened 2019-02-24
WIP: Create config

johan-bjareholt:
Need to think a few more things through to make it work in a nice manner

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ appdirs = "0.2.0"
 lazy_static = "1.2"
 log = "0.4"
 fern = { version = "0.5", features = ["colored"] }
+toml = "0.4"
 
 [lib]
 crate-type = ["lib", "cdylib"]
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,0 +1,83 @@
+use std::fs::File;
+use std::io::{Read, Write};
+use rocket::config::{Config, Environment, Limits};
+
+use dirs;
+
+#[derive(Serialize,Deserialize)]
+pub struct AWConfig {
+    #[serde(default = "default_address")]
+    address: String,
+    #[serde(default = "default_port")]
+    port: u16,
+    // TODO: add CORS origins once that is properly supported
+}
+
+impl Default for AWConfig {
+    fn default() -> AWConfig {
+        AWConfig {
+            address: default_address(),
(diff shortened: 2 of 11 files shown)


Cargo.toml
```diff
@@ -18,6 +18,7 @@ appdirs = "0.2.0"
 lazy_static = "1.2"
 log = "0.4"
 fern = { version = "0.5", features = ["colored"] }
+toml = "0.4"
```
ErikBjare: Yes! :heart:

src/config.rs
```diff
@@ -0,0 +1,83 @@
...
+pub fn get_config() -> AWConfig {
+    let mut config_path = dirs::get_config_dir().unwrap();
+    match Environment::active().expect("Failed to get current environment") {
+        Environment::Production => config_path.push("config.toml"),
+        Environment::Development => config_path.push("config-testing.toml"),
+        Environment::Staging => panic!("Staging environment not supported"),
+    };
+
+    if !config_path.is_file() {
+        debug!("Writing default config at {:?}", config_path);
```
ErikBjare: I don't think we should write a default configuration. The defaults should be loaded in memory, optionally overloaded by whatever is specified in the config file.

I know this is not how we do it in the Python code, but I've been thinking a while that we've been doing it wrong.

johan-bjareholt: Good point.
On the other hand it is nice to have all options visible in the config.
Not sure what the best option would be to be honest, both has its flaws.

ErikBjare: True, although the options should be documented anyway. 

Perhaps we could add the default options to the config, but commented out?

johan-bjareholt: Fixed

src/config.rs
```diff
@@ -0,0 +1,83 @@
+use std::fs::File;
+use std::io::{Read, Write};
+use rocket::config::{Config, Environment, Limits};
+
+use dirs;
+
+#[derive(Serialize,Deserialize)]
+pub struct AWConfig {
+    #[serde(default = "default_address")]
+    address: String,
```
johan-bjareholt: Maybe we should have a section for address and port though? What should we call it?

ErikBjare: I don't think it's needed tbh, works fine in the top-level.

PR #31, opened 2019-03-17
Use safe CORS configuration

### LINKED ISSUE #24: Safe CORS configuration
[rocket_cors now support regex in the origin configuration](https://github.com/lawliet89/rocket_cors/pull/62), we should use that to implement CORS safely. Maybe also add support for custom origin configurations in the config?

johan-bjareholt:
Will resolve #24 

Uses the rocket_cors 0.5.1-beta-1 discussed about here
https://github.com/lawliet89/rocket_cors/issues/31

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -7,10 +7,12 @@ use crate::dirs;
 #[derive(Serialize,Deserialize)]
 pub struct AWConfig {
     #[serde(default = "default_address")]
-    address: String,
+    pub address: String,
     #[serde(default = "default_port")]
-    port: u16,
-    // TODO: add CORS origins once that is properly supported
+    pub port: u16,
+    #[serde(skip, default = "is_testing")]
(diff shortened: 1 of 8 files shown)


src/config.rs
```diff
@@ -49,13 +52,20 @@ fn default_port() -> u16 {
...
         Environment::Staging => panic!("Staging environment not supported"),
-    };
+    }
+}
+
+pub fn get_config() -> AWConfig {
+    let mut config_path = dirs::get_config_dir().unwrap();
+    match is_testing() {
+        false => config_path.push("config.toml"),
+        true => config_path.push("config-testing.toml")
```
ErikBjare: Nice with separate configs for testing and production.

johan-bjareholt: This waas the same before as well, just started using the is_testing function now instead.

PR #33, opened 2019-03-19
datastore: Add 'data' field to buckets table

johan-bjareholt:
Database migration is pretty simple when using "ALTER TABLE"


johan-bjareholt:
@ErikBjare Yes, but let's keep migrations to aw-server-rust.

This only works for SQLite->SQLite migrations though

johan-bjareholt:
I have not added an API for the "data" object yet, but it's there when you do a GET request for all buckets. Merging once the tests passes.

Issue #32, opened 2019-03-18
datastore: Implement "last_updated" for buckets


johan-bjareholt:

ErikBjare:
Does this mean I'll finally get my `data` attribute for buckets as well? If so, we should probably come up with an attribute naming convention for metadata with a specific meaning (to the server) and metadata without well-defined meaning. Not sure if this should be the final distinction, but I'm thinking something along the lines of `__` (dunder) in Python and `$` in MongoDB queries.

Like so:

```js
data: {
  "$last_updated": "2019-01-01T12:00Z",
  "name": "A user-friendly name",
  "description": "Describes the content of the bucket in better detail",
}
```

Also, could we get "$last_modification" or something similar that only updates on data edits or event deletions? (i.e. not on simple heartbeat-type event extensions which merely append data to the bucket, in a sense). Would make sense for the caching method I had in mind.

>Will slow down the query of buckets quite significantly as the cache is not enough anymore

What do you mean by this, why would it slow it down?

johan-bjareholt:
> Does this mean I'll finally get my data  attribute for buckets as well? If so, we should probably come up with an  attribute naming convention for metadata with a specific meaning (to  the server) and metadata without well-defined meaning. Not sure if this  should be the final distinction, but I'm thinking something along the  lines of __ (dunder) in Python and $ in MongoDB queries.

No it doesn't unfortunately.

I don't think we should use prefixes but a sub-object "data" or "metadata" instead to make it very clear that all fields in there are not required.

But last_updated is different though as it's not really a set value, but just a computed value which is never saved. What to do about that though I'm not sure, but as I said I'm not a fan of undescriptive prefixes so maybe a "computed" object for something like that instead?.

> In the longer run, I think it would make sense to move things like hostname and client into the bucket data. What do you think?

I'd be happy to move "name" and "description" into that data/metadata section.
About moving "hostname" in there though would make it optional which I'm not really sure about, it's an important identifier for tracking what host the bucket is owned by. The fact that aw-watcher-web is not able to get the hostname is the bug, not the field itself (unfortunately an unfixable bug though). We could in the future when we have sync support maybe just let aw-server decide the hostname and break the support for the "remote aw-server" functionality.

> Also, could we get "$last_modification" or something similar that only updates on data edits or event deletions? (i.e. not on simple heartbeat-type event extensions which merely append data to the bucket, in a sense). Would make sense for the caching method I had in mind.

[...]

PR #36, opened 2019-03-21
datastore: Added bucket metadata and removed external ReplaceLast API

johan-bjareholt:
Solves #32 

The resulting data looks like this when doing a request over REST:

```
"metadata": {
    "first_starttime": "2018-01-01"
    "last_endtime": "2018-01-02"
}
```

If the bucket is empty:

```
"metadata": {
    "first_starttime": null
    "last_endtime": null
}
```

### DIFF
--- a/src/datastore/datastore.rs
+++ b/src/datastore/datastore.rs
@@ -14,6 +14,7 @@ use rusqlite::TransactionBehavior;
 use mpsc_requests;
 
 use crate::models::Bucket;
+use crate::models::BucketMetadata;
 use crate::models::Event;
 use crate::transform;
 
@@ -53,7 +54,6 @@ pub enum Commands {
     GetBucket(String),
     GetBuckets(),
     InsertEvents(String, Vec<Event>),
-    ReplaceLastEvent(String, Event),
     Heartbeat(String, Event, f64),
     GetEvents(String, Option<DateTime<Utc>>, Option<DateTime<Utc>>, Option<u64>),
     GetEventCount(String, Option<DateTime<Utc>>, Option<DateTime<Utc>>),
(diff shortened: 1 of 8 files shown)


src/datastore/datastore.rs
```diff
@@ -253,14 +244,31 @@ impl DatastoreWorker {
...
     fn get_stored_buckets(&mut self, conn: &Connection) -> Result <(), DatastoreError> {
-        let mut stmt = match conn.prepare("SELECT id, name, type, client, hostname, created, data FROM buckets") {
+        let mut stmt = match conn.prepare("
+            SELECT  buckets.id, buckets.name, buckets.type, buckets.client,
+                    buckets.hostname, buckets.created, buckets.data,
+                    min(events.starttime), max(events.endtime)
+            FROM buckets
+            INNER JOIN events ON buckets.id = events.bucketrow
+            GROUP BY buckets.id
+            ;") {
```
ErikBjare: Is this query still about as fast? (I'm kinda bad at predicting performance of SQL queries)

johan-bjareholt: Well, I'm no expert either, but these are my two cents: Firstly inner joins and group by are made to support very large data sets in real SQL servers and this should be negligible there, but if sqlite works differently I don't know. Secondly we only have a few buckets. Thirdly this query only runs once when the data store is initialized, after that the buckets are cached. Lastly I don't have any tests for this yet so I might have to modify this query if it doesn't work as expected. This PR need more tests.

johan-bjareholt: Another point is that starttime and endtime is indexed, so to get those values should be pretty much instant.

johan-bjareholt:
Splitted the large datastore test into multiple tests and created a proper test for first_starttime and last_endtime with reload of the database to make sure that it works both when cached and that the value is correct when reloading after a restart.

src/datastore/datastore.rs
```diff
@@ -269,6 +288,10 @@ impl DatastoreInstance {
                 hostname: row.get(4)?,
                 created: row.get(5)?,
                 data: serde_json::from_str(&data_str).unwrap(),
+                metadata: BucketMetadata {
+                    first_starttime: opt_first_starttime,
+                    last_endtime: opt_last_endtime,
```
ErikBjare: I don't really like the names `first_starttime` and `last_endtime`. Perhaps just `begins` and `ends`? Or a tuple (or ISO8601 formatted interval string) `interval` or `range`?

johan-bjareholt:
@ErikBjare There, finally fixed it. Will merge tomorrow when the tests have passed.

PR #37, opened 2019-04-04
Added debug symbols to release builds

ErikBjare:
I need this to be able to properly debug stuff in aw-android.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,9 +30,6 @@ log = "0.4"
 fern = { version = "0.5", features = ["colored"] }
 toml = "0.4"
 
-[profile.release]
-debug = true
-
 [target.'cfg(target_os="android")'.dependencies]
 jni = { version = "0.5", default-features = false }
 libc = "0.2"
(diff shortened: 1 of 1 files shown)


johan-bjareholt:
This bloats the library, can you verify the difference before and after? (It's probably not going to be a significant size difference, but best to make sure as I'm not certain).

Also, why only on release? Should either enable it on only debug or on both, only release makes no sense.

ErikBjare:
@johan-bjareholt It's enabled in debug builds by default. It might bloat the binary, but I really need a decent stack trace in my Android builds (and be able to read them in the Play Store crash reports) when it SIGABRT's for some unknown reason. Enabling debug symbols seems to be the only way to go about it. 

I will compare the size of the binaries before I merge. We could always disable it again in the future when the issues have been resolved.

ErikBjare:
Yeah... The size was 7M vs 62M... So I guess I'll just have to enable it some other way specifically for the Android beta builds (environment variables probably).

PR #45, opened 2019-05-12
Query improvements

### LINKED ISSUE #39: Code coverage in CI

### LINKED ISSUE #44: query: Implement concat function
Needed in the latest aw-webui version when querying all browser buckets at the same time

johan-bjareholt:
Additions:
- Boolean support
- Basic if/elif/else support
- == support
- Added concat function
- Added contains function
- Better error messages (line numbers during lexing, but not parsing)
- Event and Bucket models now use serde_json Map instead of Value as it should
- Coverage in travis with codecov.io

Issues:
- Since this is the first PR with coverage we will miss coverage report for the base branch so that will always fail

Fixes #44 
Fixes #39

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: required
 language: rust
 rust:
   - nightly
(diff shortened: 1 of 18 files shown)


ErikBjare:
Kinda lame that it needs to have `sudo: required` and custom apt installs, but oh well.

Weird that the Windows build times out, could be due to `sudo: required`?

johan-bjareholt:
The windows timeout could be because of this, I added a secret env variable in the travis project settings for codecov.io
https://travis-ci.community/t/windows-ci-not-starting/2057/5

ErikBjare:
@johan-bjareholt You should be able to put the secret env variable in the environment field for the specific build (in the Travis config, not in the Travis UI). That way it wouldn't be used in the Windows build in the first place (although it would obviously be preferred if they just fixed the issue upstream).

johan-bjareholt:
Just realized that you only needed the codecov.io token on private repositories so I removed it. We'll see if it works now...

johan-bjareholt:
@ErikBjare Works now, please review

ErikBjare:
Can I see a more complex query example using the new features (if/elif/else, ==)? 

Does this mean a further departure from the query2 in aw-server-python? If so, that should be documented somehow (or the new features at least clearly marked unstable).

johan-bjareholt:
> Can I see a more complex query example using the new features (if/elif/else, ==)? 

The tests has lots of examples on all of that with multiple variations.

>  I assume the if/elif/else syntax you've gone for would make query2 no longer a subset of Python syntax?

It has never really been a subset of python, it's just that the syntax has been so simple that it might look like python but it could've just as well have been javascript or some other scripting language which uses normal +-/*% operators, = for assigning variables without a type declaration, calls funtions with parentheses and supports lists and hashmaps with {} and [].

The reason for using {} instead of : and space indentation like in python is because that space indentation would be pretty annoying to parse and space indentation doesn't have any technical advantages.

> Does this mean a further departure from the query2 in aw-server-python? If so, that should be documented somehow (or the new features at least clearly marked unstable).

We're still backwards compatible. The day query2 in aw-server-python is dropped we can start using the new query features in the web-ui but for now we shouldn't.

johan-bjareholt:
@ErikBjare Can I merge this?

johan-bjareholt:
@ErikBjare Pls review

Issue #42, opened 2019-05-06
OPTIONS route not generated for endpoints


ErikBjare:
I get this error when running aw-server-rust on Android together with aw-watcher-web on Firefox for Android:

```
D/aw-server-rust: Error: No matching routes for OPTIONS /api/0/buckets/aw-watcher-web-firefox/heartbeat?pulsetime=80.
D/aw-server-rust: Warning: Responding with 404 Not Found catcher.
```

Flask autogenerates the `OPTIONS` route for endpoints. I had hoped Rocket would do the same, but apparently not. Do we really implement `OPTIONS` routes on every endpoint? Hasn't anyone else had this issue before? (A quick search suggested no, but could be mistaken)

johan-bjareholt:
rocket_cors should implement OPTIONS, maybe that's misconfigured somehow?

I have seen OPTIONS requests go through many times though, not sure how your scenario differs.

ErikBjare:
That's weird, I just restarted the extension to checked that the same thing happens when creating a bucket, and it does:

```
D/aw-server-rust: Warning: Responding with 404 Not Found catcher.
D/aw-server-rust: Error: No matching routes for OPTIONS /api/0/buckets/aw-watcher-web-firefox.
```

johan-bjareholt:
This is how it looks on my machine, the OPTIONS request gets a 204 and the HEARTBEAT afterwards a 200.

```
[2019-05-07][17:00:59][rocket::rocket][INFO] OPTIONS /api/0/buckets/aw-watcher-web-firefox/heartbeat?pulsetime=80:
[2019-05-07][17:00:59][_][INFO] Origin has a regex match? true
[2019-05-07][17:00:59][_][ERROR] No matching routes for OPTIONS /api/0/buckets/aw-watcher-web-firefox/heartbeat?pulsetime=80.
[2019-05-07][17:00:59][_][WARN] Responding with 404 Not Found catcher.
[2019-05-07][17:00:59][_][INFO] CORS Fairing: Turned missing route OPTIONS /api/0/buckets/aw-watcher-web-firefox/heartbeat?pulsetime=80 into an OPTIONS pre-flight request
[2019-05-07][17:00:59][_][INFO] Response succeeded.
```

johan-bjareholt:
I just realized that the webextension UUID is different on all of my computers even though all of them run on the same version of the webextension from the Mozilla add-ons store.

[...]

PR #47, opened 2019-05-21
endpoints: Allow all moz and chrome extensions to do CORS requests

johan-bjareholt:
Solves #42

### DIFF
--- a/src/endpoints/cors.rs
+++ b/src/endpoints/cors.rs
@@ -12,15 +12,10 @@ pub fn cors(config: &AWConfig) -> rocket_cors::Cors {
     if config.testing {
         allowed_exact_origins.push("http://127.0.0.1:27180".to_string());
     }
-    let mut allowed_regex_origins = vec![
-        "moz-extension://6b1794a0-5ae6-4443-aef9-7755717bb180".to_string(),
-        "moz-extension://02507908-8928-4ec3-86f5-2ae5dddf4c47".to_string(),  // The ID I got on Firefox for Android
-        "chrome-extension://nglaklhklhcoonedhgnpgddginnjdadi".to_string(),
+    let allowed_regex_origins = vec![
+        "moz-extension://.*".to_string(),
+        "chrome-extension://.*".to_string(),
     ];
-    if config.testing {
-        allowed_regex_origins.push("moz-extension://.*".to_string());
-        allowed_regex_origins.push("chrome-extension://.*".to_string());
-    }
 
     let allowed_origins = AllowedOrigins::some(&allowed_exact_origins, &allowed_regex_origins);
     let allowed_methods = vec![Method::Get, Method::Post, Method::Delete]


ErikBjare:
I don't like this, we should make sure the extension ID stays fixed, in both implementations. We do this for Thankful somehow, we should do the same for ActivityWatch.

Also, why isn't the Chrome extension included here?

https://github.com/ActivityWatch/aw-server/blob/c01ffd939b718c55152ffb81da55217ab8d799e6/aw_server/server.py#L71-L84

src/endpoints/cors.rs
```diff
@@ -12,15 +12,10 @@ pub fn cors(config: &AWConfig) -> rocket_cors::Cors {
     if config.testing {
         allowed_exact_origins.push("http://127.0.0.1:27180".to_string());
     }
-    let mut allowed_regex_origins = vec![
-        "moz-extension://6b1794a0-5ae6-4443-aef9-7755717bb180".to_string(),
-        "moz-extension://02507908-8928-4ec3-86f5-2ae5dddf4c47".to_string(),  // The ID I got on Firefox for Android
```
ErikBjare: If my permanent ID fix works, this Firefox line should no longer be needed.

johan-bjareholt:
> Also, why isn't the Chrome extension included here? Is it even required for Chrome?

What do you mean it's not included? It's there in both on master and in this commit?

> Edit: I think I fixed it here ActivityWatch/aw-watcher-web#36

It refers to gecko, is some similar solution available for chrome or do you know if it is the same on all devices there?

ErikBjare:
>What do you mean it's not included? It's there in both on master and in this commit?

I can't be blind, I see `cors_origins.append("moz-extension://*")` and `cors_origins.append("http://127.0.0.1:27180/*")` if testing but nothing else. There is **no** `cors_origins.append("chrome-extension://*")`.

>It refers to gecko, is some similar solution available for chrome or do you know if it is the same on all devices there?

It should be the same for all devices there, the Chrome Webstore assigns you a permanent one: https://stackoverflow.com/questions/8946325/chrome-extension-id-how-to-find-it

johan-bjareholt:
> I can't be blind, I see cors_origins.append("moz-extension://*") and cors_origins.append("http://127.0.0.1:27180/*") if testing but nothing else. There is no cors_origins.append("chrome-extension://*").

That's odd, this is what I see

Master:

    let mut allowed_regex_origins = vec![
        "moz-extension://6b1794a0-5ae6-4443-aef9-7755717bb180".to_string(),
        "moz-extension://02507908-8928-4ec3-86f5-2ae5dddf4c47".to_string(),  // The ID I got on Firefox for Android
        "chrome-extension://nglaklhklhcoonedhgnpgddginnjdadi".to_string(),
    ];
    if config.testing {
        allowed_regex_origins.push("moz-extension://.*".to_string());
        allowed_regex_origins.push("chrome-extension://.*".to_string());
    }

This PR:

    let allowed_regex_origins = vec![
        "moz-extension://.*".to_string(),
        "chrome-extension://.*".to_string(),
    ];

> It should be the same for all devices there, the Chrome Webstore assigns you a permanent one: https://stackoverflow.com/questions/8946325/chrome-extension-id-how-to-find-it

Awesome, let's merge that PR then and I'll abandon this and then we'll add the final firefox id in aw-server-rust in a seperate PR or just push it straight to master.

ErikBjare:
You were looking in aw-server-rust for the Chrome CORS origin. I was referring and linking to aw-server-python where it isn't needed for some reason...

PR #48, opened 2019-05-21
Make heartbeat requests respond with a merged heartbeat

johan-bjareholt:
aw-client-js expects to get a event body as a response when the request is 200 OK, aw-server-rust now responds with the same data as aw-server-python

### DIFF
--- a/tests/api.rs
+++ b/tests/api.rs
@@ -168,6 +168,7 @@ mod api_tests {
             .dispatch();
         debug!("{:?}", res.body_string());
         assert_eq!(res.status(), rocket::http::Status::Ok);
+        assert_eq!(res.body_string().unwrap(), r#"{"id":null,"timestamp":"2018-01-01T01:01:01Z","duration":2.0,"data":{}}"#);
 
         // Get heartbeat event
         res = client.get("/api/0/buckets/id/events")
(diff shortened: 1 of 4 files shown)


tests/api.rs
```diff
@@ -168,6 +168,7 @@ mod api_tests {
             .dispatch();
         debug!("{:?}", res.body_string());
         assert_eq!(res.status(), rocket::http::Status::Ok);
+        assert_eq!(res.body_string().unwrap(), r#"{"id":null,"timestamp":"2018-01-01T01:01:01Z","duration":2.0,"data":{}}"#);
```
ErikBjare: Why is `"id": null`? That's not what aw-server does is it?

johan-bjareholt: Because id is not known until the SQLite transaction is performed. The peewee storage can know the id because it actually does the write directly, but that is also why it's so incredibly slow. I haven't checked but I'm pretty certain that the sqlite storage doesn't support that either because of that reason.

ErikBjare: Oh, I was not aware of this.

johan-bjareholt:
@ErikBjare Can I merge this?

johan-bjareholt:
I fixed the android compilation issue

@ErikBjare Can I merge this?

johan-bjareholt:
Okay, now it's failing because of outdated dependencies which I fixed in #45, can I still merge this?

PR #50, opened 2019-05-24
Added x86_64 build target on android

ErikBjare:
First step towards fixing https://github.com/ActivityWatch/aw-android/issues/20

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -9,3 +9,7 @@ linker = './NDK/arm/bin/arm-linux-androideabi-clang'
 [target.i686-linux-android]
 ar = './NDK/x86/bin/i686-linux-android-ar'
 linker = './NDK/x86/bin/i686-linux-android-clang'
+
+[target.x86_64-linux-android]
+ar = './NDK/x86_64/bin/x86_64-linux-android-ar'
+linker = './NDK/x86_64/bin/x86_64-linux-android-clang'
--- a/compile-android.sh
+++ b/compile-android.sh
@@ -8,13 +8,19 @@ if [ -z "$ANDROID_NDK_HOME" ]; then
     exit 1
 fi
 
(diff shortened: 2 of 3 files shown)


compile-android.sh
```diff
@@ -8,13 +8,19 @@ if [ -z "$ANDROID_NDK_HOME" ]; then
...
+else
+    echo "Building in debug mode... (fast)"
+fi
 
 ORIG_PATH="$PATH"
 
+# Ring won't build in the below target for whatever reason
+# 'arm armv7-linux-androideabi' \
 for archtargetstr in \
     'x86 i686-linux-android' \
```
johan-bjareholt: Just remove 32-bit support completely, it just slows down testing time. I have a 10 year old laptop and even that has a 64-bit CPU. The Core 2 series were the first mainstream 64-bit Intel CPUs and they have been available for 13 years now!

.cargo/config
```diff
@@ -9,3 +9,7 @@ linker = './NDK/arm/bin/arm-linux-androideabi-clang'
 [target.i686-linux-android]
```
johan-bjareholt: Remove this as well

compile-android.sh
ErikBjare: I don't know why, but Google seems to suggest supporting 32-bit x86 is normal: https://developer.android.com/distribute/best-practices/develop/64-bit

Could check the device support matrix and see if there actually are any 32-bit x86 devices out there that don't support x86_64, but until then I'll keep it.

johan-bjareholt: Might be because of those mobile Intel mobile SoC for smartphones which Intel made one generation of and failed spectacularly in 2014, they were 32-bit. Since then they probably just want x86_64 APKs because of Chromebooks now supporting Android apps.

PR #52, opened 2019-08-13
Bucket default data fix

ErikBjare:
Fixes https://github.com/ActivityWatch/aw-android/issues/25

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -47,7 +47,7 @@ fn default_address() -> String {
 fn default_port() -> u16 {
     match Environment::active().expect("Failed to get current environment") {
         Environment::Production => 5600,
-        Environment::Development => 5667,
+        Environment::Development => 5666,
         Environment::Staging => panic!("Staging environment not supported"),
     }
 }
--- a/src/datastore/datastore.rs
+++ b/src/datastore/datastore.rs
@@ -100,6 +100,10 @@ fn _create_tables(conn: &Connection) {
     if version < 2 {
         _migrate_v1_to_v2(conn);
     }
+
+    if version < 3 {
+        _migrate_v2_to_v3(conn);
+    }
 }
 
 fn _migrate_v0_to_v1(conn: &Connection) {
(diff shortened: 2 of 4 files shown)


src/datastore/datastore.rs
```diff
@@ -261,7 +261,12 @@ impl DatastoreInstance {
             Err(err) => return Err(DatastoreError::InternalError(format!("Failed to prepare get_stored_buckets SQL statement: {}", err.to_string())))
         };
         let buckets = match stmt.query_map(&[] as &[&dyn ToSql], |row| {
-            let data_str : String = row.get(6)?;
+            // If data column is not set (possible on old installations), use an empty map as default
```
johan-bjareholt: This is not how it should be solved as it creates ugly exceptions for legacy code.

Higher up in the _create_tables function you can see this:

        conn.execute("ALTER TABLE buckets ADD COLUMN data TEXT default '{}';", &[] as &[&dyn ToSql])

To me it seems like this should work, but apparently it does not? That's what should be fixed.

ErikBjare: Huh, that's weird. Works fine when I try the equivalent commands on in an sqlite3 REPL.

```
sqlite> create table buckets (id INTEGER PRIMARY KEY, name TEXT);
sqlite> .table
buckets
sqlite> insert into buckets (name) values ("test");
sqlite> select * from buckets;
1|test
sqlite> ALTER TABLE buckets ADD COLUMN data TEXT default '{}';
sqlite> select * from buckets;
1|test|{}
sqlite> PRAGMA user_version;
0
sqlite> PRAGMA user_version = 2;
sqlite> PRAGMA user_version;
2
```

johan-bjareholt: Well, I don't know how to reproduce so I can't really dig into it. It's the classical case of "works fine on my machine".

ErikBjare: Well, after a lot of testing I've had no trouble reproducing. It happens every time on Android, **even on a fresh install**, but never when run on my laptop.

I suggest we merge this and call it a day, I really have no idea where else to go from here, unless you have an idea.

johan-bjareholt: The thing is, the code doesn't even allow "data" to contain null since the beginning of June when I merged my large PR, only a Map<String, Value> so it doesn't make any sense.

https://github.com/ActivityWatch/aw-server-rust/blob/master/src/models/bucket.rs#L21

I really want to avoid having to maintain legacy code like this.

ErikBjare: Alright, apparently Android doesn't clear app data on an uninstall. After clearing data and running again it did not occur.

ErikBjare: >The thing is, the code doesn't even allow "data" to contain null since the beginning of June when I merged my large PR, only a Map<String, Value> so it doesn't make any sense.

It's not `Bucket.data` that contains `null`, it's the sqlite field that contains the string `"null"`.

ErikBjare: People seem to suggest one should always use `NOT NULL` with `TEXT DEFAULT`: https://stackoverflow.com/a/6599261/965332

Why it doesn't respect the `DEFAULT` specifically on Android I don't know (could be due to how it's called from the code, leading to `NULL`s appearing) but it would obviously solve the problem in our case.

Now the question is: How do we migrate from a `TEXT DEFAULT` data column to a `TEXT NOT NULL DEFAULT` data column?

johan-bjareholt: 3. During migration:
 - do a select on data for all rows and save them in memory
 - delete the data column
 - create it again but now with NOT NULL
 - Restore the saved data fields, If the old data was NULL don't set it as that will make it '{}' which should be correct

ErikBjare: You cannot delete a column.

ErikBjare: I added a migration that sort of solved it: https://github.com/ActivityWatch/aw-server-rust/pull/52/commits/2b55a97549429c4b570b28bd8d05396a22d8046e

I've already spent way too much time (5h+) on what is practically nitpicking, so if this works I will merge it.

src/config.rs
```diff
@@ -47,7 +47,7 @@ fn default_address() -> String {
 fn default_port() -> u16 {
     match Environment::active().expect("Failed to get current environment") {
         Environment::Production => 5600,
-        Environment::Development => 5666,
+        Environment::Development => 5667,
```
johan-bjareholt: Revert when you are done

ErikBjare:
>This way only "legacy" versions of the database will have an extra "data_deprecated" field

Having different columns on different databases depending on when they were first installed is bad practice and only adds unnecessary complexity. It only gives yet another edge case to consider, so I won't.

johan-bjareholt:
> Having different columns on different databases depending on when they were first installed is bad practice (a SELECT * FROM <TABLE> would give different results, for example) and only adds unnecessary complexity. It only gives yet another edge case to consider, so I won't.

If we actually cared about that table I would agree, but we are naming it because we can't delete it. Now we will instead have a trash column which no one uses on all databases, even for future databases which don't need a migration which is stupid.

ErikBjare:
That does not justify practically having two schemas for the same version. Doing so would create a whole host of hard to debug/replicate errors (like the `SELECT *` example) and makes writing future migrations harder. It should be obvious, but two databases with the same version should always have an identical schema, otherwise they are not actually the same version.

>If we actually cared about that column I would agree, but we are renaming it because we can't delete it.

We can delete it, it just requires a more complicated process of moving every row over to a new table. We can clean it up whenever we want to by creating another, more involved, migration. If that extra column really is that annoying to you, feel free to write such a migration.

johan-bjareholt:
@ErikBjare Fair enough, we'll let this be for now

PR #54, opened 2019-08-17
Preparations for packaging aw-server-rust

johan-bjareholt:
I have made some changes so it becomes easier to package aw-server-rust both for a self-contained install in a folder (windows/mac/linux) with the package.sh script and a nicer search path for the webui assets.

With the change of the search path I also was able to make the aw-server-rust.service a bit more simple.

Currently I have decided this:
- aw-server-rust is the name of the binary
- aw_server_rust is the folder for all assets
- aw_server_rust/static is the folder for the webui assets

Currently there are only webui assets, but I want to be prepared for the future.


ErikBjare:
Why not use `/opt`? Seems like the most common thing for cross-platform applications like ours.

My `/usr/local/bin` is empty, and my `/usr/local/share` is almost empty:

```
➤➤➤ la /usr/local/share/
total 16K
drwxr-xr-x  3 root root 4.0K Jun  7 14:59 ./
drwxr-xr-x 11 root root 4.0K Jan  4  2018 ../
drwxr-xr-x  3 root root 4.0K Dec 27  2018 .cache/
lrwxrwxrwx  1 root root    6 May 23 16:18 man -> ../man/
-rw-r--r--  1 root root  116 Jul 13 23:18 .yarnrc
```

Some discussion: https://unix.stackexchange.com/questions/11544/what-is-the-difference-between-opt-and-usr-local

johan-bjareholt:
@ErikBjare Your /usr/local/bin is empty because you rarely compile and install programs by yourself which are not cross-platform.

I chose `/usr/local` because `/usr/local/bin` is by default in my `PATH` so I don't have to make any symlinks. If we made a symlink though it doesn't really matter. If we do make a symlink however it should be in `/usr/local/bin` rather than `/usr/bin` as that's supposed to be reserved for the package manager.

What I'll do is to do like you said but with the symlink in /usr/local/bin instead. I will also replace the .sh scripts with a Makefile.

alerque:
An app like this setting up a build / install routine has no business using `/opt` for anything. Most platforms have a variable (or `pkg-config` can report) for where software should go. Usually this will be `/usr/bin` for the system default (used by the system package manager) or `/usr/local/bin` for software installed to the system but not packaged using the system's package tools. Not not only distro managed packages will use `/usr/bin` but anything compiled and installed through the package manager even if it is done my users on the local system.

johan-bjareholt:
> An app like this setting up a build / install routine has no business using /opt for anything. Most platforms have a variable (or pkg-config can report) for where software should go. Usually this will be /usr/bin for the system default (used by the system package manager) or /usr/local/bin  for software installed to the system but not packaged using the  system's package tools. Not not only distro managed packages will use /usr/bin but anything compiled and installed through the package manager even if it is done my users on the local system.

This solution should be flexible to solve both situations, the steps are split into package and install where install depends on package.

The binary is self-contained and does not need any dependencies, so there's not really any advantage over `/opt` over `/usr/local/bin`. If someone still prefers using `/usr/local/bin` it is possible to simply only do the package step and then make your own installation step which moves the executable into `/usr/local/bin`.

The only issue with this would be how aw-server-rust looks for its aw-webui assets, either you have to put the assets inside `/usr/local/bin` (not nice) or change your cwd to where you copied the assets before starting aw-server-rust (not nice either, but an acceptable workaround for now).

johan-bjareholt:
> The only issue with this would be how aw-server-rust looks for its aw-webui assets, either you have to put the assets inside /usr/local/bin (not nice) or change your cwd to where you copied the assets before starting aw-server-rust (not nice either, but an acceptable workaround for now).

Now that I think of it, this would be solvable by iterating over XDG_DATA_DIRS and see if the aw_server_rust folder exists in any of them.

ErikBjare:
>Now that I think of it, this would be solvable by iterating over XDG_DATA_DIRS and see if the **aw_server_rust** folder exists in any of them.

@johan-bjareholt Can you please keep kebab-case for all uses of aw-server-rust when possible?

Makefile
```diff
@@ -0,0 +1,30 @@
+.PHONY: all aw-server-rust aw-webui install package user-symlink
```
alerque:
```suggestion
DESTDIR := /usr/local

.PHONY: all aw-server-rust aw-webui install package user-symlink
```

Makefile
```diff
@@ -0,0 +1,30 @@
...
+	# Copy binary
+	cp target/release/aw-server-rust target/package
+	# Copy webui assets
+	mkdir -p target/package/aw_server_rust
+	cp -rf aw-webui/dist target/package/aw_server_rust/static
+	# Copy service file
+	cp -f aw-server-rust.service target/package/aw_server_rust
+
+install: target/package
+	mkdir -p /opt/activitywatch
```
alerque:
```suggestion
	mkdir -p $(DESTDIR)/share/activitywatch
```

Makefile
```diff
@@ -0,0 +1,30 @@
...
+	cp -rf target/package/* /opt/activitywatch
```
alerque:
```suggestion
	cp -rf target/package/* $(DESTDIR)/share/activitywatch
```

Makefile
```diff
@@ -0,0 +1,30 @@
...
+	ln -s /opt/activitywatch/aw-server-rust /usr/local/bin/aw-server-rust
```
alerque:
```suggestion
	install -m755 target/package/aw-server-rust $(DESTDIR)/bin/aw-server-rust
```

Makefile
```diff
@@ -0,0 +1,30 @@
...
+
+service-symlink:
+	ln -sf /opt/activitywatch/aw_server_rust/aw-server-rust.service ~/.config/systemd/user/
```
alerque:
```suggestion
	ln -sf $(DESTDIR)/share/activitywatch/aw_server_rust/aw-server-rust.service ~/.config/systemd/user/
```

aw-server-rust.service
```diff
@@ -17,8 +16,7 @@
 
 [Service]
 Type=simple
-ExecStart=/usr/local/bin/aw-server-rust
-WorkingDirectory=/usr/local/share/aw-server-rust/
+ExecStart=/opt/activitywatch/aw-server-rust
```
alerque: This file needs to be built too at package time and the makefile's variable for DESTDIR needs to be used for the exec path.

johan-bjareholt:
@alerque I added the option to choose if they want it installed in `/usr` or in `/opt` with the targets `install-usr` and `install-opt`.

The flaw in this solution is that the path in the service file is no longer static, but just aw-server-rust. But as long as users has /usr/local/bin in their PATH this shouldn't be an issue.

alerque:
Why have two separate targets? This not only restricts the possible outcomes but adds a learning curve to the install process as opposed to being able to use muscle memory. Is there some reason you cannot use the typical variables `make install DESTDIR=/path PREFIX=/usr`?

Keep in mind most packagers need to prefix this with something besides the system root for the purpose of bundling up the package, so even on systems that will install to `/usr` will need the installer to install to `/path/to/packagers/sandbox/aw-server-rust/root/usr` or something like that.

johan-bjareholt:
@alerque 

> Why have two separate targets?

Well, to be fair the reason why I'm making two separate targets is because using `/opt` is easier as the folder is self-contained rather than spread out just as the bundled version.  We don't have anyone packaging activitywatch anyway so I see the usr as a bonus. The day someone actually starts using the usr target we can continue improving it.

> This not only restricts the possible outcomes but adds a learning curve to the install process as opposed to being able to use muscle memory.

You make a good point, but to me that mostly means that I'm more interested in only doing the /opt solution as that means less maintenance for us as it works the same way as the bundled version.

Still though, just running "make" and "make install" should work fine as install is aliased to install-opt

> Is there some reason you cannot use the typical variables make install DESTDIR=/path PREFIX=/usr?
>
> Keep in mind most packagers need to prefix this with something besides the system root for the purpose of bundling up the package, so even on systems that will install to /usr will need the installer to install to /path/to/packagers/sandbox/aw-server-rust/root/usr or something like that.

I think I fixed the DESTDIR and PREFIX issue now in the latest commit, works for me but some verification from someone else would be nice.

alerque:
> Well, to be fair the reason why I'm making two separate targets is because using `/opt` is easier as the folder is self-contained rather than spread out just as the bundled version.

Easier isn't better. In this case what you're trying to push on people is _harder_ for packagers, and packagers are who is going to make this easy for end users, and that's what should take priority here.

And it's not really _harder_ to maintain for you, you just need to learn to code the makefile right in the first place. Once done it will require no extra work from you.

> We don't have anyone packaging activitywatch anyway so I see the usr as a bonus. The day someone actually starts using the usr target we can continue improving it.

I am someone who is trying to package this for you. I've written, contributed to, and maintain hundreds of packages on several distros. Being hard to install is one of the biggest barriers to adoption an app can face, and being hard work with by insisting on non-standard procedures is a very real barrier stopping people from putting together packages.

> You make a good point, but to me that mostly means that I'm more interested in only doing the /opt solution as that means less maintenance for us as it works the same way as the bundled version.

I just reviewed the current iteration of the Makefile in this PR and have come to the conclusion that you need to drop the `/opt` thing entirely. You're creating a maintenance nightmare, not solving one. You _need_ your software to work installed per system norms (many distros will require this), so long term making that work isn't optional. Having a second option to install to a non-standard location is just adding complexity.

> I think I fixed the DESTDIR and PREFIX issue now in the latest commit, works for me but some verification from someone else would be nice.

At the very least the variables need to be overridable (`?=` not `:=`). I'm looking through the rest now.

alerque:
I started fixing some things in [this branch](https://github.com/alerque/aw-server-rust/commits/make-sanity) but ran into a problem testing the actual install targets because the package itself doesn't build for me:

On a totally clean checkout on running `make` it looks like cargo goes out and fetches 209 dependencies, then tries to compile. Compiling runs into an error and re-running `make` shortcuts pretty much straight to the error:

```
❯❯❯ make
cargo build --release
   Compiling devise_core v0.2.0
   Compiling pear_codegen v0.1.2
   Compiling plex v0.2.5
   Compiling multipart v0.16.1
error[E0554]: #![feature] may not be used on the stable release channel
 --> /home/caleb/.cargo/registry/src/github.com-1ecc6299db9ec823/pear_codegen-0.1.2/src/lib.rs:1:1
  |
1 | #![feature(crate_visibility_modifier)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0554]: #![feature] may not be used on the stable release channel
 --> /home/caleb/.cargo/registry/src/github.com-1ecc6299db9ec823/pear_codegen-0.1.2/src/lib.rs:2:1
  |
2 | #![feature(proc_macro_diagnostic, proc_macro_span)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0554]: #![feature] may not be used on the stable release channel
 --> /home/caleb/.cargo/registry/src/github.com-1ecc6299db9ec823/devise_core-0.2.0/src/lib.rs:1:1
  |
1 | #![feature(proc_macro_diagnostic, proc_macro_span)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0554]: #![feature] may not be used on the stable release channel
 --> /home/caleb/.cargo/registry/src/github.com-1ecc6299db9ec823/devise_core-0.2.0/src/lib.rs:2:1
  |
2 | #![feature(crate_visibility_modifier)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0554]: #![feature] may not be used on the stable release channel
 --> /home/caleb/.cargo/registry/src/github.com-1ecc6299db9ec823/devise_core-0.2.0/src/lib.rs:3:1
  |
3 | #![feature(concat_idents)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0554]: #![feature] may not be used on the stable release channel
 --> /home/caleb/.cargo/registry/src/github.com-1ecc6299db9ec823/plex-0.2.5/src/lib.rs:2:1
  |
2 | #![feature(proc_macro_diagnostic)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0554]: #![feature] may not be used on the stable release channel
 --> /home/caleb/.cargo/registry/src/github.com-1ecc6299db9ec823/plex-0.2.5/src/lib.rs:3:1
  |
3 | #![feature(proc_macro_span)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0554`.
The following warnings were emitted during compilation:

warning: Pear was unable to check rustc compatibility.
warning: Build may fail due to incompatible rustc version.

error: Could not compile `pear_codegen`.
warning: build failed, waiting for other jobs to finish...
error: aborting due to 3 previous errors

For more information about this error, try `rustc --explain E0554`.
error: Could not compile `devise_core`.
warning: build failed, waiting for other jobs to finish...
error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0554`.
error: Could not compile `plex`.
warning: build failed, waiting for other jobs to finish...
error: build failed
Makefile:11: *** [aw-server-rust] Error 101

#0  aw-server-rust at /home/caleb/projects/aw-server-rust/Makefile:11
#1  all at /home/caleb/projects/aw-server-rust/Makefile:3
```

I have rust 1.37. Is there really something in nightly that this project depends on? How soon is whatever feature that is going to land? Because packaging for distros is going to be hard as long as you're dependent on other un-released software.

ErikBjare:
@alerque That's just the state of the Rust crate ecosystem as of 2019. A lot of crates depend on nightly, including some of our dependencies.

This isn't really a problem since rust nightly is ridiculously stable, just get the nightly and enable it as default with `rustup` before you run make.

johan-bjareholt:
> We need to ship a single-folder bundle equivalent to the /opt solution for Windows and macOS as that's how things are expected to be packaged there.

While this is true, I actually think I've changed my mind. The "package" step should still do everything needed for macOS and windows without the "install" step, just using that folder should be good enough.

Will do that later this weekend if it's alright to you @ErikBjare ?

ErikBjare:
Right, sounds good to me.

src/main.rs
```diff
@@ -48,10 +47,95 @@ fn main() {
...
-        asset_path: PathBuf::from("aw-webui").join("dist"),
+        asset_path: asset_path,
     };
 
     endpoints::build_rocket(server_state, &config).launch();
 }
+
+use std::path::PathBuf;
+
+// The appdirs implementation of site_data_dir is broken on computers which has flatpak installed
```
johan-bjareholt: Found this issue when I ran on my laptop.
Kind of a shame that I had to add this, but it works.

johan-bjareholt:
> @alerque That's just the state of the Rust crate ecosystem as of 2019. A lot of crates depend on nightly, including some of our dependencies (as you noticed).

@ErikBjare "A lot" is a bit excessive, most are on Rust stable. The only package which uses nightly in aw-server-rust as far as I'm aware is Rocket which should hopefully support stable in a year or two once a few "unstable" macros have landed https://github.com/SergioBenitez/Rocket/issues/19

PR #57, opened 2019-08-20
datastore: Make it possible to replace events

### LINKED ISSUE #35: Stopwatch not working

johan-bjareholt:
- Sets the default duration to 0 if not specified
- Makes it possible to replace events
- insert_events now returns the events with an assigned id

### DIFF
--- a/tests/datastore.rs
+++ b/tests/datastore.rs
@@ -181,6 +181,7 @@ mod datastore_tests {
         assert_eq!(fetched_events[0].timestamp, e1.timestamp);
         assert_eq!(fetched_events[0].duration, e1.duration);
         assert_eq!(fetched_events[0].data, e1.data);
+        let e1 = &fetched_events[0];
 
         // Heartbeat match
         ds.heartbeat(&bucket.id, e2.clone(), 10.0).unwrap();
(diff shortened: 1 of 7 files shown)


tests/datastore.rs
```diff
@@ -183,6 +183,33 @@ mod datastore_tests {
...
+            duration: Duration::seconds(0),
+            data: json_map!{"key": json!("value")}
+        };
+        ds.insert_events(&bucket.id, &[e1.clone()]).unwrap();
+
+        let mut fetched_events = ds.get_events(&bucket.id, None, None, None).unwrap();
+        assert_eq!(fetched_events.len(), 1);
+        let e2 = fetched_events.pop().unwrap();
+
+        ds.insert_events(&bucket.id, &[e2.clone()]).unwrap();
```
ErikBjare: You're not really modifying the event though?

Maybe try changing the duration and the data field?

johan-bjareholt:
I cannot apparently deny my own pull request, but it actually seems like it's possible to get the lastrowid before a transaction is commited.

If it works I will rewrite this change instead to always return a list of all events with a id on all of them (looks possible, but not sure until I have tested).

johan-bjareholt:
A couple of clones that needs to be fixed and some more testing, it is verified that it works though!

johan-bjareholt:
Fixed, waiting for travis to finish then I'm probably merging

PR #58, opened 2019-08-22
fixed hostname and version fields in /api/0/info

ErikBjare:


Cargo.toml
```diff
@@ -29,6 +29,7 @@ lazy_static = "1.2"
 log = "0.4"
 fern = { version = "0.5", features = ["colored"] }
 toml = "0.5"
+gethostname = "0.2"
```
johan-bjareholt: Only depends on libc and winapi, nice!

src/endpoints/mod.rs
```diff
@@ -71,9 +72,12 @@ fn server_info() -> JsonValue {
         testing = false;
     }
 
+    let hostname = gethostname().into_string().unwrap_or("unknown".to_string());
+    const VERSION: &'static str = env!("CARGO_PKG_VERSION");
```
johan-bjareholt: Wait, does this really work when not running with "cargo run"?

ErikBjare: The following worked:

```
$ cargo build
...
$ ./target/debug/aw-server-rust
...
$ curl localhost:5667/api/0/info
{"hostname":"erb-laptop2-arch","testing":true,"version":"aw-server-rust v0.1.0"}
```

ErikBjare: Added a fallback, see: https://stackoverflow.com/questions/27840394/how-can-a-rust-program-access-metadata-from-its-cargo-package

ErikBjare: Apparently `env!` fetches the variable at compile-time: https://doc.rust-lang.org/std/macro.env.html

PR #62, opened 2019-08-28
Syncing

ErikBjare:
Started working on syncing


ErikBjare:
@johan-bjareholt Could I get some comments on this?

src/sync.rs
```diff
@@ -0,0 +1,159 @@
+// Basic syncing for ActivityWatch
```
johan-bjareholt: Use triple /// instead to make actual rustdoc comments instead of just code comments

src/sync.rs
```diff
@@ -0,0 +1,159 @@
...
+        let buckets_to = ds_to.get_buckets().unwrap();
+        let new_id = format!("{}-synced", bucket.id);
+        if !buckets_to.contains_key(new_id.as_str()) {
+            let mut bucket_new = bucket.clone();
+            bucket_new.id = new_id.clone();
+            ds_to.create_bucket(&bucket_new).unwrap();
+        }
+
+        // Sync events
+        // FIXME: Events are not being saved, does the datastore worker need more time before exit?
```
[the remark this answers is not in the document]
ErikBjare: I fixed it by simply moving the references to the datastores out of the main function and then leaving a `sleep` in main to give it some time to commit (is this necessary though?).

src/datastore/datastore.rs
```diff
@@ -677,11 +679,13 @@ impl Datastore {
 
     fn _new_internal(method: DatastoreMethod) -> Self {
         let (responder, requester) = crossbeam_requests::channel::<Commands, Result<Responses, DatastoreError>>();
+        let method_clone = method.clone();
         let _thread = thread::spawn(move || {
             let mut di = DatastoreWorker::new(responder);
-            di.work_loop(method);
+            di.work_loop(method_clone);
```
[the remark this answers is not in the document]
ErikBjare: `method.clone()` can't be done inside the `move ||`, it would be moved then cloned.

src/sync.rs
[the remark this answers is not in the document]
ErikBjare: While that kinda fixed it, it still leaves all kinds of unexpected behavior so until we add a `ForceCommit` or something I've just set `commit: true` when initializing the `DatastoreInstance` for now.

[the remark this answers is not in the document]
ErikBjare: Nevermind, added the `ForceCommit`, was really easy.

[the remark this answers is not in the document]
ErikBjare: Still leads to loads of unexpected behavior though. For example: If I do a `insert_events` and then a `get_event_count` I expect it to return the new number, not the old one (there are a lot of cases like this).

The solution as I see it is to commit pending operations *before* a get operation.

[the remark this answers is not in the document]
johan-bjareholt: > The solution as I see it is to commit pending operations before a get operation.

This is how the sqlite storage method in python does it, might be a good idea yes.

src/datastore/datastore.rs
```diff
@@ -213,35 +216,40 @@ impl DatastoreWorker {
...
                             Ok(b) => Ok(Responses::Bucket(b)),
                             Err(e) => Err(e)
                         }
                     },
                     Commands::GetBuckets() => {
+                        ds.commit = true;
                         Ok(Responses::BucketMap(ds.get_buckets()))
                     },
                     Commands::InsertEvents(bucketname, events) => {
+                        ds.commit = true;
```
johan-bjareholt: Remove

src/datastore/datastore.rs
```diff
@@ -762,4 +776,14 @@ impl Datastore {
             Err(e) => Err(e)
         }
     }
+
+    pub fn commit(&self) -> Result<(), DatastoreError> {
```
johan-bjareholt: Name this force_commit to make it clear that it is usually not necessary.
You could even add a sentence of why it exists.

src/sync_main.rs
```diff
@@ -0,0 +1,21 @@
...
+    println!("Started aw-sync-rust...");
+    logging::setup_logger().expect("Failed to setup logging");
+
+    aw_server::sync::sync_run();
+    info!("Finished successfully, exiting...");
+
+    // Needed to give the datastores some time to commit before program is shut down.
+    // 100ms isn't actually needed, seemed to work fine with as little as 10ms, but I'd rather give
+    // it some wiggleroom.
+    std::thread::sleep(std::time::Duration::from_millis(100));
```
johan-bjareholt: In C you don't need to sleep but can wait for a thread to close down.

How to do it properly in Rust without any risk of deadlocks though I have no idea...

src/sync.rs
```diff
@@ -0,0 +1,157 @@
...
+        //let new_eventcount = ds.get_event_count(bucket.id.as_str(), None, None).unwrap();
+        //info!("Eventcount: {:?} ({} new)", new_eventcount, events.len());
+        datastores.push(ds);
+    };
+    Ok(datastores)
+}
+
+/// Returns the sync-destination bucket for a given bucket, creates it if it doesn't exist.
+fn get_or_create_sync_bucket(bucket_from: &Bucket, ds_to: &Datastore) -> Bucket {
+    // Ensure the bucket ID ends in "-synced"
```
johan-bjareholt: Would be cool with a "synced" property to the buckets instead.

ErikBjare:
@johan-bjareholt The syncing could either happen in the same process, or in a separate process if all write-access to the local `aw-server` happens through the REST API (which is what I'm considering for the MVP).

src/datastore/datastore.rs
ErikBjare: I can't until we've fixed the commit-before-get. If I do I have to put `ds.force_commit()` all over the place which is really annoying and hard to debug.

src/sync_main.rs
ErikBjare: If you find a way we could always wait and have a timeout to avoid a deadlock.

ErikBjare:
>Have you tried to change the transaction behavior to IMMEDIATE instead of EXCLUSIVE?

Nope.

---

Can I merge this so we can get on with merging #65?

johan-bjareholt:
Feel free to merge, I don't approve of the syncing code yet but that's just a prototype. I don't approve of the ds.commit=true either but it's not broken and I can fix that in #65 so it's fine.

You just need to fix some conflict?

PR #67, opened 2019-09-20
Implemented classifier transform

ErikBjare:


Cargo.toml
```diff
@@ -36,6 +36,7 @@ log = "0.4"
 fern = { version = "0.5", features = ["colored"] }
 toml = "0.5"
 gethostname = "0.2"
+regex = "1.0.5"
```
johan-bjareholt: Don't specify patch version

src/models/event.rs
```diff
@@ -16,6 +16,17 @@ pub struct Event {
 }
 
 impl Event {
+    pub fn new_test() -> Event {
```
johan-bjareholt: You should not put test code functions in test structs

src/transform/classify.rs
```diff
@@ -0,0 +1,64 @@
...
+/// Based on code in aw_research: https://github.com/ActivityWatch/aw-research/blob/master/aw_research/classify.py
+
+use regex::Regex;
+use serde_json;
+
+use std::collections::HashSet;
+use crate::models::Event;
+
+pub fn classify(events: Vec<Event>, classes: Vec<(String, Regex)>) -> Vec<Event> {
+    // TODO: There is probably a better way that avoids the clone?
```
johan-bjareholt: iter() doesn't take ownership of Event and classify_one takes in a mut event so it's not possible currently.
I would suggest to instead of using iter use some iteration technique which takes ownership of the event and actually consumes the vector (drain for example does this).

src/transform/classify.rs
```diff
@@ -0,0 +1,64 @@
...
+        if item.matches("->").count() >= acc.matches("->").count() {
+            item
+        } else {
+            acc
+        }
+    }).clone()
+}
+
+#[test]
+fn test_classify() {
```
johan-bjareholt: Currently we have all tests in the tests folder.

The Rust way to do things is to have unittests inside the source and integration tests inside the tests directory. Not sure which way is the best way for us.

ErikBjare: I quite like having small unittests close to the function definition. Although it doesn't seem to show up in the coverage properly?

src/models/event.rs
ErikBjare: I know, I just put it there because I didn't know where to put test utils. Instead, I rewrote it into a plain `new` helper function, is that fine?

src/transform/classify.rs
ErikBjare: Nice, fixed.

src/models/event.rs
```diff
@@ -16,6 +16,15 @@ pub struct Event {
 }
 
 impl Event {
+    pub fn new() -> Event {
```
johan-bjareholt: Doesn't really make sense to me to have a `new()` function which only sets default values and doesn't take any args, better then to implement the Default trait for Event and then do `Event { ..Default::default() }`

src/query/functions.rs
```diff
@@ -282,6 +304,20 @@ mod validate {
...
     }
 
+    pub fn arg_type_list_of_list_of_strings (arg: &DataType) -> Result<Vec<Vec<String>>, QueryError> {
+        let mut tagged_lists = arg_type_list(arg)?.clone();
+        let mut lists: Vec<Vec<String>> = Vec::new();
+        for list in tagged_lists.drain(..) {
+            match list {
+                DataType::List(_) => lists.push(arg_type_string_list(&list)?.clone()),
+                ref invalid_type => return Err(QueryError::InvalidFunctionParameters(
+                    format!("Expected function parameter of type list of tuples of strings, list contains {:?}", invalid_type)
```
johan-bjareholt: You don't check for a tuple here, you check for an array. An array of a forced length would be similar to a tuple though as long as the tuple contains only one type if that's what you want to achieve.

src/transform/classify.rs
```diff
@@ -0,0 +1,69 @@
...
+///
+/// Based on code in aw_research: https://github.com/ActivityWatch/aw-research/blob/master/aw_research/classify.py
+
+use regex::Regex;
+use serde_json;
+
+use std::collections::HashSet;
+use crate::models::Event;
+
+pub fn classify(mut events: Vec<Event>, classes: Vec<(String, Regex)>) -> Vec<Event> {
```
johan-bjareholt: Maybe I'm missing something, but isn't it unnecessary to take ownership of the classes here?

src/transform/classify.rs
```diff
@@ -0,0 +1,69 @@
...
+        ("#test-tag".into(), Regex::new(r"test").unwrap()),
+        ("Test".into(), Regex::new(r"test").unwrap()),
+        ("Test -> Subtest".into(), Regex::new(r"test").unwrap()),
+        ("Other".into(), Regex::new(r"nonmatching").unwrap()),
+    );
+    let events_classified = classify(events, classes);
+
+    assert_eq!(events_classified.len(), 1);
+    assert_eq!(events_classified.first().unwrap().data.get("$tags").unwrap().as_array().unwrap().len(), 3);
+    assert_eq!(events_classified.first().unwrap().data.get("$category").unwrap(), &serde_json::json!("Test -> Subtest"));
```
johan-bjareholt: More asserts.

I think you should extend or write more tests for this.

src/models/event.rs
johan-bjareholt: See my other comment about default trait

src/transform/classify.rs
johan-bjareholt: Yeah, coverage is kind of broken currently anyway though, so whatever I guess

src/transform/classify.rs
```diff
@@ -0,0 +1,69 @@
...
+    // for instance: "Work -> ActivityWatch -> aw-server-rust"
+    // A category is chosed out of the tags used some rule (such as picking the one that's deepest in the hierarchy)
+    let category = choose_category(tags);
+    event.data.insert("$category".into(), serde_json::json!(category));
+    event
+}
+
+fn choose_category(tags: HashSet<String>) -> String {
+    tags.iter().fold(&"Uncategorized".to_string(), |acc: &String, item: &String| {
+        if item.matches("->").count() >= acc.matches("->").count() {
```
johan-bjareholt: This is bad practice, please don't use -> anywhere in the server code and let the web-ui pretty-print it with -> later.

src/transform/classify.rs
```diff
@@ -0,0 +1,69 @@
...
+    for event in events.drain(..) {
+        classified_events.push(classify_one(event, &classes));
+    }
+    return classified_events;
+}
+
+fn classify_one(mut event: Event, classes: &Vec<(String, Regex)>) -> Event {
+    let mut tags: HashSet<String> = HashSet::new();
+    for (cls, re) in classes {
+        for val in event.data.values() {
```
johan-bjareholt: I'm thinking out loud now:

Maybe there's an issue with going through all values by default? Maybe we should supply which fields in the event we want to match data with? Or maybe it's better to just filter the fields before calling classify?

Probably the latter now that I think of it...

src/transform/classify.rs
```diff
@@ -0,0 +1,69 @@
...
+            // TODO: Recurse if value is object/array
+            if val.is_string() && re.is_match(val.as_str().unwrap()) {
+                tags.insert(cls.clone());
+                break;
+            }
+        }
+    }
+
+    // An event can have many tags
+    event.data.insert("$tags".into(), serde_json::json!(tags));
```
johan-bjareholt: What does the "$" mean?

src/models/event.rs
ErikBjare: Huh, had no idea about that. Might give it a shot.

src/transform/classify.rs
ErikBjare: Just some prefix I chose for "computed property". I'm not sure if this is how we want to do it but it's how I've been doing it for a while and it's not that bad.

src/transform/classify.rs
ErikBjare: Having tags/categories be arrays doesn't make sense imo.

The "->" is not primarily intended to be pretty, it's intended to:

 1. Convey the hierarchy order
 2. Act as an intuitive separator
 3. Differentiate a category from a tag

I'm really unconvinced about changing this into an array, I've been writing all the classifiers like this and it's really not that bad. I'm actually very happy with using "->" in this way, even though I get why you might not like it.

src/transform/classify.rs
ErikBjare: In the aw-research classifier I actually only match on the fields "title", "app", and "url". I couldn't decide on what was the right approach here so I settled for the thing that made the least assumptions.

src/transform/classify.rs
ErikBjare: It is, I'll fix it.

src/transform/classify.rs
ErikBjare: Also, dot uses the "->" separator (which is where I got the idea from, I think). 

A nice/fun consequence of this is that if you basically dumped all your categories into a `.dot` file then you'd get a nice visualization of the category structure.

johan-bjareholt: > having classes be a Vec<(Vec<String>, Regex)> would be annoying

I disagree, it clearly defines what the data is without hiding anything

> The "->" is not primarily intended to be pretty, it's intended to:
>
> - Convey the hierarchy order
> - Act as an intuitive separator
> - Differentiate a category from a tag (a category is a special case of a tag)

To me all of those reasons say "pretty" but in a different way.

- Arrays have a order
- It's a separator, that's for sure. I don't know what intuitive means in this case as it's just a separator, but I assume that you mean more readable which I agree but that also comes at a cost of being a exotic construction which otherwise doesn't fit how the other transforms work which might actually make it less intuitive.
- A comment could just as well serve as a differentiation

> Also, dot uses the "->" separator (which is where I got the idea from, I think).
>
> A nice/fun consequence of this is that if you basically dumped all your categories into a .dot file then you'd get a nice visualization of the category structure.

Doing `["a", "b"].join("->")` would do the same thing, it's not something complex we're talking about here.

Other reasons to avoid it:

- Dot is a whole markup language, this is an entry in an already existing language which means that it's an format within a format which is messy
- Using -> means extra parsing and simply matching them with the currently weak system by simply splitting them would mean possible errors. There's a huge amount of edge cases in both the aw-server-python and aw-server-rust code which needs extra testing because of this.
- If someone automates something which creates tags and it contains "->" it will get messy unless you escape the -> somehow.
- Arrays already have a standard hierarchy order which is left to right (same as your "->" solution)

Just because you are used to something and that it currently works doesn't mean that it's a good idea.

src/transform/classify.rs
johan-bjareholt: Sounds  sane.
Maybe we should do the same with the fields we generate with "split_url_event" function which is already in aw-server?

src/transform/classify.rs
ErikBjare: >I disagree, it clearly defines what the data is without hiding anything

It doesn't. It says a tiny bit about how tags are for whatever reason `Vec`'s, which helps the Rust compiler a little bit but doesn't define anything about the underlying behavior of tags/categories. This is unlike using `#` for tags and `->` for categories, which makes it apparent to the reader that there is a distinction.

>A comment could just as well serve as a differentiation

No. Tags begin with `#`, categories have a hierarchy order. These need to be checked at runtime, and I'm not going to both check if an array is 1-length **and** starts with a `#` to determine if it's a tag or not.

>Using -> means extra parsing and simply matching them with the currently weak system by simply splitting them would mean possible errors. There's a huge amount of edge cases in both the aw-server-python and aw-server-rust code which needs extra testing because of this.

Not really, the parsing is literally just a check for `#` or `->` and then a `split`. I'm not sure what edge cases you are referring to and I can't recall any open issues about it.

>If someone automates something which creates tags and it contains "->" it will get messy unless you escape the -> somehow.

This seems extremely unlikely, and would be easy for them to just avoid doing.

---

Regardless, I've gone over solutions like these before in the aw-research code, and what I've come up with was the easiest to work with in the end. So unless you want to get into a long debate over all the other options I've explored that solve the same problem in a better way than the `Vec` you're proposing (such as letting tags have metadata, or letting the class spec be a dict) I'm going to go ahead and merge this.

src/transform/classify.rs
ErikBjare: We should, I'll fix it in this batch of related PRs.

src/transform/classify.rs
johan-bjareholt: > So unless you want to get into a long debate over all the other options I've explored that solve the same problem in a better way than the Vec you're proposing (such as letting tags have metadata, or letting the class spec be a dict) I'm going to go ahead and merge this.

I'd love to be in the loop about that. Please don't merge this.

johan-bjareholt: Another option would be to have a category and/or tag type in the query language.

ErikBjare: As I said, the other two options are basically:

 1. Letting tags have metadata, giving the class spec the format `[[name, regex, metadata], ...]` or `[{name: "ActivityWatch", regex: "", metadata: { type: "category", parent: "Programming" }}, ...]`. It has a few upsides like making it slightly more flexible (one could, for example, set a priority for categories that is used instead of depth when ranking categories to pick). But mostly seems like unecessarily adding complexity to me, and makes for annoying type signatures (`Vec<(String, Regex, Map<String, String>)>` or `Vec<Map<String, JSON::Value>>`). We could of course make structs for these, but again it leads to a messier code.
 2. Make the class spec a dict, such as: `{ Programming: { $re: "", ActivityWatch: { $re: "[Aa]ctivity[Ww]atch|aw-"}}}`. This is nice for putting the thing in a TOML-file as I have in aw-research, but I normalize the entire thing to the `Parent-cat -> Child-cat` format before I do the actual classification anyway.

Both those options have terrible readability and would complicate the code a fair bit.

A third option is to split up the behavior for tags and categories entirely, in which case I'd be fine with using your `Vec<(Vec<String>, Regex)>` for the category part and keep using `Vec<(String, Regex)>` for plain tags. But I don't think that makes sense since it's more or less identical code apart from selecting the highest-ranking category.

Overall I think you're bikeshedding this. It really isn't worth bickering over.

johan-bjareholt: > Overall I think you're bikeshedding this. It really isn't worth bickering over.

This is something we will maintain for years, it's worth getting it right so we don't end up like how we have changed the HTTP API over the years with breaking changes left and right.

johan-bjareholt: Sorry for the short answer before, did not have time to write a proper answer (and probably don't have now either).

> I obviously can't disagree with that, but it doesn't help me to merge this PR

Reviews aren't supposed to help getting something merged, it's supposed to increase the quality of the code getting merged.
It's also already been as a PR for 9 days and nothing else has been merged since so there shouldn't be a hurry.
You should also expect that a bigger PR should get a longer review time, I have only taken a quick glance at the code as of now.

> A third option is to split up the behavior for tags and categories entirely, in which case I'd be fine with using your Vec<(Vec<String>, Regex)> for the category part and keep using Vec<(String, Regex)> for plain tags. But I don't think that makes sense since it's more or less identical code apart from selecting the highest-ranking category.

To me that sounds reasonable. If they have no reason to be coupled and they do different things they shouldn't be coupled.

> What does this even mean? How does that affect the underlying transform?

It was just kind of a stupid idea to maybe have a construct in the language to make things more clear. Either just defining a function called category("work", "project") which returns a actual category type, or possibly even have a custom syntax such as category["work" -> "project"]. I don't think it's a good idea, just brainstorming to see if it's possible to get the best of both worlds.

ErikBjare: >Reviews aren't supposed to help getting something merged, it's supposed to increase the quality of the code getting merged.

It's the same thing in the end: We only merge code that's in good shape, and reviews help get code in shape, therefore reviews help get things merged.

>It's also already been as a PR for 9 days and nothing else has been merged since so there shouldn't be a hurry.

It's been stale for 9 days in part because I myself have been bikeshedding over the very thing you requested changes on since you were seemingly adamant about the status quo not being satisfactory.

>You should also expect that a bigger PR should get a longer review time, I have only taken a quick glance at the code as of now.

But it's not big? It's +177 -11. (You might be thinking about the aw-webui PR)

>To me that sounds reasonable. If they have no reason to be coupled and they do different things they shouldn't be coupled.

Really? This would lead to a lot of almost duplicated code/APIs in several places (two query2 functions, two transforms, two settings in the web UI). They don't really do different things, they do 90% exactly the same thing, apart from the category selection at the end. All that work just to get rid of a few string splits with `->` seems ridiculous tbh.

johan-bjareholt: > All I was saying was that your short comment didn't help me fix the issues you mentioned.

Fair enough, misunderstood you

> But it's not big? It's +177 -11. (You might be thinking about the aw-webui PR)

The API part is a big thing to me. This will be with us for a long time.

> Really? This would lead to a lot of almost duplicated code/APIs in several places (two query2 functions, two transforms, two settings in the web UI). They don't really do different things, they do 90% exactly the same thing, apart from the category selection at the end. All that work just to get rid of a few string splits with -> seems ridiculous tbh.

Two query2 functions is fine, that boilerplate code is minimal and stable.
Two transforms is acceptable. It's worth it for having to avoid a custom string format IMO.
Two settings in the web UI is probably a good thing if they don't belong with each other, no point in mixing them really.

Also, currently there does not seem to be any examples of using tagging in the visualization yet?

johan-bjareholt: Another option would be to try to be more generic and merge the functionality of categories and tags and just have them as one implementation and instead add the ability to save different sets of classes and choose which one you want to use.

Maybe make it optional to have a single event have multiple tags/categories or not and always allow the depth to be more than 1. If you don't want the depth to be more than 1, simply don't specify any classes which has more than 1 entry. That's essentially the same way we do with filter_events_by_keys, it allows supplying multiple keys but most often we only supply one.

Kind of hard to explain, but I hope I get my point across.

src/transform/classify.rs
```diff
@@ -34,56 +65,194 @@ fn classify_one(mut event: Event, classes: &Vec<(String, Regex)>) -> Event {
...
+pub fn categorize(mut events: Vec<Event>, rules: &Vec<(Vec<String>, Regex)>) -> Vec<Event> {
+    let mut classified_events = Vec::new();
+    for event in events.drain(..) {
+        classified_events.push(categorize_one(event, rules));
+    }
+    return classified_events;
+}
+
+// TODO: Classes should be &Vec<(String, Rule)>
+pub fn autotag(mut events: Vec<Event>, rules: &Vec<(String, Regex)>) -> Vec<Event> {
```
johan-bjareholt: autotag makes it sound like it comes up with the tags itself (which would be really cool, but unfortunately that's not what it does).
Just "match_tags" would better fit as a name.

src/transform/classify.rs
```diff
@@ -34,56 +65,194 @@ fn classify_one(mut event: Event, classes: &Vec<(String, Regex)>) -> Event {
...
+    assert_eq!(
+        events
+            .first()
+            .unwrap()
+            .data
+            .get("$tags")
+            .unwrap()
+            .as_array()
+            .unwrap()
+            .len(),
```
johan-bjareholt: Lol, that's one way to assert I guess.
Would be more readable with something like

```
let event = events.first.unwrap();
let tags = event.get("$tags").unwrap();
assert_eq!(tags.as_arrat().unwrap().len(), 3);
```

src/transform/classify.rs
ErikBjare: I renamed it to just `tag` because otherwise, I feel like we should also rename the `categorize` function to `match_category` and that just seems unnecessarily wordy.

src/query/datatype.rs
```diff
@@ -0,0 +1,253 @@
...
+    fn try_from(value: &DataType) -> Result<Self, Self::Error> {
+        let rulemap: HashMap<String, String> = (match value {
+            DataType::Dict(d) => {
+                let map: HashMap<String, String> = d.iter().map(|(k, v)| {
+                    let s: String = v.try_into().unwrap();
+                    (k.clone(), s.clone())
+                }).collect();
+                Ok(map)
+            },
+            DataType::String(s) => {
```
johan-bjareholt: I can't decide if it is a good idea to accept strings and assume that they are a regex.
It feels like it's a nice way to easily use the API, but at the same time I want things to be "explicit rather than implicit".

src/query/datatype.rs
```diff
@@ -0,0 +1,253 @@
...
+            (DataType::Event(e1), DataType::Event(e2)) => e1 == e2,
+            (DataType::List(l1), DataType::List(l2)) => l1 == l2,
+            (DataType::Dict(d1), DataType::Dict(d2)) => d1 == d2,
+            // We do not care about comparing functions
+            _ => false
+        }
+    }
+}
+
+impl TryFrom<&DataType> for Vec<DataType> {
```
johan-bjareholt: Awesome with all these TryFrom!

Also nice that you moved types to a seperate file.

src/query/datatype.rs
ErikBjare: This will be removed, almost forgot about it. Thanks!

PR #69, opened 2019-10-27
Added device_id and exposed it through /api/0/info

ErikBjare:
This is needed for syncing.

Equivalent PR in aw-server: https://github.com/ActivityWatch/aw-server/pull/61

### DIFF
--- a/src/endpoints/mod.rs
+++ b/src/endpoints/mod.rs
@@ -1,3 +1,4 @@
+use std::fs;
 use std::path::PathBuf;
 use std::sync::Mutex;
 
@@ -6,7 +7,9 @@ use rocket::response::{NamedFile};
 use rocket::State;
 use rocket_contrib::json::JsonValue;
 use gethostname::gethostname;
+use uuid::Uuid;
 
+use crate::dirs;
 use crate::config::AWConfig;
 
 #[macro_export]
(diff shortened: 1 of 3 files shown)


src/endpoints/mod.rs
```diff
@@ -60,6 +63,23 @@ fn root_favicon(state: State<ServerState>) -> Option<NamedFile> {
...
 }
 
+/// Retrieves the device ID, if none exists it generates one (using UUID v4)
+fn get_device_id() -> String {
+    // TODO: Cache to avoid retrieving on every /info call
+    // TODO: How should these unwraps be removed?
+    //       Should this be propagated into a 500 Internal Server Error? How?
+    // I chose get_data_dir over get_config_dir since the latter isn't yet supported on Android.
+    let mut path = dirs::get_data_dir().unwrap();
+    path.push("device_id");
```
johan-bjareholt: Why not put it in the usual config file?

ErikBjare: Because it isn't really a configurable. You could change it, but you really shouldn't. Copying a config file to a different installation should be okay, and copying a device ID isn't (unless you really know what you are doing). There also isn't a default value to fall back to, unlike every other configurable. 

Also, as mentioned in the comment config files aren't supported on Android. 

Also, the entire config in aw-server-rust is commented out by default.

PR #70, opened 2019-10-29
Merge in aw-client-rust

ErikBjare:


ErikBjare:
@johan-bjareholt I made another change, could you quickly look at https://github.com/ActivityWatch/aw-server-rust/pull/70/commits/ac10094b60db846d931970d5ec4f2e3008252868?

ErikBjare:
> Looks fine, the only thing which is a bit confusing is that aw-server-rust also contains aw-client-rust which can be misleading.

I also realized this, we might want to rename the repo to reflect this monorepo-change sometime in the future.

ErikBjare:
I fixed some panic's (apparently no one has ever tried to actually run aw-server-rust on Windows, lol), but now rustc appears to complain when building aw-server-rust for the aw-client-rust tests? (it seems like the exit code 101 is from rustc anyway)

Could you take a look at it @johan-bjareholt?

PR #73, opened 2019-11-03
datastore: Implement aw-server-python db import

johan-bjareholt:


johan-bjareholt:
@ErikBjare Compiles for android, merging now.

Feel free to review still though, I will fix whatever comments you write.

Also, please test this with your database as well, I've only tried this on my desktop and laptop.

Issue #60, opened 2019-08-27
DELETE endpoint to delete event


johan-bjareholt:

ErikBjare:
We should keep in mind that we might want to use the pragma [`secure_delete`](https://www.sqlite.org/pragma.html#pragma_secure_delete) or run a [`VACUUM`](https://www.sqlite.org/lang_vacuum.html).

johan-bjareholt:
Maybe run VACUUM when aw-server is starting up?
Seems like a sane time to do it.

johan-bjareholt:
secure_delete=FAST seems like the best approach for us I believe.

ErikBjare:
> Maybe run VACUUM when aw-server is starting up?
> Seems like a sane time to do it.

If the operation is somewhat costly then I'd rather not as that could slow down the startup time of the Android app (less of a concern on desktop, where the user is unlikely to use ActivityWatch the moment it's started).

johan-bjareholt:
> If the operation is somewhat costly then I'd rather not as that could slow down the startup time of the Android app (less of a concern on desktop, where the user is unlikely to use ActivityWatch the moment it's started).

Hm, I guess that's true.

A solution would be to have it as both an API endpoint to manually start a VACUUM as well as maybe store the last time a VACUUM was performed and do it once every X months.

johan-bjareholt:
So the delete endpoint for a single event just as for aw-server-python is now implemented but there is no VACUUM for now.
Maybe we should have a new issue for VACUUM?

johan-bjareholt:
Updated the issue body

PR #81, opened 2020-01-12
Add support for deleting events

johan-bjareholt:
Solves #60


johan-bjareholt:
macOS finishes but stalls on the caching step, merging anyway

PR #87, opened 2020-02-20
Add key value store

xylix:
This PR adds key/value storage functionality. Currently implementing changes based on review feedback.
        
- [x] implement PR feedback
  - [x] rename create_value() to insert_value or something of the sort
  - [x] 204 -> 404 for missing value
  - [x] log errors
  - [x] Add key length limit (maybe 100-128?)
  - [x] rename fns frddom x_value to x_key_value
  - [x] use ? operator in settings.rs
  - [x] nosuchvalue Error -> no such key error
  - [x] Maybe rename functions from insert to set?
  - [x] add response body to _set_setting_request parameters in tests

new features coming into this PR:
- [x] Take json object as input request body `{"key": "sample-key", "value": "sample_Value"}`
- [x] Return json objects (probably `{"key": "settings.sample-key", "value": "sample_Value"}
- [x] Add last_modified column to sql
  - [x] Test timestamps in some sane way
- [x] endpoint for listing all settings key-values
  - [x] test

Open questions:
  - [x] ~~Maybe different status msgs (created vs updated)~~ No.
  - [x] ~~Maybe return timestamp on setting set~~ No.

### DIFF
--- a/aw-datastore/src/datastore.rs
+++ b/aw-datastore/src/datastore.rs
@@ -599,7 +599,7 @@ impl DatastoreInstance {
         return Ok(count);
     }
 
-    pub fn create_value(&self, conn: &Connection, key: &str, data: &str)
+    pub fn insert_value(&self, conn: &Connection, key: &str, data: &str)
         -> Result<(), DatastoreError> {
 
         let mut stmt = match conn.prepare("
--- a/aw-datastore/src/datastore.rs
+++ b/aw-datastore/src/datastore.rs
@@ -644,23 +644,30 @@ impl DatastoreInstance {
         }
     }
 
-    pub fn get_values_starting(&self, conn: &Connection, pattern: &str) 
+    pub fn get_values_starting(&self, conn: &Connection, pattern: &str)
         -> Result<HashMap<String, String>, DatastoreError>
     {
         let mut stmt = match conn.prepare("SELECT * FROM key_value WHERE key LIKE '.%' ") {
             Ok(stmt) => stmt,
-            Err(err) => return Err(DatastoreError::InternalError(
-                format!("Failed to prepare get_value SQL statement: {}", err)
-            ))
+            Err(err) => {
+                return Err(DatastoreError::InternalError(format!(
+                    "Failed to prepare get_value SQL statement: {}",
+                    err
+                )))
+            }
         };
 
         let mut output = HashMap::<String, String>::new();
         // Unwrap to String tuple or panic on SQL row if types are invalid.
         // should never happen with a properly initialized table
-        let result = stmt.query_map(&[pattern], |row| Ok(
-            (row.get::<usize, String>(0).expect("Invalid sql column types"), 
-             row.get::<usize, String>(1).expect("Invalid sql column types")))
-        );
+        let result = stmt.query_map(&[pattern], |row| {
+            Ok((
+                row.get::<usize, String>(0)
+                    .expect("Invalid sql column types"),
+                row.get::<usize, String>(1)
+                    .expect("Invalid sql column types"),
+            ))
+        });
 
         match result {
             Ok(kv_pairs) => {
@@ -669,13 +676,14 @@ impl DatastoreInstance {
                     output.insert(k, v);
                 }
                 Ok(output)
-            },
+            }
             Err(err) => match err {
                 rusqlite::Error::QueryReturnedNoRows => Err(DatastoreError::NoSuchValue),
-                _ => Err(DatastoreError::InternalError(
-                    format!("Failed to get key_value rows starting with pattern {}", pattern))
-                )
-            }
+                _ => Err(DatastoreError::InternalError(format!(
+                    "Failed to get key_value rows starting with pattern {}",
+                    pattern
+                ))),
+            },
         }
     }
 }
--- a/aw-datastore/src/datastore.rs
+++ b/aw-datastore/src/datastore.rs
@@ -129,7 +129,8 @@ fn _migrate_v3_to_v4(conn: &Connection) {
     conn.execute("CREATE TABLE key_value (
         key TEXT PRIMARY KEY,
         value TEXT
-    );", rusqlite::NO_PARAMS).expect("Failed to upgrade db and add key-value storage table");
+    );", rusqlite::NO_PARAMS)
+        .expect("Failed to upgrade db and add key-value storage table");
 
     conn.pragma_update(None, "user_version", &4).expect("Failed to update database version!");
 }
@@ -633,14 +634,13 @@ impl DatastoreInstance {
         return match stmt.query_row(&[key as &str], |row|{
             row.get(1)
         }) {
-            Ok(result)  => if result == "QueryReturnedNoRows" {
-                Ok("No rows found".to_string())
-            } else {
-                Ok(result)
-            },
-            Err(err) => Err(DatastoreError::InternalError(
-                format!("Get value query failed for key {}", key))
-            )
+            Ok(result)  => Ok(result),
+            Err(err) => match err {
+                rusqlite::Error::QueryReturnedNoRows => Err(DatastoreError::NoSuchValue),
+                _ => Err(DatastoreError::InternalError(
+                              format!("Get value query failed for key {}", key))
+                )
+            }
         }
     }
 }
--- a/aw-datastore/src/datastore.rs
+++ b/aw-datastore/src/datastore.rs
@@ -643,4 +643,34 @@ impl DatastoreInstance {
             }
         }
     }
+
+    pub fn get_values_starting(&self, conn: &Connection, pattern: &str) -> Result<HashMap<String, String>, DatastoreError>{
+        let mut stmt = match conn.prepare("
+                SELECT * FROM key_value WHERE pattern LIKE ?1%") {
+            Ok(stmt) => stmt,
+            Err(err) => return Err(DatastoreError::InternalError(
+                format!("Failed to prepare get_value SQL statement: {}", err)
+            ))
+        };
+
+        let mut output = HashMap::<String, String>::new();
+        let result = stmt.query(&[pattern as &str]);
+
+        match result {
+            Ok(rows) => {
+                // Unwrap to String tuple or panic on SQL row if types are invalid.
+                // should never happen with a properly initialized table
+                rows.map(|row| Ok(output.insert(
+                    row.get::<usize, String>(0).expect("Invalid sql structure"),
+                    row.get::<usize, String>(1).expect("Invalid sql structure"))));
+                Ok(output)
+            },
+            Err(err) => match err {
+                rusqlite::Error::QueryReturnedNoRows => Err(DatastoreError::NoSuchValue),
+                _ => Err(DatastoreError::InternalError(
+                    format!("Failed to get key_value rows starting with pattern {}", pattern))
+                )
+            }
+        }
+    }
 }
--- a/aw-datastore/src/datastore.rs
+++ b/aw-datastore/src/datastore.rs
@@ -655,15 +655,19 @@ impl DatastoreInstance {
         };
 
         let mut output = HashMap::<String, String>::new();
-        let result = stmt.query(&[pattern]);
+        // Unwrap to String tuple or panic on SQL row if types are invalid.
+        // should never happen with a properly initialized table
+        let result = stmt.query_map(&[pattern], |row| Ok(
+            (row.get::<usize, String>(0).expect("Invalid sql column types"), 
+             row.get::<usize, String>(1).expect("Invalid sql column types")))
+        );
 
         match result {
-            Ok(rows) => {
-                // Unwrap to String tuple or panic on SQL row if types are invalid.
-                // should never happen with a properly initialized table
-                rows.map(|row| Ok(output.insert(
-                    row.get::<usize, String>(0).expect("Invalid sql structure"),
-                    row.get::<usize, String>(1).expect("Invalid sql structure"))));
+            Ok(kv_pairs) => {
+                for row in kv_pairs {
+                    let (k, v) = row.unwrap();
+                    output.insert(k, v);
+                }
                 Ok(output)
             },
             Err(err) => match err {
--- a/aw-datastore/src/datastore.rs
+++ b/aw-datastore/src/datastore.rs
@@ -637,9 +637,9 @@ impl DatastoreInstance {
 
         return match stmt.query_row(&[key], |row|{
             Ok(KeyValue {
-                key: row.get(0).unwrap(), 
-                value: row.get(1).unwrap(), 
-                timestamp: DateTime::from_utc(NaiveDateTime::from_timestamp(row.get(2).unwrap(), 0), Utc)
+                key: row.get(0)?, 
+                value: row.get(1)?,
+                timestamp: DateTime::from_utc(NaiveDateTime::from_timestamp(row.get(2)?, 0), Utc)
             }
             )}) 
         {
@@ -667,12 +667,13 @@ impl DatastoreInstance {
         };
 
         let mut output = Vec::<String>::new();
-        // Unwrap to String or panic on SQL row if type is invalid.
-        // should never happen with a properly initialized table
+        // Rusqlite's get wants index and item type as parameters.
         let result = stmt.query_map(&[pattern], |row| row.get::<usize, String>(0));
         match result {
             Ok(keys) => {
                 for row in keys {
+                    // Unwrap to String or panic on SQL row if type is invalid. Can't happen with a
+                    // properly initialized table.
                     output.push(row.unwrap());
                 }
                 Ok(output)
--- a/aw-datastore/src/datastore.rs
+++ b/aw-datastore/src/datastore.rs
@@ -610,7 +610,7 @@ impl DatastoreInstance {
                 format!("Failed to prepare insert_value SQL statement: {}", err)
             ))
         };
-        stmt.execute(&[key as &str, data as &str]).expect(
+        stmt.execute(&[key, data]).expect(
             &format!("Failed to insert key-value pair: {}", key)
         );
         return Ok(())
@@ -631,7 +631,7 @@ impl DatastoreInstance {
             ))
         };
 
-        return match stmt.query_row(&[key as &str], |row|{
+        return match stmt.query_row(&[key], |row|{
             row.get(1)
         }) {
             Ok(result)  => Ok(result),
@@ -644,9 +644,10 @@ impl DatastoreInstance {
         }
     }
 
-    pub fn get_values_starting(&self, conn: &Connection, pattern: &str) -> Result<HashMap<String, String>, DatastoreError>{
-        let mut stmt = match conn.prepare("
-                SELECT * FROM key_value WHERE pattern LIKE ?1%") {
+    pub fn get_values_starting(&self, conn: &Connection, pattern: &str) 
+        -> Result<HashMap<String, String>, DatastoreError>
+    {
+        let mut stmt = match conn.prepare("SELECT * FROM key_value WHERE key LIKE '.%' ") {
             Ok(stmt) => stmt,
             Err(err) => return Err(DatastoreError::InternalError(
                 format!("Failed to prepare get_value SQL statement: {}", err)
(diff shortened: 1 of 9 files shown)


aw-datastore/src/worker.rs
```diff
@@ -69,6 +74,17 @@ struct DatastoreWorker {
     quit: bool
 }
 
+fn _unwrap_response(receiver: ResponseReceiver<Result<Responses, DatastoreError>>)
```
johan-bjareholt: Regarding this function I think either we skip this and just continue copy-pasting as we did before or we make a general-purpose solution for more responses than just empty otherwise things get inconsistent. If we make a general purpose solution it might be necessary for it to be a macro though which gets more complicated.

aw-datastore/src/datastore.rs
```diff
@@ -581,4 +598,49 @@ impl DatastoreInstance {
 
         return Ok(count);
     }
+
+    pub fn create_value(&self, conn: &Connection, key: &str, data: &str)
```
johan-bjareholt: Create isn't a proper name as it also can replace. Insert or put something similar is more suitable.

aw-server/src/endpoints/key_value.rs
```diff
@@ -0,0 +1,37 @@
...
+#[post("/<key>", data="<message>")]
+pub fn value_new(state: State<ServerState>, key: String, message: Json<String>)
+    -> Result<Status, Status> {
+
+    let data = message.into_inner();
+    let datastore: MutexGuard<'_, Datastore> = endpoints_get_lock!(state.datastore);
+    let result = datastore.create_value(&key, &data);
+    return match result {
+        Ok(_) => Ok(Status::Created),
+        Err(_) => Err(Status::InternalServerError)
```
johan-bjareholt: We should also print the error to the log so it's actually debuggable.

aw-server/src/endpoints/key_value.rs
```diff
@@ -0,0 +1,37 @@
...
+
+#[post("/<key>", data="<message>")]
+pub fn value_new(state: State<ServerState>, key: String, message: Json<String>)
+    -> Result<Status, Status> {
+
+    let data = message.into_inner();
+    let datastore: MutexGuard<'_, Datastore> = endpoints_get_lock!(state.datastore);
+    let result = datastore.create_value(&key, &data);
+    return match result {
+        Ok(_) => Ok(Status::Created),
```
johan-bjareholt: Should it be the same status no matter whether if it's created or replaced?
Probably not, but maybe not very important to have that distinction for now either. Maybe just add a "TODO" comment for now if it's a significant amount of work.

aw-server/src/endpoints/key_value.rs
```diff
@@ -0,0 +1,37 @@
...
+        Err(_) => Err(Status::InternalServerError)
+    }
+}
+
+#[get("/<key>")]
+pub fn value_get(state: State<ServerState>, key: String) -> Result<String, Status> {
+    let datastore = endpoints_get_lock!(state.datastore);
+    return match datastore.get_value(&key) {
+        Ok(result) => Ok(result),
+        Err(DatastoreError::NoSuchValue) => Ok("No such value".to_string()),
```
johan-bjareholt: Return a 404?

aw-server/src/endpoints/key_value.rs
```diff
@@ -0,0 +1,37 @@
+use std::sync::MutexGuard;
+use rocket_contrib::json::Json;
+use rocket::http::Status;
+use rocket::State;
+use crate::endpoints::ServerState;
+
+use aw_datastore::{Datastore, DatastoreError};
+
```
johan-bjareholt: Maybe a endpoint for listing all keys?

aw-datastore/src/datastore.rs
```diff
@@ -581,4 +598,49 @@ impl DatastoreInstance {
...
+        -> Result<(), DatastoreError> {
+
+        let mut stmt = match conn.prepare("
```
johan-bjareholt: Maybe add an limit on how long keys can be? 100 chars?
SQLite TEXT does not have a limit, but we should probably add one anyway.

The values can be unlimited.

xylix: I'm thinking of adding a check to the api endpoints like this ```if key.len() > 128 { return Err(Status::BadRequest) }``` (128 is a nice and familiar number for max size imo. 100 is fine as well.)

I wouldn't think the internal implementation needs a check if we don't allow too long inputs, but could add that if it's necessary?

aw-server/src/endpoints/key_value.rs
xylix: Changed and resolved this in 6c7dd29

aw-server/src/endpoints/key_value.rs
xylix: Changed and fixed this in 6c7dd29

aw-datastore/src/datastore.rs
xylix: Changed and fixed this in 6c7dd29

aw-datastore/src/worker.rs
xylix: Hm I personally find that just having this helper function did make my code more readable. I experimented a bit with a general purpose solution using Rust's <> generics but had problems making it work. Would probably require at least changing the `Responses` from an enum to a trait or something like that. 

Won't probably be getting that ready for this PR, so I guess I'll remove this to make things more consistent?

aw-server/src/endpoints/key_value.rs
xylix: Hm I guess it might be useful? For the first use (storing settings as `key` - `value` pairs) it's probably not necessary but it's pretty low effort to add. I'll look into this.

xylix:
> I think a last_modified should be added, possibly also "created".

What would be the use cases for a "created" field if there already exists a last_modified one? 

I'll work on adding the last_modified column.

> I don't think we should just call it "key_value". The name should clearly define what it's made for. Maybe just call it "settings" with "sections"?

I personally would internally call it key_value for reusability, but perhaps rename the API endpoints to settings or something of the sort.

aw-datastore/src/datastore.rs
johan-bjareholt: Sounds fine!

johan-bjareholt:
> What would be the use cases for a "created" field if there already exists a last_modified one?

I don't know, that's why i said "possibly". It's fine to skip that.

> I personally would internally call it key_value for reusability, but perhaps rename the API endpoints to settings or something of the sort. Can you clarify did you wish to change everything or just some part of the chain?

Could be done, but if you want it to be reusable you should have a separate namespace for settings, maybe just prefix each key like "settings.key".

The risk of having it too general-purpose is that it might be that if we start using it for too much it might be able to do a lot but being bad at everything. For example if we want to start saving icons for every application we could base64 all of them and put them inside. The proper solution would be to add a new table to the database which saves it as blobs and add a separate API for it, but since it's so general purpose nothings stopping it from being used like that even though its a suboptimal solution. Restricting APIs down to what they're intended to do is good practice if it doesn't disturb usability. If an API can be misused it's just a matter of time before it will be misused.

So my opinion is that it's important to change the API name to something more specific as the API should preferably never change once settled. What it's called inside the code doesn't matter as much, but I think we should avoid using it for more than once purpose.

aw-datastore/src/worker.rs
johan-bjareholt: It's possible to do it with a macro, I can fix that later. If we consider it temporary it doesn't really matter if you keep it as is for now.

aw-datastore/src/datastore.rs
xylix: Implemented and fixed in fdc4554

aw-server/src/endpoints/settings.rs
```diff
@@ -6,15 +6,22 @@ use crate::endpoints::ServerState;
...
 use aw_datastore::{Datastore, DatastoreError};
 
+fn parse_key(key: String) -> Result<String, Status> {
+    if key.len() >= 128 { return Err(Status::BadRequest) }
+    else { Ok(key + ".settings")  }
+}
+
 #[post("/<key>", data="<message>")]
-pub fn value_new(state: State<ServerState>, key: String, message: Json<String>)
+pub fn setting_new(state: State<ServerState>, key: String, message: Json<String>)
```
ErikBjare: I'd simply call this `setting_set`.

aw-datastore/src/worker.rs
```diff
@@ -347,4 +381,33 @@ impl Datastore {
             Err(e) => Err(e)
         }
     }
+
+    pub fn insert_value(&self, key: &str, data: &str) -> Result<(), DatastoreError> {
```
ErikBjare: `set` instead of `insert`

aw-server/src/endpoints/settings.rs
```diff
@@ -0,0 +1,67 @@
...
+use rocket_contrib::json::Json;
+use rocket::http::Status;
+use rocket::State;
+use crate::endpoints::ServerState;
+
+use aw_datastore::{Datastore, DatastoreError};
+
+fn parse_key(key: String) -> Result<String, Status> {
+    if key.len() >= 128 { return Err(Status::BadRequest) }
+    else { Ok(key + ".settings")  }
```
ErikBjare: This is confusing to me for two reasons:

 - The `settings` should be prepended, not appended.
 - I'm not sure if the API should restrict access to other namespaces, but maybe you're right that it should.

xylix: Hm yes great point about it needing to be prepended, not appended haha. That's the intended behaviour, somehow it slipped my mind.

I think I'm convinced with @johan-bjareholt 's arguments about making the API's purpose clear and restricted, but creating the sql support in a way that theoretically could support new endpoints in the future is for the best.

xylix:
Hm yeah the `insert` was used to try be consistent with the internal bucket and events methods. I do agree with `set` being a bit clearer though, especially in the API endpoints code.

I'm currently working on adding a /settings endpoint returning a json of all the {key, value} pairs. I'd guess having the namespacing in server-side would be most clear along with implementing a listing feature as an API endpoint.

Agreed with the last_modified point as well.

xylix:
@johan-bjareholt My implementation behind the "get all settings" endpoint is failing with an unwrap issue Im having problems debugging. (backtrace: https://gist.github.com/xylix/0b09370eb57dc42df8ff5024f20c16b1 , commit 488762c probably easiest way to see the code)

Somehow the receiver is failing to unwrap but no other addition I have done to the worker.rs has had this problem, even when the SQL queries were failing. Now I think (I tried validating in multiple ways) the SQL and unwrapping the response is correct, but the worker is getting some weird result.

Have you had any problems like this / any tips on debugging this one?

johan-bjareholt:
> I'm currently working on adding a /settings endpoint returning a json of all the {key, value} pairs.

I think you should only return the keys, I hope we never get a use-case for dumping all the values. It's an API, we have no idea what's going to be put in there so dumping all values could become very slow.

johan-bjareholt:
> My implementation behind the "get all settings" endpoint is failing with an unwrap issue Im having problems debugging. (backtrace: https://gist.github.com/xylix/0b09370eb57dc42df8ff5024f20c16b1 , commit 488762c probably easiest way to see the code)

Replace the "expect" with an unwrap. Otherwise you can only see in the backtrace that it's an RecvError but can't see the reason for the RecvError. My first guess would be that the worker fails ungracefully with something and drops the sender and if there's no sender the receiver part will get an RecvError.

aw-server/src/endpoints/settings.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+#[post("/<key>", data = "<message>")]
+pub fn setting_new(
+    state: State<ServerState>,
+    key: String,
+    message: Json<String>,
+) -> Result<Status, Status> {
+    let setting_key = match parse_key(key) {
```
johan-bjareholt: You don't have to match it like that, you can just do the following on a single like.

    let setting_key = parse_key(key)?;

The ? operator is nice for removing this kind of boilerplate code.

aw-server/tests/api.rs
```diff
@@ -334,4 +334,110 @@ mod api_tests {
...
+    fn test_updating_setting() {
+        let server = setup_testserver();
+        let client = rocket::local::Client::new(server).expect("valid instance");
+
+        let response_status = create_value_request(&client, "test_key");
+        assert_eq!(response_status, rocket::http::Status::Created);
+
+        let res = client.post("/api/0/settings/test_key")
+            .header(ContentType::JSON)
+            .body(r#""changed_test_value""#)
```
johan-bjareholt: If you modify create_value_request to also take in the value you could re-use it here.

aw-datastore/src/datastore.rs
```diff
@@ -581,4 +598,84 @@ impl DatastoreInstance {
 
         return Ok(count);
     }
+
+    pub fn insert_value(&self, conn: &Connection, key: &str, data: &str)
```
johan-bjareholt: I'd prefer if these function names said "key_value" rather than just "value".
For example, insert_value does not necessarily only set the value but also sets a new key if it does not yet exist so calling it just "value" doesn't really make any sense.
Same thing goes for "delete" and all the others.

aw-datastore/src/datastore.rs
```diff
@@ -581,4 +598,84 @@ impl DatastoreInstance {
...
+            Err(err) => return Err(DatastoreError::InternalError(
+                format!("Failed to prepare get_value SQL statement: {}", err)
+            ))
+        };
+
+        let mut output = HashMap::<String, String>::new();
+        // Unwrap to String tuple or panic on SQL row if types are invalid.
+        // should never happen with a properly initialized table
+        let result = stmt.query_map(&[pattern], |row| Ok(
+            (row.get::<usize, String>(0).expect("Invalid sql column types"),
```
johan-bjareholt: What's up with the usize here? Isn't it supposed to just get the strings key and value? What is the usize for?

aw-datastore/src/datastore.rs
```diff
@@ -581,4 +598,84 @@ impl DatastoreInstance {
...
+        match result {
+            Ok(kv_pairs) => {
+                for row in kv_pairs {
+                    let (k, v) = row.unwrap();
+                    output.insert(k, v);
+                }
+                Ok(output)
+            },
+            Err(err) => match err {
+                rusqlite::Error::QueryReturnedNoRows => Err(DatastoreError::NoSuchValue),
```
johan-bjareholt: Shouldn't it be "NoSuchKey" rather than "NoSuchValue"?
A key should always have a value so "NoSuchValue" should be impossible.

aw-server/src/endpoints/key_value.rs
ErikBjare: I think it should be the same for created and updated. I don't see a reason for why you'd want to know the difference.

aw-datastore/src/datastore.rs
xylix: I couldn't get row.get to work with just one type parameter, it also wanted a type param for the index which is an usize. There should be a way to give in now type param and force it to a string, but I didn't figure out one when writing this.

aw-server/src/endpoints/settings.rs
xylix: Good point. I already used it at some of the newer changes but probably missed some.

aw-datastore/src/datastore.rs
xylix: True. I I'll change that.

aw-datastore/src/datastore.rs
xylix: That does make more sense. Will change.

xylix: Fixed in 8e07f60

xylix:
I guess it makes sense to change the list endpoint to keys only. I'll do that.

Also I changed the .expect to .unwrap, It didn't really improve the stacktrace :/ (https://gist.github.com/xylix/1b85a4fe805241862440ea8cd420ef44 )

johan-bjareholt:
https://github.com/ActivityWatch/mpsc_requests/blob/master/src/lib.rs#L128

The only time RecvError is returned is if the response channel is broken so something is going wrong in the worker thread. If you run with a debugger like gdb on the test binary you will see what the other thread is doing (or of it is still alive).

xylix:
Allright I did some debugging and peeked at the rusqlite code, it seems like the problem was the sqlite `LIKE '?%'` in my query, sqlite didn't like the escaping by the ?% combination. Will probably have a working commit in a bit.

aw-server/tests/api.rs
xylix: Done in abcfc49

xylix:
Last modified column has now been added, altough the tests are still lacking. I couldn't figure out a sane way to test the timestamps of settings since there's really no way for the tests now at what time an insertion query finished.

aw-datastore/src/datastore.rs
xylix: Fixed in 8277c1b

ErikBjare:
For the test: Just check that `last_modified` is in the range `(timestamp just before set, timestamp just after set)`. Should be enough.

xylix:
There's still the open question about different return statuses on first insert vs update, and also I was thinking that maybe it'd be useful to return timestamp on setting set, but I don't really have any idea about it's uses outside of confirming time of insertion in client side, which shouldn't be needed anyway.

aw-models/src/key_value.rs
```diff
@@ -0,0 +1,27 @@
+use chrono::{DateTime, Utc};
+
+#[derive(Serialize, Deserialize)]
+pub struct Key {
+    pub key: String
+}
+
+// TODO: Invent a better naming scheme than calling the non-timestampy one "KV"
```
johan-bjareholt: The simple solution would be to wrap the timestamp in the KeyValue in an Option

aw-models/src/key_value.rs
```diff
@@ -0,0 +1,27 @@
+use chrono::{DateTime, Utc};
+
+#[derive(Serialize, Deserialize)]
+pub struct Key {
```
johan-bjareholt: I think this might be overkill, might as well just use a String wherever this is used.

aw-server/src/endpoints/settings.rs
```diff
@@ -0,0 +1,84 @@
...
+#[post("/", data = "<message>")]
+pub fn setting_set(state: State<ServerState>, message: Json<KV>) -> Result<Status, Status> {
+    let data = message.into_inner();
+
+    let setting_key = parse_key(data.key)?;
+
+    let datastore: MutexGuard<'_, Datastore> = endpoints_get_lock!(state.datastore);
+    let result = datastore.insert_key_value(&setting_key, &data.value);
+    return match result {
+        // TODO: Different status for replacement / creation (requires some sql adjustment)
```
johan-bjareholt: No need to do this I think

First of all I have no idea how to do it in sqlite in a performant way.
Secondly while there might possibly in the future be a use-case for checking this, you can just as well use the settings_list_get first to see if it will create or update. Not as fast, but should be good enough for most use-cases.

aw-models/src/key_value.rs
xylix: Hm yeah I actually tried to use a string, but concatenating the `{key: <key_variable>}` string together seemed much less clear and harder to write without typo's than just using this. I guess wrapping the timestamp in an option would solve this and the upper comment though.

aw-models/src/key_value.rs
xylix: Fixed in 37e7623

aw-server/src/endpoints/settings.rs
xylix: Yeah makes sense. I'll remove the TODO, let's just do an issue / feature request if for some reason this becomes necessary in the future.

xylix: fixed in 28f4c75

aw-models/src/key_value.rs
johan-bjareholt: Alright!

ErikBjare:
I'm happy! Merging, and then fixing #88.

PR #88, opened 2020-02-23
Add git hooks to enforce formatting, clippy and tests

Dentosal:
I also ran rustfmt for the whole codebase, and fixed one clippy error.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all aw-server aw-webui build install package test
+.PHONY: all aw-server aw-webui build install package test install-pre-commit pre-commit
 
 all: build
 build: aw-server aw-webui
@@ -11,6 +11,19 @@ aw-server:
 
 aw-webui:
 	make -C ./aw-webui build
+
+install-git-hooks:
+	@printf "make pre-commit\n" > .git/hooks/pre-commit
+	@printf "make pre-push\n" > .git/hooks/pre-push
+	@chmod +x .git/hooks/pre-commit .git/hooks/pre-push
+	@printf "Hooks installed\n"
+
+pre-commit:
+	@cargo fmt -- --check || (printf "Error: Run cargo fmt before committing\n"; exit 1)
+
+pre-push: pre-commit
+	@cargo clippy || (printf "Error: Clippy reported error(s)\n"; exit 1)
+	@cargo test || (printf "Error: Some test(s) failed\n"; exit 1)
 
 test:
 	cargo test
--- a/README.md
+++ b/README.md
@@ -40,3 +40,10 @@ If you want to quick-compile for debugging, run cargo run from the project root
 *NOTE:* this will start aw-server-rust on the testing port 5666 instead of port 5600
 
 ``` cargo run ```
+
+
+### Contributing
+
+Please run `make install-git-hooks` before you start working. This will ensure you only commit properly formatted files, and only push changes that pass tests.
+
+Please use `cargo fmt` and `cargo clippy`; they are quite helpful when developing. Formatter will make the code style consistent, and Clippy will make sure that most common errors and readibility issues will not slip through.
--- a/aw-server/src/config.rs
+++ b/aw-server/src/config.rs
@@ -79,7 +79,9 @@ pub fn get_config() -> AWConfig {
         for line in default_config_str.lines() {
             default_config_str_commented.push_str(&format!("#{}\n", line));
         }
-        wfile.write(&default_config_str_commented.into_bytes()).expect("Failed to write config to file");
+        wfile
+            .write_all(&default_config_str_commented.into_bytes())
+            .expect("Failed to write config to file");
         wfile.sync_all().expect("Unable to sync config file");
     }
 


ErikBjare:
Could you please avoid running `cargo fmt` directly in the PR? We'll run it on master when there are no other open PRs to cause conflicts.

Dentosal:
I removed the commit for `cargo fmt`. This PR should not be merged before the master is formatted, as the git hooks verify that the formatting is correct. I'm hoping that the master can this can be done soon, as it's hard to work with an unformatted codebase.

ErikBjare:
@Dentosal Gotcha! I'll fix that as soon as #87 is merged.

Makefile
```diff
@@ -12,6 +12,19 @@ aw-server:
 aw-webui:
 	make -C ./aw-webui build
 
+install-git-hooks:
+	@printf "make pre-commit\n" > .git/hooks/pre-commit
+	@printf "make pre-push\n" > .git/hooks/pre-push
```
ErikBjare: Do git hooks always execute with the root project directory as the working directory?

Dentosal: Yes, as per [the docs](https://git-scm.com/docs/githooks/2.24.0).

Makefile
```diff
@@ -12,6 +12,19 @@ aw-server:
...
+	@chmod +x .git/hooks/pre-commit .git/hooks/pre-push
+	@printf "Hooks installed\n"
+
+pre-commit:
+	@cargo fmt -- --check || (printf "Error: Run cargo fmt before committing\n"; exit 1)
+
+pre-push: pre-commit
+	@cargo clippy || (printf "Error: Clippy reported error(s)\n"; exit 1)
+	@cargo test || (printf "Error: Some test(s) failed\n"; exit 1)
```
johan-bjareholt: I don't think we should run this in pre-push as running all tests takes a long time.
It will be verified in CI as well so you can't skip it anyway.

Makefile
```diff
@@ -12,6 +12,19 @@ aw-server:
...
+	@printf "make pre-commit\n" > .git/hooks/pre-commit
+	@printf "make pre-push\n" > .git/hooks/pre-push
+	@chmod +x .git/hooks/pre-commit .git/hooks/pre-push
+	@printf "Hooks installed\n"
+
+pre-commit:
+	@cargo fmt -- --check || (printf "Error: Run cargo fmt before committing\n"; exit 1)
+
+pre-push: pre-commit
+	@cargo clippy || (printf "Error: Clippy reported error(s)\n"; exit 1)
```
johan-bjareholt: I think it would make more sense if this simply printed the error, the custom error message doesn't say anything useful but clippy itself does so might as well just let it print.

Makefile
```diff
@@ -12,6 +12,19 @@ aw-server:
 aw-webui:
 	make -C ./aw-webui build
 
+install-git-hooks:
+	@printf "make pre-commit\n" > .git/hooks/pre-commit
+	@printf "make pre-push\n" > .git/hooks/pre-push
+	@chmod +x .git/hooks/pre-commit .git/hooks/pre-push
+	@printf "Hooks installed\n"
```
johan-bjareholt: This print is pointless, it's never printed due to the @

Makefile
xylix: In my experience running aw-server-rust's texts usually takes 10-20 seconds. Fair point about the CI though.

johan-bjareholt: The tests execute fast yes, but when you count in compilation time it is very slow. Sometimes you only want to modify a single crate and not the whole project, in that case it might take 5 minutes extra with this commit hook.

ErikBjare:
I've run `cargo fmt` and applied a modified patch from https://github.com/ActivityWatch/aw-server-rust/pull/88/commits/6bd027aeff160eb7330a46c398004cd02f7e831e onto master.

Makefile
Dentosal: The default hook installed from the Makefile should definitely skip this if it takes whole 5 minutes, and CI checks it anyways. I personally prefer always running tests or modified code before pushing, but I can keep my local git hooks like that without affecting others.

Makefile
Dentosal: Currently the project has so many clippy warnings that it's might be unclear that an error has cancelled the commit instead of just having the warnings printed. Maybe the hook here should only test for errors anyways?

Makefile
Dentosal: This is incorrect; `@` only supresses printing of the command, not stdout

Makefile
xylix: I'd guess long build times mean building for release, either by the make build task (which also builds aw-webui which is quite slow if not ready). 

I run `cargo check``cargo run` and `cargo test` on my machine and it is quite fast after the first build.

Dentosal: @xylix: Please note that currently this PR only enables format checking for commits; clippy and tests are only ran pre-push. Still, relying only on CI is certainly possible. However, when working with feature branches or on my own fork, requiring tests to pass before pushing is quite useful.

Makefile
johan-bjareholt: Aha, sorry

Makefile
johan-bjareholt: It's pre-push (not post) so it should still be clear that this is an warning rather than error.

Makefile
johan-bjareholt: I agree that it's useful, but I don't think it's nice to enforce a workflow on users. Especially when it can potentially take a lot of time.

ErikBjare:
I've added CI-checks for `cargo fmt` and a (non-failing) run of `clippy` in #105. With the new improved build times the contributor will get feedback in just a couple of minutes (much faster than building locally from scratch), so shouldn't be too bad.

Proposing close.

johan-bjareholt:
Thanks for submitting this, I have done clippy fixes now for the whole project and you ran rustfmt on the whole codebase so things are a lot better now.

Adding git hooks seems excessive however now when our CI times is so much faster, closing.

PR #91, opened 2020-02-27
Add CORS support to config and default config

xylix:
Adds reading CORS entries from config and writing default CORS entry to config as comment like other config options on first start without config.

### DIFF
--- a/aw-server/src/endpoints/cors.rs
+++ b/aw-server/src/endpoints/cors.rs
@@ -6,9 +6,13 @@ use crate::config::AWConfig;
 
 pub fn cors(config: &AWConfig) -> rocket_cors::Cors {
     let root_url = format!("http://127.0.0.1:{}", config.port).to_string();
-    let mut allowed_exact_origins = vec![root_url];
+    let root_url_localhost = format!("http://localhost:{}", config.port).to_string();
+    let mut allowed_exact_origins = vec![root_url, root_url_localhost];
+    allowed_exact_origins.extend(config.cors.clone());
+
     if config.testing {
         allowed_exact_origins.push("http://127.0.0.1:27180".to_string());
+        allowed_exact_origins.push("http://localhost:27180".to_string());
     }
     let mut allowed_regex_origins = vec![
         "chrome-extension://nglaklhklhcoonedhgnpgddginnjdadi".to_string(),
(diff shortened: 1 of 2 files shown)


aw-server/src/endpoints/cors.rs
```diff
@@ -7,6 +7,8 @@ use crate::config::AWConfig;
 pub fn cors(config: &AWConfig) -> rocket_cors::Cors {
     let root_url = format!("http://127.0.0.1:{}", config.port).to_string();
     let mut allowed_exact_origins = vec![root_url];
+    allowed_exact_origins.extend(config.cors.clone());
+
     if config.testing {
         allowed_exact_origins.push("http://127.0.0.1:27180".to_string());
```
johan-bjareholt: 27180 port is now added twice?

I don't think we need to add it to both the config and hard code it, hard coding it should be enough.

xylix: It's added both from 127.0.0.1 and from localhost. Personally I found this the easiest setup because aw-webui's command line default link is to localhost (`App running at:
  - Local:   http://localhost:27180/`)

Altough maybe the right approach would be for me to prose we change the hard coded one from 127.0.0.1:port to localhost:port. I'll change the default .testing config to an empty vector to remove the unnecessary duplication.

johan-bjareholt: Correct me if I'm wrong, but if the OS properly does address translation shouldn't "localhost" and "127.0.0.1" mean the same thing? I don't think both are necessary.

At least on my computer both 127.0.0.1:5600 and localhost:5600 both work fine even though only 127.0.0.1:5600 is configured. I can even do myhostname:5600 which also works fine if that's configured in your OS.

xylix: On my macOS box I need to have the exact ip / localhost string in my browser to not get 403. I'm currently running the mac build with these changes and http://127.0.0.1:5600 works, http://localhost:5600 causes 403.

Also using my local IP requires a third string with the hard coded local IP in my enabled CORS. 

Should probably test on windows too.

There might also be some rocket method to translate it correctly on operating systems that require it?

johan-bjareholt: Hm, alright.

In that case, hardcode in both 127.0.0.1 and localhost for both config.port and 27180 I guess.

xylix: Done in a34ce90

johan-bjareholt:
Also, feel free to squash your changes, there's no point in 5 mini commits going back and fourth.

Looks good otherwise!

xylix:
Yeah good point. Squashed now.

johan-bjareholt:
Awesome! :)

PR #100, opened 2020-03-22
added RELEASE option to Makefile (for faster non-release builds)

ErikBjare:
Helps us avoid building aw-server-rust twice in CI (first in release mode, and then in debug mode when tests are run).


ErikBjare:
@johan-bjareholt Maybe we should flip this flag around so that it builds in `--release` mode by default?

ErikBjare:
Since there is no change of the defaults, I'll go ahead and merge this.

Issue #102, opened 2020-03-24
aw-server-rust does not accept `--testing` flag


ErikBjare:
Tried running aw-server-rust in v0.9.2 with `--testing` but it didn't use the testing port.

xylix:
Yeah currently you need to run it with `cargo run aw-server` from the aw-server-rust project root to run with testing params. This would be useful to support though.

ErikBjare:
You should get the same result by setting `ROCKET_ENV=development`, but setting it with a flag is what we do in aw-qt so it kinda needs to be supported.

I tried looking at `config.rs` but it looks like we currently use Rocket `Environment` to detect testing/non-testing mode everywhere which makes it messy to change. Might be easy for @johan-bjareholt though who knows the code.

johan-bjareholt:
@ErikBjare The reason why the change would require quite a bit of rework of the code is because rocket Environment is stateless while if we want to provide an option into this code we would have to provide a state (if --testing is set or not) when creating an AwConfig.

I guess it was just a matter of time before we would have to do something like this though, will probably be needed for more CLI options and such in the future as well.

johan-bjareholt:
Solved in #107

PR #104, opened 2020-03-24
reduced logging and simplified logging pipeline

ErikBjare:
I know this will leave ANSI color escape sequences in the logfiles, but dumb to repeat the entire formatting, and if its a big issue there are better ways to do it (but I got tired of looking at fern docs to figure out the way).

### DIFF
--- a/aw-server/src/config.rs
+++ b/aw-server/src/config.rs
@@ -48,22 +48,22 @@ fn default_address() -> String {
 }
 
 fn default_port() -> u16 {
-    match Environment::active().expect("Failed to get current environment") {
-        Environment::Production => 5600,
-        Environment::Development => 5666,
-        Environment::Staging => panic!("Staging environment not supported"),
+    if is_production() {
+        5600
+    } else {
+        5666
     }
 }
 
 fn default_cors() -> Vec<String> {
-    match Environment::active().expect("Failed to get current environment") {
-        Environment::Production => Vec::<String>::new(),
-        Environment::Development => Vec::<String>::new(),
-        Environment::Staging => panic!("Staging environment not supported"),
-    }
+    Vec::<String>::new()
 }
 
-fn is_testing() -> bool {
+pub fn is_production() -> bool {
+    !is_testing()
+}
+
+pub fn is_testing() -> bool {
     match Environment::active().expect("Failed to get current environment") {
         Environment::Production => false,
         Environment::Development => true,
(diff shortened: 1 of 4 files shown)


aw-server/src/endpoints/mod.rs
```diff
@@ -88,24 +89,13 @@ fn get_device_id() -> String {
...
-
     let hostname = gethostname().into_string().unwrap_or("unknown".to_string());
     const VERSION: Option<&'static str> = option_env!("CARGO_PKG_VERSION");
 
     json!({
         "hostname": hostname,
-        "version": format!("aw-server-rust v{}", VERSION.unwrap_or("(unknown)")),
-        "testing": testing,
+        "version": format!("v{} (rust)", VERSION.unwrap_or("(unknown)")),
+        "testing": is_testing,
```
johan-bjareholt: This is not a good solution either as it won't work with an "--testing" option either.

The proper way would be to add it to the ServerState struct and make server_info read that.

aw-server/src/config.rs
```diff
@@ -48,22 +48,22 @@ fn default_address() -> String {
...
-    }
+    Vec::<String>::new()
+}
+
+pub fn is_production() -> bool {
+    !is_testing()
 }
 
-fn is_testing() -> bool {
+pub fn is_testing() -> bool {
```
johan-bjareholt: This should not be public, the info should be shared through the config instead.
This can't be a function without a state once we add the "--testing" option, let's get ownership right already.

aw-server/src/endpoints/mod.rs
ErikBjare: Sure, but as I mentioned in #102 I don't know the best way to do that. So can we merge this in the meanwhile?

aw-server/src/config.rs
ErikBjare: This was supposed to be a small PR, and what you're asking for sounds like a lot of work that I'm not sure how I'd write, so merge and fix that later?

johan-bjareholt: Not too much work if you know how to do it, I can fix it later. This is till better than before!

aw-server/src/endpoints/mod.rs
johan-bjareholt: Sure!

johan-bjareholt:
Let's merge this, then I can more easily start making changes on top of this.

PR #107, opened 2020-03-25
Add cli option parsing, --testing and refactoring

johan-bjareholt:
Solves #104


johan-bjareholt:
Not sure what's going on there in the CI with aw-sync failing, but I'm ignoring that.

Issue #117, opened 2020-05-02
Building cargo-kcov in CI takes >5min


ErikBjare:
Slowing down the build significantly: https://github.com/ActivityWatch/aw-server-rust/pull/89/checks?check_run_id=638369573

Running the tests themselves under coverage also take considerable time (>5min), but I noticed that just building cargo-kcov takes ~5min.

johan-bjareholt:
The coverage is also completely broken, some files say that we for example have 133 lines not hit while the file is only 13 lines long...

johan-bjareholt:
grcov is merged and working, closing

PR #123, opened 2020-06-11
Use grcov instead of kcov

johan-bjareholt:
Wants to solve #117 

TODO:
- [x] Fix codecov.io not being able to read the lcov report
- [x] Install grcov from pre-built instead of building from source via cargo

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,15 +35,37 @@ jobs:
       run: cargo build --workspace --verbose
     - name: Run tests
       run: cargo test --workspace --verbose
-    - name: Run tests with coverage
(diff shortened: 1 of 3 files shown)


johan-bjareholt:
Works really well locally, not sure why codecov.io is not picking up the new coverage data even though it's clearly being run in github actions.

johan-bjareholt:
It's not updating because codecov.io fails to find/parse the lcov report.

johan-bjareholt:
Oops, closed by mistake...

johan-bjareholt:
This can be merged now, the only issue is a cache inconsistency breaking the linux build currently.

PR #143, opened 2020-07-03
fix: If events merge during flood, choose the longest endtime

### LINKED ISSUE #140: Significant differences between reporting of aw-server and aw-server-rust
I looked at my events for a day a month ago, and there are a couple of significant differences between what aw-server-rust and what aw-server reports.

Not sure if it's due to the migration or one/multiple transform issues.

1. Individual application time:

| Application |  aw-server | aw-server-rust |
| --- | --- | --- |
| Firefox       | 48m26s   | 46m38s           |
| Chromium | 38m41s    | 9m31s            |
|Gedit|9m|8m57s|
|Guake|1m|1m35s|
|Unknown|14s|15s|

*Note*: Individual application time seems to vary everywhere, and in all of the tabs.

2. Total Active time

 |  aw-server | aw-server-rust |
| --- | ---- | 
| 1h 38min | 1hr 6 min |

3. Active browser time in Browser tab:

|   aw-server | aw-server-rust |
| ---- | ---- |
|  1 hr 21 min | 52 min |

**Another potential bug**: If you sum up ALL of the times for each browser domain, they won't be equal to the total active browser time for aw-server-rust. This issue doesn't occur for aw-server. 

4. Top categories amounts are also all very different between the 2 servers. This however, is expected given point 1.

4. Timelines look the same(from a high level view)

OS: Fedora 32, Gnome 3.36, Xorg

johan-bjareholt:
Fixes #140

### DIFF
--- a/aw-transform/src/flood.rs
+++ b/aw-transform/src/flood.rs
@@ -8,13 +8,15 @@ pub fn flood(events: Vec<Event>, pulsetime: chrono::Duration) -> Vec<Event> {
     let mut events_sorted = sort_by_timestamp(events);
     let mut e1_iter = events_sorted.drain(..).peekable();
     let mut new_events = Vec::new();
-    let mut drop_next = false;
     let mut gap_prev: Option<chrono::Duration> = None;
-    while let Some(mut e1) = e1_iter.next() {
-        if drop_next {
-            drop_next = false;
-            continue;
+    let mut retry_e: Option<Event> = None;
+    while let Some(mut e1) = match retry_e {
+        Some(e) => {
+            retry_e = None;
+            Some(e)
         }
+        None => e1_iter.next(),
+    } {
         if let Some(gap) = gap_prev {
             e1.timestamp = e1.timestamp - (gap / 2);
             e1.duration = e1.duration + (gap / 2);
@@ -37,7 +39,6 @@ pub fn flood(events: Vec<Event>, pulsetime: chrono::Duration) -> Vec<Event> {
                     warned_negative_gap_safe = true;
                 }
                 // Choose the longest event and set the endtime to it
-                // TODO: Also possibly extend to an e3 if that exists?
                 let e1_endtime = e1.calculate_endtime();
                 let e2_endtime = e2.calculate_endtime();
                 if e2_endtime > e1_endtime {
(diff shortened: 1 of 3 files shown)


aw-transform/src/flood.rs
```diff
@@ -36,8 +36,15 @@ pub fn flood(events: Vec<Event>, pulsetime: chrono::Duration) -> Vec<Event> {
                     warn!("Gap was of negative duration ({}s), but could be safely merged. This error will only show once per batch.", gap);
                     warned_negative_gap_safe = true;
                 }
-                // Extend e1 to the middle between e1 and e2
-                e1.duration = e2.calculate_endtime() - e1.timestamp;
+                // Choose the longest event and set the endtime to it
+                // TODO: Also possibly extend to an e3 if that exists?
```
ErikBjare, line 40: Hmm, I thought the Python version would do this too, but maybe it doesn't?

aw-transform/src/flood.rs
```diff
@@ -126,4 +132,37 @@ mod tests {
         assert_eq!(&res[0], &e1_expected);
         assert_eq!(&res[1], &e2_expected);
     }
+
+    #[test]
+    fn test_flood_same_timestamp() {
+        let e1 = Event {
+            id: None,
+            timestamp: DateTime::from_str("2020-06-16T03:29:11.334000+00:00").unwrap(),
+            duration: Duration::microseconds(305402078),
```
ErikBjare, line 141: It's a bit hard to read these tests when you involve millisecond-precision. Would be nicer with round datetimes.

johan-bjareholt: Sorry, my bad. Was copied straight from a real broken example.
Fixed.

aw-transform/src/flood.rs
johan-bjareholt: I'm not sure if I'm reading the code correctly, but this is how it works I think.

It iterates over the events with zip which creates an array in the form of `[[1,2],[2,3],[3,4]]`, but if it sees a change to merge an event it extends the first event to the end of the second one and sets the duration of the second one to 0 and afterwards filter out all events with an duration of 0. So that means that today it "kind of" works because it will not merge the same event over and over but create multiple merged events even if they have the same data and could be merged, but it would be easy to fix by simply extending the second event instead of the first one so on the next iteration it would have an event which does not have a duration of 0.

johan-bjareholt: In the latest commit I have fixed it.

aw-transform/src/flood.rs
```diff
@@ -36,25 +38,36 @@ pub fn flood(events: Vec<Event>, pulsetime: chrono::Duration) -> Vec<Event> {
...
+                }
                 // Drop next event since they are merged and flooded into e1
-                drop_next = true;
+                e1_iter.next();
+                // Retry this event again to give it a change to merge e1
+                // with 'e3'
+                retry_e = Some(e1);
+                // Since we are retrying on this event we don't want to push it
+                // to the new_events vec
+                continue;
```
ErikBjare, line 56: Cool! Took me a while to understand what's going on with `retry_e`, but looks performant :smiley:

johan-bjareholt: It's mostly to make the borrow checker happy, there are more readable ways to implement it if you write "unsafe" code.

PR #158, opened 2020-08-05
Fix TODOs

johan-bjareholt:
- Make UUID no longer be re-read from file on each /info request
- More testing
- Make all endpoints return a HttpErrorJson which has a "message" field with a descriptive error
    - Can hopefully in the future be used in the web-ui to better describe what goes wrong in case of for example an import without having to look at the server log
    - Will hopefully make the future swagger ui look nice
- Implement the "Into" trait for DatastoreError to HttpErrorJson to be able to remove a lot of boilerplate code

### DIFF
--- a/aw-server/src/endpoints/mod.rs
+++ b/aw-server/src/endpoints/mod.rs
@@ -1,4 +1,3 @@
-use std::fs;
 use std::path::PathBuf;
 use std::sync::Mutex;
 
@@ -6,10 +5,8 @@ use gethostname::gethostname;
 use rocket::response::NamedFile;
 use rocket::State;
 use rocket_contrib::json::JsonValue;
-use uuid::Uuid;
 
 use crate::config::AWConfig;
-use crate::dirs;
 
 #[macro_export]
 macro_rules! endpoints_get_lock {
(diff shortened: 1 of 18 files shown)


aw-server/src/endpoints/mod.rs
```diff
@@ -95,7 +100,7 @@ fn server_info(config: State<AWConfig>) -> JsonValue {
         "hostname": hostname,
         "version": format!("v{} (rust)", VERSION.unwrap_or("(unknown)")),
         "testing": config.testing,
-        "device_id": get_device_id(),
+        "device_id": state.device_id,
```
ErikBjare, line 103: Nice

johan-bjareholt:
@ErikBjare I'm satisfied now, feel free to review if you want. Otherwise I'll merge it this evening.

johan-bjareholt:
@ErikBjare The coverage has been messed up for a while on rust now. Here it fluctuated even though only the Cargo.lock file was updated

https://codecov.io/gh/ActivityWatch/aw-server-rust/commit/03080379695d5e0618d3a4b7f507d7533f38165a

PR #159, opened 2020-08-06
aw-models: Add schemars support

johan-bjareholt:


johan-bjareholt:
I'm having a hard time why it can't find serde and it's respective crates on macOS only. Seems very strange...

ErikBjare:
No idea why that happens, but when searching for it I found: https://www.reddit.com/r/rust/comments/a8ykj8/til_you_dont_need_serde_derive_when_using_the/

Apparently you should just be able to do `use serde::{Serialize, Deserialize}`.

johan-bjareholt:
@ErikBjare Yeah, I also saw that yesterday which is why I removed the "#[macro_use]" and replaced it with use.

I now see in that thread though that it should also be possible to remove serde_derive as a dependency from Cargo.toml and instead add the "derive" feature to serde, will try that.

johan-bjareholt:
Wow, it actually worked!

Will now add a second change which does this for all crates, then I'll merge.

PR #161, opened 2020-08-10
Add swagger for some endpoints

johan-bjareholt:


johan-bjareholt:
@ErikBjare I'll assume that there's something wrong with Rust Nightly because now the same issue also happens with pear and pear_codegen instead of the previous serde and serde_derive.

PR #178, opened 2020-10-17
Make settings value possibly any json type

johan-bjareholt:
Before it was forced to be a string

We should probably discuss this a little further before merging. @ErikBjare and @xylix, what do you think?

Pros:
- Client can send any JSON type and not only string, doesn't have to manually deserialize json within the string if it wants to use some other type than string

Cons:
- The datastore now contains serialized JSON rather than just a string

Fixes #178

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -510,9 +510,9 @@ checksum = "134951f4028bdadb9b84baf4232681efbf277da25144b9b0ad65df75946c422b"
 
 [[package]]
 name = "dyn-clone"
-version = "1.0.2"
+version = "1.0.3"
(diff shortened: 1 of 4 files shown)


xylix:
I think we can go forward with merging this, and if serialized JSON in the datastore makes things confusing we can always remove this feature in the future. The first use will be for string only anyway, if I remember correctly.

ErikBjare (approved):
Looks good, but none of the tests are actually testing other JSON types than strings. Might want to add a simple test case for objects and arrays.

johan-bjareholt:
@ErikBjare Done!

Issue #119, opened 2020-05-23
Add explicit cloning to query2


johan-bjareholt:
Currently every time we reference a variable we are cloning the whole value to guarantee that the value of the variable cannot be modified in some function without re-assigning it.
If we had explicit cloning built-in to the language so cloning is only done when necessary querying would probably be insanely faster.

johan-bjareholt:
I updated this issue with ideas of how this could be solved.

My first impression is that 3 might be the best solution. Could even add some logic to tell a friendly error like "The variable you are referencing was defined before but has been consumed, maybe you want to clone it?"

4 is also a tempting solution, but less sure how that would work. I'm pretty sure it's possible though.

ErikBjare:
I think in the Python version of query2 all functions that modify data do deepcopy's on input arguments in order to avoid mutating them.

>querying would probably be insanely faster

Really? Are clones that expensive? 

Don't have the time to review the best approach right now, but ping me after the 5th of June and I'll look over it.

johan-bjareholt:
@ErikBjare 

> I think in the Python version of query2 all functions that modify data do deepcopy's on input arguments in order to avoid mutating them.

It seems like at some places we use deepcopy on the events and in other places we iterate on the passed events and create new ones with the slight modifications. In some places such as flood it would be possible to optimize because there we use deepcopy when in fact we wouldn't need to if we had these ownership things in place (because most of the time we don't use the non-flooded events after we have called flood).

> Really? Are clones that expensive?

memory allocation and memory copying is expensive and currently we do it a lot of time in query2.

I used the query benchmark I wrote last week and made the simple change that consumes all variable references instead of cloning them (likely breaks aw-webui queries, but not the simple query in the bench).

The query in the bench is the following, so there should only be one clone I believe (when the events are assigned to the events variable). There are 3000 generated events into the "testid" bucket.

    events = query_bucket(\"testid\");
    return events;

These are the results

    # With cloning (not consuming)
    test query_benchmarks::bench_many_events ... bench:   7,237,172 ns/iter (+/- 636,117)
    # Without cloning (consuming)
    test query_benchmarks::bench_many_events ... bench:   5,660,517 ns/iter (+/- 543,986)

|min | avg | max |
|------|-------|--------|
| 6601055 | 7237172 | 7873289 | 
| 5116531 | 5660517 | 6204503 |

max perf increase = 7873289/5116531 = ~53%
average perf increase = 7237172/5660517 = ~27.8%
[...]

PR #179, opened 2020-10-18
aw-transform: Add union_events_split

johan-bjareholt:
One step in fixing the "Merge web watcher events into window events in the query? (to allow for classifying by url/domain)" https://github.com/ActivityWatch/aw-webui/issues/151. We already have it in the same query, but the data is not merged.
Hopefully this will be a good transform to merge window and browser data in that manner.


aw-transform/src/union.rs
```diff
@@ -0,0 +1,178 @@
...
+///     | events2 |    [c     ]    [d ]|
+///     | result  |[a ][ac][bc][b ]    |
+///     |---------|--------------------|
+/// ```
+pub fn union_events_split(events1: Vec<Event>, events2: &Vec<Event>) -> Vec<Event> {
+    let mut events: Vec<Event> = Vec::new();
+
+    'event1: for mut event1 in events1 {
+        let event1_endtime = event1.calculate_endtime();
+        'event2: for event2 in events2 {
```
ErikBjare, line 57: Instead of a nested loop one might want to step through each list under certain conditions (similarly to how we do it in aw-server-python for some transforms: `e1_i++` and `e2_i++`). 

The borrow checker would probably hate that though, and I guess the timestamp checks are pretty fast despite the worst-case `O(N^2)`.

ErikBjare (commented):
Only taken a quick look but looks alright. 

However, as we discussed AFK, I'm not convinced this is the best approach to merging the web events compared to simply replacing all browser-window events when an active browser event exists, but that has its own issues (like having to inject the browser appname, slightly changing the title by removing the appended ` - Mozilla Firefox`, and maybe more).

aw-transform/src/union.rs
johan-bjareholt: I don't think the borrow checker would have any issues with that, you could just use iterators which should work I think.

johan-bjareholt:
> However, as we discussed AFK, I'm not convinced this is the best approach to merging the web events compared to simply replacing all browser-window events when an active browser event exists, but that has its own issues (like having to inject the browser appname, slightly changing the title by removing the appended - Mozilla Firefox, and maybe more).

So like the opposite of filter_period_intersect and then concatenating window and browser events?

ErikBjare:
@johan-bjareholt I think something like:

```py
window_events = query(...)   # query and do AFK filtering

for browser in browsers:  # this for-loop would need to be expressed in the function constructing the query
    web_events = query("aw-watcher-web-$browser")
    window_events_browser = filter_by(window_events, {app: $browserAppname})  # can't remember what transform is used to do this
    web_events_active = filter_period_intersect(browser_events, window_events_browser)`  # to get browser events when the browser is the active window
    window_events = union(web_events_active, window_events)  # picks events from web_events_active first, then fills rest with window_events
```

Not sure if we have something like `union` (in the way I mean it) already, but I mean something that does:

```
"bucket"          | events
---------------------------------------------------------------
web_events_active |           [github  ]           [youtube   ]
window            |[spotify  ][browser ][terminal ][browser   ]
---------------------------------------------------------------
result            |[spotify  ][github  ][terminal ][youtube   ]
```

A complete end-to-end example would be:

```
bucket  | events
-----------------------------------------------
window  | [terminal  ][firefox  ][terminal  ]
browser | [youtube   ][google               ]
-----------------------------------------------
result  | [terminal  ][google   ][terminal  ]
```

johan-bjareholt:
>     web_events_active = amend_data(web_events_active, {app: $browserAppname})  # optional, amends the missing {app: $browserAppname}

I think this would be a strange transform, inserting data into an event which is hardcoded in the query and does not come from a bucket feels like something we should try to avoid. One reason would be because it encourages people to write dynamically generated queries like we do in the web-ui which is something we don't want.

>    window_events = union(web_events_active, window_events)  # picks events from web_events_active first, then fills rest with window_events

To call that "union" is very misleading in my opinion. A union should be a common ground, not exclude anything.

johan-bjareholt:
> However, as we discussed AFK, I'm not convinced this is the best approach to merging the web events compared to simply replacing all browser-window events when an active browser event exists, but that has its own issues (like having to inject the browser appname, slightly changing the title by removing the appended - Mozilla Firefox, and maybe more).

What this transform does is essentially this, but it also injects the browsers appname.

ErikBjare:
>To call that "union" is very misleading in my opinion. A union should be a common ground, not exclude anything.

For sure, I was just unsure what to call it.

>I think this would be a strange transform, inserting data into an event which is hardcoded in the query and does not come from a bucket feels like something we should try to avoid.

I agree in general, but in this case the data *does* come from a bucket (and doesn't actually introduce new hardcoding, since we already have our list of `$browserAppname`). We're reinserting exactly what we know we lost before (with `filter_by` and `filter_period_intersect`).

But on second thought I realize it'll still be messy, due to there being multiple possible appnames for each browser.

>So if we put that into perspective of window events and browser events already filtered with the browser window: `<example>`

I thought it would become:

```
///     |---------|---------------------------|
///     | window  |[firefox   ][terminal     ]|
///     | browser |    [google     ]          |
///     | result  |[f ][f+g   ][t+g][terminal]|
///     |---------|---------------------------|
```

Basically what I'm trying to get rid of splitting events into two.

**Edit:** Ah nvm, your example would indeed become as you wrote after filtering the browser events.

But it would still lead to things like this, no?:

```
/// Abbreviations:
///  - ff: Firefox
///  - t1: Tab 1, with a `title` and corresponding `url`
///  - ff(t1): Firefox window event polled from when `t1` was active
///  - ff(t1)+t2: Firefox event with title from `t1`, but URL from `t2` (after `merge_map`)
///
///     |---------|------------------------------------|
///     | window  |[ff(t1)                ][ff(t2)    ]|
///     | browser |[t1        ][t2                    ]|
///     | result  |[ff(t1)+t1 ][ff(t1)+t2 ][ff(t2)+t2 ]|
///     |---------|------------------------------------|
```

So if I'm not mistaken, this would result in 'middle-events' where the title (which is gotten from the window event, if I understood `merge_map` correctly) and URLs (from the web event) are misaligned. This will happen anytime two events don't perfectly overlap (which is always), leading to a lot of small misaligned events like this.

aw-transform/src/union.rs
```diff
@@ -0,0 +1,178 @@
...
+        let d2 = json_map! {"test": json_map!{"b": json!(2)}};
+        merge_map(&mut d1, &d2);
+        assert_eq!(
+            d1,
+            json_map! {"test": json_map!{"a": json!(1), "b": json!(2)}}
+        );
+
+        /* test non-object conflict, prefer map1 value */
+        // TODO: This does not work yet!
+        //       It should be a pretty rare use-case anyway
```
ErikBjare, line 140: This wouldn't be rare? Both window-events and web-events have a `title`?

aw-transform/src/union.rs
```diff
@@ -0,0 +1,178 @@
...
+    #[test]
+    fn test_union_events_split() {
+        // Test intersection, before and after
+        let e1 = Event {
+            id: None,
+            timestamp: DateTime::from_str("2000-01-01T00:00:00Z").unwrap(),
+            duration: Duration::seconds(3),
+            data: json_map! {"test": json!(1)},
+        };
+        let mut e2 = e1.clone();
```
ErikBjare, line 159: I'd like to have more realistic data here, for example:

For e1: `{app: "firefox", title: "google - firefox"}`
For e2: `{title: "google", url: "google.com"}`

Event series that are at least two events long, like the one I recently commented about, would also be nice (would also help with ensuring behavior stays consistent when we eventually remove the nested loops).

ErikBjare (commented):
Second review, now that I understand things better.

johan-bjareholt:
> So if I'm not mistaken, this would result in 'middle-events' where the title (which is gotten from the window event, if I understood merge_map correctly) and URLs (from the web event) are misaligned. This will happen anytime two events don't perfectly overlap (which is always), leading to a lot of small misaligned events like this.

That's a very good point which I thought of before but forgot.

> Edit 2: I'm not sure, but maybe this could be resolved by using non-flooded window-events for the union_events_split and flood after?

That sounds like a very clever way of solving it, will try that out.

> This might require a lot of extra memory though, since we'd need both window_events (not flooded, for filtering the browser events that go into union_events_split) and window_events_flooded (flooded, for filtering the browser events that go into filter_period_intersect).

I took a deep dive into the query code and it seems like you are correct here. Here's an example of how we would have to change our transforms

```
- events = flood(query_bucket("bucketname"));
+ events_unflooded = query_bucket("bucketname");
+ events_flooded = flood(events_unflooded)
```

The query language is very inefficient with its assignments, every time we assign something it gets cloned every time it's used afterwards because we do not know if the variable will be used afterwards or not and we need to guarantee that the variable won't change. So previously when we just called `flood(query_bucket("bucketname"))` we could just re-use the previous events while when we assign it we need to first keep the value in the variable and then clone it on each reference as well as not release the value of the variable until the whole query is complete. This can be improved in the future by #119.

But I think that the impact of just one more clone would be minimal compared to the whole issue we have today with #119, there are lots of more unnecessary clones than just this.

johan-bjareholt:
Here's a query that works with this transform, the web-ui queries.ts code is really messy though so I'm having a hard time adapting it.

```
{
  "timeperiods": [
    "2020-11-01T01:00:00Z/2020-11-30T01:00:00Z"
  ],
  "query": [
    "window_events = query_bucket(\"aw-watcher-window_johan-laptop2\");",
    "browser_events = query_bucket(\"aw-watcher-web-firefox\");",
    "browser_events = split_url_events(browser_events);",
    "firefox_events = filter_keyvals(window_events, \"app\", [\"firefox\"]);",
    "events = union_events_split(firefox_events, browser_events);",
    "events = merge_events_by_keys(events, [\"$domain\"]);",
    "events = categorize(events, [
        [[\"docs.rs\"], {\"type\": \"regex\", \"regex\": \"^docs.rs$\"}],
        [[\"reddit\"], {\"type\": \"regex\", \"regex\": \"^reddit.com$\"}]
        [[\"crates.io\"], {\"type\": \"regex\", \"regex\": \"^crates.io$\"}]
    ]);",
    "events = sort_by_duration(events);",
    "RETURN = events;"
  ]
}
```

AllanChain:
Hi! Seems that all major problems have been solved in above discussions. What's currently blocking this PR?

PR #185, opened 2020-11-08
feat: added WIP apikey request guard

ErikBjare:


aw-server/src/endpoints/auth.rs
```diff
@@ -0,0 +1,51 @@
...
+        // TODO: How will this key be configured by the user?
+        let config = request.guard::<State<AWConfig>>().unwrap();
+        match &config.apikey {
+            None => Outcome::Success(ApiKey(None)),
+            Some(apikey) => {
+                // TODO: How will this header be set in the browser?
+                let keys: Vec<_> = request.headers().get("x-api-key").collect();
+                match keys.len() {
+                    0 => Outcome::Failure((Status::BadRequest, ApiKeyError::Missing)),
+                    1 if apikey == keys[0] => Outcome::Success(ApiKey(Some(keys[0].to_string()))),
```
johan-bjareholt, line 44: That's a very convoluted way of expressing it.
Better to just have a single case for one and have the of statement inside a block.

johan-bjareholt (commented):
What is this supposed to solve?
Won't the API key be exposed since we are not using HTTPS?

ErikBjare:
@johan-bjareholt I've explained the reasoning for why it's needed in the docstring for `auth.rs`

aw-server/src/endpoints/auth.rs
ErikBjare: Sure, just picked it straight from the example.

johan-bjareholt:
> However, for Android apps it's still an improvement (since I assume they can't snoop on local packets however they like).

Aha, that makes sense.

johan-bjareholt (approved):

PR #186, opened 2020-11-23
fix: correct lookup of asset_path for .app bundles on macOS

ErikBjare:
Needed for https://github.com/ActivityWatch/activitywatch/pull/506


johan-bjareholt (approved):
A cfg to make it only work on macOS wouldn't hurt, but the same goes for some of the Linux paths. This can also be easily fixed later.

Issue #189, opened 2020-11-23
Erratic coverage reports


ErikBjare:
As has been the case for some time now, coverage is erratic.

I tried to figure out the cause, but after looking at the Makefile and CI workflow I'm not finding anything that could be the culprit (I thought it could have something to do with the cache, but now I'm not so sure since the cache keys seem to be correctly set).

Do you still not have any lead @johan-bjareholt?

ErikBjare:
We should probably adopt the recently-announced "source-based coverage": https://blog.rust-lang.org/inside-rust/2020/11/12/source-based-code-coverage.html

Seems like support has landed in grcov.

ErikBjare:
Fixed in #192

PR #192, opened 2020-11-26
ci: switched to source-based coverage

ErikBjare:
Largely based on how grcov themselves do it: https://github.com/mozilla/grcov/blob/master/.travis.yml

From https://github.com/mozilla/grcov/commit/e98d5ddc6570a9f4b6d113b6931fec53357c064b, it seems we can go even further and remove `CARGO_INCREMENTAL` and `RUSTDOCFLAGS`.

Hopefully helps with #189 

Screenshot of HTML report when I run locally. Note that ~~source-based coverage~~ grcov doesn't support branch coverage (I asked them about it in https://github.com/mozilla/grcov/issues/520):

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,6 +58,9 @@ jobs:
         key: ${{ runner.os }}-${{ env.cache-name }}-${{ hashFiles('**/Cargo.lock') }}
         restore-keys: |
           ${{ runner.os }}-${{ env.cache-name }}-
+    - name:  Install llvm-tools
+      run: |
+        rustup component add llvm-tools-preview
     - name: Download grcov
       run: |
         curl -L https://github.com/mozilla/grcov/releases/latest/download/grcov-linux-x86_64.tar.bz2 | tar jxf -
@@ -67,7 +70,7 @@ jobs:
         export PATH=$PATH:.
         make coverage-lcov COVERAGE_CACHE=1
     - name: Upload coverage files
-      run: bash <(curl -s https://codecov.io/bash);
+      run: bash <(curl -s https://codecov.io/bash) -f target/debug/lcov.info;
 
   build-android:
     name: Build for Android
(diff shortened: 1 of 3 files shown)


ErikBjare:
Getting a better error message for the blinking macOS builds:

```
 error[E0514]: found crate `serde_derive` compiled by an incompatible version of rustc which `serde` depends on
 --> aw-models/src/timeinterval.rs:3:5
  |
3 | use serde::de::{self, Deserialize, Deserializer, Visitor};
  |     ^^^^^
  |
  = help: please recompile that crate using this compiler (rustc 1.50.0-nightly (b48cafd9e 2020-11-25))
  = note: the following crate versions were found:
          crate `serde_derive` compiled by rustc 1.50.0-nightly (a0d664bae 2020-11-22): /Users/runner/work/aw-server-rust/aw-server-rust/target/debug/deps/libserde_derive_internals-67951bb09d951f1c.rmeta
          crate `serde_derive` compiled by rustc 1.50.0-nightly (a0d664bae 2020-11-22): /Users/runner/work/aw-server-rust/aw-server-rust/target/debug/deps/libserde_derive_internals-67951bb09d951f1c.rlib

error: aborting due to previous error

error: could not compile `aw-models`
```

I assume we'll have to somehow discard the cache if we're using a new compiler. Weird that cargo/rustc doesn't do this itself.

.github/workflows/build.yml
```diff
@@ -67,7 +70,7 @@ jobs:
         export PATH=$PATH:.
         make coverage-lcov COVERAGE_CACHE=1
```
ErikBjare, line 71: @johan-bjareholt Why is COVERAGE_CACHE set?

ErikBjare:
@johan-bjareholt It seems like coverage reporting is consistent now (at least from the two commits in this PR :crossed_fingers:)... I will go ahead and merge this so I rebase another PR and hopefully get reasonable results.

ErikBjare:
From looking at the rebased #185 it looks like coverage works properly now :tada: 

Hoping it stays that way :crossed_fingers:

PR #193, opened 2020-12-28
ci: set cache key to include rustc_hash

### LINKED ISSUE #180: CI (sometimes) fails on macOS
Sometimes it's this

```
 error[E0463]: can't find crate for `pear_codegen` which `pear` depends on
  --> /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/rocket_http-0.4.5/src/lib.rs:15:14
   |
15 | #[macro_use] extern crate pear;
   |              ^^^^^^^^^^^^^^^^^^ can't find crate

error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
error: could not compile `rocket_http`
```

While other times it's a similar thing but it can't find serde_derive or serde or whatever. I have no idea why and it's even more strange that it only occurs on macOS as it doesn't seem to have to do with anything platform specific.

ErikBjare:
Fixes #180

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,6 +18,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Set up Rust nightly
+      id: toolchain
       uses: actions-rs/toolchain@v1
       with:
         profile: minimal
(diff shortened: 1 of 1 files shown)


ErikBjare:
Weird that there is a codecov diff here, maybe #189 isn't fixed after all?

PR #235, opened 2021-05-10
feat: added period_union transform

### LINKED ISSUE #66: Query features missing
Grammar:
- [ ] Support ending with a semicolon

Functions:
- [ ] find_bucket
- [ ] adjacent, union, period_union (https://github.com/ActivityWatch/aw-core/pull/65)

ErikBjare:
Needed for new visualizations coming in v0.11

Partially fixes #66


ErikBjare:
@johan-bjareholt I think this is all that's needed for the new aw-webui query to work, but haven't tested it.

aw-models/src/event.rs
```diff
@@ -34,6 +35,9 @@ impl Event {
         self.timestamp
             + chrono::Duration::nanoseconds(self.duration.num_nanoseconds().unwrap() as i64)
     }
+    pub fn interval(&self) -> TimeInterval {
+        TimeInterval::new(self.timestamp, self.calculate_endtime())
```
johan-bjareholt, line 39: I still have not found out why, but it calculating endtime is incredibly slow.
It shouldn't be a problem right now, just something that I found out when trying out "yearly" view and benchmarking. Was able to speed up some queries by 50% by reducing the amount of calculate_endtime().

johan-bjareholt (approved):
Looks good!

Just one thing, but that was more of a note than a comment.

PR #89, opened 2020-02-24
Sync stuff

ErikBjare:
Resuming work from #71 

What's missing:

 - [x] "2-way" test with actual folder 
 - [ ] Test for when events have been modified/deleted (when should things get synced?) 
 - [x] Use device IDs to identify remotes
    - Partially done, but can probably be done better.
 - [ ] refactor (some things can be broken down/reused)
 - [ ] better error handling (proper aw-client-rust errors, get rid of unwraps) 
 - [ ] more comprehensive tests (should be >90% covered)  
 - [ ] review processes/conventions
 - [ ] See comments in #71 and fix them
 - [ ] anything else?

PR #237, opened 2021-05-15
feat: added query parameters to aw-client-rust's get_events

ErikBjare:
Also fixed clippy complaints


ErikBjare:
Ugh, I think this was already done in #89

PR #242, opened 2021-06-08
feat: added hostname filtering support to find_bucket

ErikBjare:
This functionality was available in aw-core. 

Needed for new queries which don't assume a certain bucket ID containing the hostname.

Also needed for my PR in https://github.com/ErikBjare/quantifiedme/pull/4


ErikBjare:
@johan-bjareholt I'm going to go ahead and merge this. But please review, there might very well be better ways to do things.

PR #250, opened 2021-10-21
fix: added host header check (to protect against DNS rebinding attacks)

ErikBjare:
@johan-bjareholt Could you take a look at this and finish it? The Python-implementation is here: https://github.com/ActivityWatch/aw-server/commit/3e8731fd70fbf05d842ec73b86533fd04f17c606

Fixes https://github.com/ActivityWatch/activitywatch/security/advisories/GHSA-v9fg-6g9j-h4x4

### DIFF
--- a/aw-server/src/endpoints/hostcheck.rs
+++ b/aw-server/src/endpoints/hostcheck.rs
@@ -13,6 +13,7 @@ use rocket::http::{Method, Status};
 use rocket::{Data, Request, Rocket, Route};
 
 use crate::config::AWConfig;
+use crate::endpoints::HttpErrorJson;
 
 static FAIRING_ROUTE_BASE: &str = "/checkheader_fairing";
 
@@ -29,8 +30,9 @@ impl HostCheck {
 }
 
 /// Route for HostCheck Fairing error
-fn fairing_error_route<'r>(_request: &'r Request<'_>, _: Data) -> Outcome<'r> {
-    Outcome::Failure(Status::BadRequest)
+fn fairing_error_route<'r>(req: &'r Request<'_>, _: Data) -> Outcome<'r> {
+    let err = HttpErrorJson::new(Status::BadRequest, "Host header is invalid".to_string());
+    Outcome::from(req, err)
 }
 
 /// Create a new `Route` for Fairing handling
--- a/aw-server/src/endpoints/hostcheck.rs
+++ b/aw-server/src/endpoints/hostcheck.rs
@@ -89,7 +89,7 @@ impl Fairing for HostCheck {
             return;
         }
 
-        // Deny requests from hosts that are not localhost
+        // Deny requests to hosts that are not localhost
         let valid_hosts: Vec<&str> = vec!["127.0.0.1", "localhost"];
         let host = host_opt.unwrap();
         if !valid_hosts.contains(&host) {
@@ -97,7 +97,7 @@ impl Fairing for HostCheck {
             redirect_bad_request(request);
         }
 
-        // host is verified, proceed with request
+        // host header is verified, proceed with request
     }
 }
 
--- a/aw-server/src/endpoints/hostcheck.rs
+++ b/aw-server/src/endpoints/hostcheck.rs
@@ -1,0 +1,45 @@
+/// Host header check needs to be performed to protect against DNS poisoning attacks.
+///
+/// Based on API key PR in [2].
+///
+/// [1]: https://github.com/ActivityWatch/activitywatch/security/advisories/GHSA-v9fg-6g9j-h4x4
+/// [2]: https://github.com/ActivityWatch/aw-server-rust/pull/185
+use rocket::http::Status;
+use rocket::request::{self, FromRequest, Request};
+use rocket::{Outcome, State};
+
+use crate::config::AWConfig;
+
+pub struct HostCheck();
+
+#[derive(Debug)]
+pub enum HostCheckError {
+    Invalid,
+}
+
+// TODO: Should this be an app-wide fairing instead? (apparently fairings can't cancel/reject requests?)
+// TODO: Use guard on any remaining sensitive endpoints
+// TODO: Add tests to ensure enforced
+impl<'a, 'r> FromRequest<'a, 'r> for HostCheck {
+    type Error = HostCheckError;
+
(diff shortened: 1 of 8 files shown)


aw-server/src/endpoints/hostcheck.rs
```diff
@@ -0,0 +1,45 @@
...
+use crate::config::AWConfig;
+
+pub struct HostCheck();
+
+#[derive(Debug)]
+pub enum HostCheckError {
+    Invalid,
+}
+
+// TODO: Should this be an app-wide fairing instead? (apparently fairings can't cancel/reject requests?)
```
johan-bjareholt, line 20: I also thought that fairings couldn't cancel/reject, but then I remembered that rocket_cors does this.
Seems to be doable when looking at their source code which looks quite simple, will give that an attempt.

aw-server/src/endpoints/hostcheck.rs
```diff
@@ -0,0 +1,45 @@
...
+// TODO: Use guard on any remaining sensitive endpoints
```
johan-bjareholt, line 21: If we use a fairing it will be applied to every endpoint, regardless if it is sensitive or not.

johan-bjareholt:
@ErikBjare Fixed a fairing, will add tests soon.
Just pushed my commit on top of yours, probably a good idea to squash them when merging.

johan-bjareholt:
Tests are now added as well!

aw-server/src/endpoints/hostcheck.rs
ErikBjare: That's the correct behavior anyway.

aw-server/tests/api.rs
```diff
@@ -239,6 +253,7 @@ mod api_tests {
         res = client
             .get("/api/0/buckets/id/events")
             .header(ContentType::JSON)
+            .header(Header::new("Host", "127.0.0.1:5600"))
```
ErikBjare, line 256: You needed to add this explicitly? The `Host` header should be set correctly automatically, and you shouldn't be testing against port 5600, right?

ErikBjare (commented):
I don't think all the `client.header(Header::new("Host", "127.0.0.1:5600"))...` should be needed, otherwise looks good!

aw-server/tests/api.rs
ErikBjare: Tried removing them, which caused the tests to break.

aw-server/src/endpoints/hostcheck.rs
```diff
@@ -0,0 +1,182 @@
...
+        // We only validate requests if the server binds a local address
+        let validate = config.address == "127.0.0.1" || config.address == "localhost";
+        HostCheck { validate }
+    }
+}
+
+/// Route for HostCheck Fairing error
+fn fairing_error_route<'r>(_request: &'r Request<'_>, _: Data) -> Outcome<'r> {
+    Outcome::Failure(Status::BadRequest)
+}
```
ErikBjare, line 34: Any chance we could have a more informative error in this particular case of BadRequest? Just getting a BadRequest without any indication of why went wrong is kinda bad dev UX.

aw-server/tests/api.rs
johan-bjareholt: >  and you shouldn't be testing against port 5600, right?

The port or address actually doesn't matter. We are never starting an actual server or starting any ports in these unit tests, they are simply injected into Rocket as if they were real requests.

> The Host header should be set correctly automatically

Since the testing framework is a very basic I'll assume that they either simply have not seen this issue before (since the requests are internal to rocket, the Host header is technically pointless) or simply because the Rocket developers find it to make more sense for all headers to be explicit rather than implicit.

aw-server/src/endpoints/hostcheck.rs
johan-bjareholt: Done

aw-server/tests/api.rs
ErikBjare: Ah, I suspected something like that, but found it a bit weird and couldn't quite explain why it would work that way when the `Host` header is so well specified and ubiquitous.

>or simply because the Rocket developers find it to make more sense for all headers to be explicit rather than implicit.

This would seem odd to me in any other context where you make requests, but perhaps it makes sense for tests like these.

ErikBjare:
@johan-bjareholt According to codecov the tests don't cover the failing cases, any idea why?

See here: https://app.codecov.io/gh/ActivityWatch/aw-server-rust/compare/250/tree/aw-server/src/endpoints/hostcheck.rs

aw-server/src/endpoints/hostcheck.rs
```diff
@@ -0,0 +1,184 @@
...
+        // hostname contains port, which we don't care about and filter out
+        let hostheader = hostheader_opt.unwrap();
+        let host_opt = hostheader.split(":").next();
+        if host_opt.is_none() {
+            info!("Host header '{}' not allowed, denying request", hostheader);
+            redirect_bad_request(request);
+            return;
+        }
+
+        // Deny requests from hosts that are not localhost
```
ErikBjare, line 92:
```suggestion
        // Deny requests to hosts that are not localhost
```

aw-server/src/endpoints/hostcheck.rs
```diff
@@ -0,0 +1,184 @@
...
+        let valid_hosts: Vec<&str> = vec!["127.0.0.1", "localhost"];
+        let host = host_opt.unwrap();
+        if !valid_hosts.contains(&host) {
+            info!("Host header '{}' not allowed, denying request", hostheader);
+            redirect_bad_request(request);
+        }
+
+        // host is verified, proceed with request
```
ErikBjare, line 100:
```suggestion
        // host header is verified, proceed with request
```

ErikBjare (commented):
Some comment nits

johan-bjareholt:
> According to codecov the tests don't cover the failing request handlers, any idea why?
> Any chance there's another cause for the BadRequest that you are checking for in the tests? Perhaps check if the returned body of the request actually contains a mention of the failing host-header check?

Simply adding a panic! in the fairing_error_route makes the unittest fail, so the tests work as they should.
I think there's still something broken with our coverage tests, I've been trying to find that out for an hour no without much success.

ErikBjare:
Alright, ready to merge then. Thanks!

PR #262, opened 2022-01-25
add cmdline arg and change the `XDG_DATA_DIRS`

Y7n05h:
feat: Add cmdline arg which can override asset path
feat: fallback to `/usr/local/share:/usr/share` when `XDG_DATA_DIRS` is not set or empty

References:
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

### DIFF
--- a/aw-server/src/main.rs
+++ b/aw-server/src/main.rs
@@ -29,6 +29,9 @@ struct Opts {
     /// Path to database override
     #[clap(long)]
     dbpath: Option<String>,
+    /// Path to webui override
+    #[clap(long)]
+    webpath: Option<String>,
     /// Device ID override
     #[clap(long)]
     device_id: Option<String>,
@@ -77,7 +80,10 @@ async fn main() -> Result<(), rocket::Error> {
     };
     info!("Using DB at path {:?}", db_path);
 
-    let asset_path = get_asset_path();
+    let asset_path = match opts.webpath {
+        Some(webpath) => PathBuf::from(webpath),
+        None => get_asset_path(),
+    };
     info!("Using aw-webui assets at path {:?}", asset_path);
 
     let legacy_import = !opts.no_legacy_import;
@@ -99,6 +105,7 @@ async fn main() -> Result<(), rocket::Error> {
     endpoints::build_rocket(server_state, config).launch().await
 }
 
+use std::ffi::OsString;
 use std::path::PathBuf;
 
 // The appdirs implementation of site_data_dir is broken on computers which has flatpak installed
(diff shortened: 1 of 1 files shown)


aw-server/src/main.rs
```diff
@@ -110,16 +117,26 @@ use std::path::PathBuf;
 // don't want this change?
 fn site_data_dir(app: Option<&str>, _: Option<&str>) -> Result<PathBuf, ()> {
     // Iterate over all XDG_DATA_DIRS and return first match that exists
-    if let Some(joined) = env::var_os("XDG_DATA_DIRS") {
-        for mut data_dir in env::split_paths(&joined) {
-            if let Some(app) = app {
-                data_dir.push(app);
+    let joined = match env::var_os("XDG_DATA_DIRS") {
+        // If $XDG_DATA_DIRS is either not set or empty, a value equal to /usr/local/share/:/usr/share/ should be used.
```
ErikBjare, line 121:
```suggestion
        // If $XDG_DATA_DIRS is either not set or empty, a value equal to /usr/local/share/:/usr/share/ should be used.
        // https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
```

aw-server/src/main.rs
```diff
@@ -110,16 +117,26 @@ use std::path::PathBuf;
...
+        Some(path) => {
+            if path.is_empty() {
```
ErikBjare, line 123: Doesn't look like this allows several values in `XDG_DATA_DIRS`? No biggie, but technically not up to spec so might break if users have `XDG_DATA_DIRS` set to several paths.

Should also probably log a warning message if the path set by the env var is being ignored.

ErikBjare:
I wonder why CI isn't running?

Y7n05h:
> I wonder why CI isn't running?


```
When an outside contributor submits a pull request to a public repository, a maintainer with write access may need to approve any workflow runs.
```

https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

aw-server/src/main.rs
Y7n05h: I'm sorry, I didn't understand what you meant. [Here](https://github.com/ActivityWatch/aw-server-rust/pull/262/files/fb2d067ccf3255429a55c145c764975b8c5eb2d4#diff-39252f270b66dc7653d440ab41e5f0050cd2f1abaf3387dafa033be75cf22b24R132), we traverse all paths contained in XDG_DATA_DIRS.
I tested it on my computer and the program works correctly when XDG_DATA_DIRS contains multiple paths.

ErikBjare: Ah, I was confused about what happened there. I thought `is_empty()` meant if the *directory* was empty. All good then!

PR #270, opened 2022-03-20
fix: improving flooding

ErikBjare:
Changes the behavior of aw-server-rust flooding to work the same as flooding in aw-core.

Part of the issue is described here: https://github.com/ActivityWatch/activitywatch/issues/602#issuecomment-970240728

 - Better tests
 - ~~Forward flooding instead of meet-in-middle (aw-core uses flood-from-longest though)~~
   - Tbh I'm not sure what is best, but forward flooding feels like it introduces *some* duration-bias towards later events, which seems unfair.
   - Flood-from-longest will lead to the lowest relative error, I guess, and they have some sort of higher probability to occur anyway.

## TODO

 - More tests (goal: ~100% coverage for `flood.rs`)

### DIFF
--- a/aw-transform/src/flood.rs
+++ b/aw-transform/src/flood.rs
@@ -2,9 +2,14 @@ use aw_models::Event;
 
 use crate::sort_by_timestamp;
 
-/// Floods event to the nearest neighbouring event if within the specified pulsetime
+/// Floods event to the nearest neighbouring event if within the specified pulsetime.
 ///
-/// Also merges events if they have the same data and are within the pulsetime
+/// Also merges events if they have the same data and are within the pulsetime.
+///
+/// "Flooding" actually performs two similar, but slightly different tasks, which we can describe as:
+///  - Flooding: removes small gaps between events, by extending nearby events to fill the space.
+///  - Merging: merging adjacent or overlapping events that have the same data.
+///     - This is sometimes done in conjunction with flooding, if events with nearby events have the same data.
 ///
 /// Python implementation:
 ///     https://github.com/ActivityWatch/aw-core/blob/master/aw_transform/flood.py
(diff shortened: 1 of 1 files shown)


aw-transform/src/flood.rs
```diff
@@ -65,14 +135,15 @@ pub fn flood(events: Vec<Event>, pulsetime: chrono::Duration) -> Vec<Event> {
...
             } else {
-                if chrono::Duration::seconds(0) > gap {
+                if gap < chrono::Duration::seconds(0) {
                     if !warned_negative_gap_unsafe {
                         warn!("Gap was of negative duration ({}s) and could NOT be safely merged. This error will only show once per batch.", gap);
                         warned_negative_gap_unsafe = true;
                     }
+                } else {
+                    // Extend e1 to the middle between e1 and e2
+                    e1.duration = e1.duration + (gap / 2);
```
ErikBjare, line 145: This was a major issue, as if the gap is negative that could lead to e1.duration getting *shorter*.

aw-transform/src/flood.rs
```diff
@@ -65,14 +135,15 @@ pub fn flood(events: Vec<Event>, pulsetime: chrono::Duration) -> Vec<Event> {
...
                 }
-                // Extend e1 to the middle between e1 and e2
-                e1.duration = e1.duration + (gap / 2);
```
ErikBjare, line 75: This was a major issue, as if the gap is negative that could lead to e1.duration getting shorter.

aw-transform/src/flood.rs
```diff
@@ -65,14 +135,15 @@ pub fn flood(events: Vec<Event>, pulsetime: chrono::Duration) -> Vec<Event> {
...
                 // Make sure next event is pre-extended
                 gap_prev = Some(gap);
```
ErikBjare, line 148: Should probably move this too...

aw-transform/src/flood.rs
```diff
@@ -6,20 +6,45 @@ use crate::sort_by_timestamp;
...
 /// input:  [a] [a]  [b][b]    [b][c]
 /// output: [a     ][b    ]    [b][c]
 /// ```
+/// Example with
+///
+/// ```ignore
+/// pulsetime: 1 second (one space)
+/// input:  [a] [b] [c]
+/// output: [a ][b ][c]
+/// ```
```
ErikBjare, line 33: Redundant, add better example?

ErikBjare:
Hoping coverage will improve with that last commit...

ErikBjare:
Can't figure out why coverage is how it is (more should be covered than is reported? or am I just dumb?), and unable to generate reports locally...

@johan-bjareholt could you take a look?

aw-transform/src/flood.rs
```diff
@@ -40,23 +67,78 @@ pub fn flood(events: Vec<Event>, pulsetime: chrono::Duration) -> Vec<Event> {
...
+            //         )
+            //     )
+            //     warned_about_negative_gap_unsafe = True
+
+            // Ensure that gap is actually non-negative here, at least in tests
+            debug_assert!(gap >= chrono::Duration::seconds(0));
+
+            // If data is the same, we should merge them.
+            if e1.data == e2.data {
                 // Choose the longest event and set the endtime to it
```
ErikBjare, line 136:
```suggestion
                // Set the endtime of e1 to the last endtime of [e1, e2]
```

aw-transform/src/flood.rs
johan-bjareholt: I guess that's true, but if you have negative gaps the data is already unreliable as it is. So there's not much of an improvement, since the raw data is already "bad".

ErikBjare: It's intended to fix this: https://github.com/ActivityWatch/activitywatch/issues/602

Which I consider a significant improvement, as now it should handle it as well as aw-server-python.

aw-transform/src/flood.rs
ErikBjare: >very unlikely scenario

Unfortunately not too unlikely. Not sure if they appear due to some faulty queuing, heartbeat logic, or race condition, but they pop up every now and then...

ErikBjare:
@johan-bjareholt My best theory is that Rust somehow optimizes away a codepath (it notably shows lines that are later duplicated in a different if-clause as "oncovered") and that leads to it now showing in the coverage?

Anyway, would be really nice to have coverage working on the long run. Spent a couple hours trying to get it working better yesterday, but not much luck.

Also, you should check out https://github.com/ActivityWatch/activitywatch/issues/724, might provide better background for some of the things I'm trying to make more robust.

ErikBjare:
According to https://shift.click/blog/github-actions-rust/#coverage-with-cargo-tarpaulin, [cargo-tarpaulin] should be a lot more reliable.

Not LLVM-based, nor supports branch coverage, but might still be better than the status quo... (where entire tests seem to not show up in the coverage)

ErikBjare:
Anyway, merging this. Might open another PR to switch to tarpaulin.

PR #291, opened 2022-07-06
chore(docs): document makefile build and venv usage

iloveitaly:


ErikBjare:
Nice, was actually wondering how I'd do exactly this yesterday :sweat_smile:

PR #301, opened 2022-09-03
feat: add custom visualization support (custom_static)

LockBlock-dev:
I struggled a bit with this since it is my first time with Rust but here it is. I used the FileServer from Rocket because that is the easiest way I found to get custom HTML per watcher.

How to use:
Create a `visualizations` folder in your [data](https://docs.activitywatch.net/en/latest/directories.html#data) folder. Then create a folder for your custom watcher and put your index.html here. Example: `/home/USER/.local/share/activitywatch/aw-server-rust/visualizations/aw-watcher-test/index.html`
You can also load JavaScript from other files, just put them in the same directory.

I tested it with my custom watcher and it works with [custom visualization support](https://github.com/ActivityWatch/aw-server/pull/83) of aw-server web-ui (and so aw-server-rust).

### DIFF
--- a/aw-server/src/main.rs
+++ b/aw-server/src/main.rs
@@ -39,6 +39,10 @@ struct Opts {
     #[clap(long)]
     webpath: Option<String>,
 
+    /// Mapping of custom static paths to serve, in the format: watcher1=/path,watcher2=/path2
+    #[clap(long)]
+    custom_static: Option<String>,
+
     /// Device ID override
     #[clap(long)]
     device_id: Option<String>,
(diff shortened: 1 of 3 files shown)


ErikBjare:
Nice, had no idea it was that easy!

Just a couple thoughts though: this behavior diverges slightly from aw-server-python, where you explicitly need to config the directory `/pages/<custom vis>` -> `/home/user/somewhere/custom_vis` in the `config.toml`.

I want to keep things consistent, so I think aw-server-rust should work the same way.

If you have time to address this, I'd be very happy. Otherwise, I'll find time for it a bit later :)

ErikBjare:
Seems like CI agrees with me (crashes if dir doesn't exist) ;)

LockBlock-dev:
Ah yes I did not implement any kind of check my bad. Also, I tried to mimic the aw-server-python implementation but failed due to my lack of Rust knowledge... I guess you'll have to implement this when you have time.

ErikBjare:
Alright, thanks for your work so far @LockBlock-dev! :heart: 

I'll try and get around to it soonish :)

ErikBjare:
Finally got around to this! Seems to be working now :)

aw-server/src/main.rs
```diff
@@ -75,6 +79,20 @@ async fn main() -> Result<(), rocket::Error> {
...
+        let custom_static_map: std::collections::HashMap<String, String> = custom_static_str
+            .split(',')
+            .map(|s| {
+                let mut split = s.split('=');
+                let key = split.next().unwrap().to_string();
+                let value = split.next().unwrap().to_string();
+                (key, value)
+            })
+            .collect();
+        config.custom_static = custom_static_map;
```
ErikBjare, line 93: Maybe join them here instead of overriding all?

ErikBjare:
Looks like it's working! Merging.

Thanks for getting this work started @LockBlock-dev! ❤️

PR #304, opened 2022-10-01
Source log level from ENV

iloveitaly:
- Pull log level from LOG_LEVEL env var
- Log when assets cannot be found


iloveitaly:
@ErikBjare reminder on this one!

ErikBjare:
@iloveitaly Merged!

PR #310, opened 2022-10-28
Extra debug logging when in testing mode

ShootingKing-AM:
- Dont filter out Rocket messages when in Debug or Trace log level
- also fixes inverted testing condition when config

Eg logs. - https://github.com/ActivityWatch/aw-webui/actions/runs/3341509927/jobs/5532792619#step:15:16

### DIFF
--- a/aw-server/src/logging.rs
+++ b/aw-server/src/logging.rs
@@ -44,65 +44,43 @@ pub fn setup_logger(testing: bool) -> Result<(), fern::InitError> {
         _ => default_log_level,
     };
 
-    match log_level {
-        log::LevelFilter::Debug | log::LevelFilter::Trace => fern::Dispatch::new()
-            // Set some Rocket messages to debug level
-            .level(log_level)
-            .format(move |out, message, record| {
-                out.finish(format_args!(
-                    "[{}][{}][{}]: {}",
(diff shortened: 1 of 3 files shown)


aw-server/src/logging.rs
```diff
@@ -44,37 +44,65 @@ pub fn setup_logger(testing: bool) -> Result<(), fern::InitError> {
...
+                        out.finish(format_args!(
+                            // TODO: Strip color info
+                            "{}",
+                            message,
+                        ))
+                    })
+                    .chain(fern::log_file(logfile_path)?),
+            )
+            .apply()?,
+    }
```
ErikBjare, line 105: This is a lot of code duplication, can we refactor this to improve that?

If not, we can add a TODO comment suggesting doing so later and merge it for now.

ErikBjare: I fixed it :)

ErikBjare:
Fixed some code duplication, ready to merge when I see CI green!

ShootingKing-AM:
Aah! I felt that there is code duplication but dunno why i didnt think about breaking the chain ^^' Thanks !

But you inverted the logging condtion :3 its better to be more verbose if its LogLevel >= Debug, hence no need to filter out any messages from modules if LogLevel >= Debug - #319

PR #319, opened 2022-11-03
fix: only filter when not in debug mode

ShootingKing-AM:
its better to be more verbose if its LogLevel >= Debug, hence no need to filter out any messages from modules if LogLevel >= Debug


ErikBjare:
Ahhh, woops! 😅

Thanks for the fix!

PR #255, opened 2021-11-14
Update rocket to 0.5.0-rc1

johan-bjareholt:

PR #323, opened 2022-11-05
fix: several android fixes

ShootingKing-AM:
- the previous android fix to rocket update was not functional #255
- previous version rocket was async, newer version needs to be awaited
- reaf'ed code a little

### DIFF
--- a/aw-server/Cargo.toml
+++ b/aw-server/Cargo.toml
@@ -46,4 +46,3 @@ jni = { version = "0.19", default-features = false }
 libc = "0.2"
 android_logger = "0.11"
 openssl-sys = { version = "0.9.58", features = ["vendored"]}
-once_cell = "1.16.0"
--- a/aw-server/Cargo.toml
+++ b/aw-server/Cargo.toml
@@ -46,3 +46,4 @@ jni = { version = "0.19", default-features = false }
 libc = "0.2"
 android_logger = "0.11"
 openssl-sys = { version = "0.9.58", features = ["vendored"]}
+once_cell = "1.16.0"
--- a/aw-server/src/android/mod.rs
+++ b/aw-server/src/android/mod.rs
@@ -37,17 +37,17 @@ pub mod android {
     use self::jni::JNIEnv;
     use super::*;
 
+    use once_cell::sync::Lazy;
     use std::path::PathBuf;
     use std::sync::RwLock;
-    use once_cell::sync::Lazy;
 
+    use crate::config::AWConfig;
+    use crate::endpoints;
+    use crate::endpoints::ServerState;
     use android_logger::FilterBuilder;
-    use log::LevelFilter;
     use aw_datastore::Datastore;
     use aw_models::{Bucket, Event};
-    use crate::endpoints;
-    use crate::config::AWConfig;
-    use crate::endpoints::ServerState;
+    use log::LevelFilter;
 
     static mut DATASTORE: Option<Datastore> = None;
 
@@ -102,7 +102,7 @@ pub mod android {
         let mut obj = json!({ "error": &msg });
         string_to_jstring(&env, obj.to_string())
     }
-    
+
     lazy_static! {
         static ref ASSET_PATH: RwLock<String> = RwLock::new(String::new());
     }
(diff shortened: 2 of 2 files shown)


aw-server/src/android/mod.rs
```diff
@@ -93,33 +103,50 @@ pub mod android {
...
+    }
 
-        let server_state = endpoints::ServerState {
-            datastore: Mutex::new(openDatastore()),
-            asset_path: PathBuf::from(asset_path),
-            device_id: device_id::get_device_id(),
-        };
+    #[rocket::main]
+    async fn start_server() {
+        unsafe {
```
ErikBjare, line 128: Why add `unsafe` to the entire block?

ShootingKing-AM: I think only the first info line is safe, can commit to remove the first line, just threw all in together for now

aw-server/src/android/mod.rs
```diff
@@ -93,33 +103,49 @@ pub mod android {
...
+        info!("Server exited");
+    }
 
+    #[rocket::main]
+    async fn start_server() {
         info!("Building server state...");
 
-        let asset_path = jstring_to_string(&env, java_asset_path);
-        info!("Using asset dir: {}", asset_path);
+        unsafe {
```
ErikBjare, line 130: I don't really understand why unsafe is needed here at all? Asked @johan-bjareholt and he was confused too.

```suggestion
        // FIXME: Why is unsafe needed here? Can we get rid of it?
        unsafe {
```

Maybe this inline comment would be good enough for now?

ShootingKing-AM: @ErikBjare its because the openDatastore() is marked unsafe, hence it spirals down to other vars

aw-server/src/android/mod.rs
```diff
@@ -36,8 +36,18 @@ pub mod android {
     use self::jni::sys::{jdouble, jint, jstring};
     use self::jni::JNIEnv;
     use super::*;
+
+    use once_cell::sync::Lazy;
```
ErikBjare, line 40: Unused?

aw-server/Cargo.toml
```diff
@@ -46,3 +46,4 @@ jni = { version = "0.19", default-features = false }
 libc = "0.2"
 android_logger = "0.11"
 openssl-sys = { version = "0.9.58", features = ["vendored"]}
+once_cell = "1.16.0"
```
ErikBjare, line 49: Unused?

aw-server/src/android/mod.rs
```diff
@@ -93,33 +103,50 @@ pub mod android {
         string_to_jstring(&env, obj.to_string())
     }
 
+    lazy_static! {
+        static ref ASSET_PATH: RwLock<String> = RwLock::new(String::new());
```
ErikBjare, line 107: Why not simply use a Mutex here?

https://stackoverflow.com/questions/50704279/when-or-why-should-i-use-a-mutex-over-an-rwlock

aw-server/src/android/mod.rs
```diff
@@ -36,8 +36,18 @@ pub mod android {
...
     use super::*;
+
+    use once_cell::sync::Lazy;
+    use std::path::PathBuf;
+    use std::sync::RwLock;
+
+    use crate::config::AWConfig;
+    use crate::endpoints;
+    use crate::endpoints::ServerState;
+    use android_logger::FilterBuilder;
```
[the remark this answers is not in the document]
ShootingKing-AM: Oops ! I thought cargo check/clippy at least prettier would warn me about unused imports ^^" i guess not :3

aw-server/src/android/mod.rs
ShootingKing-AM: Didn't know much difference about those until you gave the ref; thanks ! Thought rwlock is rust's super safe method better than mutex, guess there are some tradeoffs :3

aw-server/src/android/mod.rs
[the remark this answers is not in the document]
ErikBjare: So did I actually, but apparently tricky: https://github.com/rust-lang/rust/issues/47888

aw-server/src/android/mod.rs
ShootingKing-AM: Still didnt understand the difference 😕 but just impl'ed a simple mutex

ErikBjare:
Going to have a look after #325 is merged.

ErikBjare:
Merged. Thank you very much Sir! 😄

ShootingKing-AM:
Thank you too Sir ! :D

PR #325, opened 2022-11-24
build: downgrade NDK to r21e and added macOS Android build support

ErikBjare:
We briefly had working builds for r25 (https://github.com/ActivityWatch/aw-server-rust/commit/6f1d5b2242ddc7c9cd17580ae4c556a8c210ca55), but they started failing with a weird:

```
  = note: ld: error: unable to find library -lgcc
          clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
```

This change works locally for me (on macOS), hoping it fixes CI too.

### DIFF
--- a/compile-android.sh
+++ b/compile-android.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -e
+platform="$(uname -s | tr '[:upper:]' '[:lower:]')"
 
 if [ -z "$ANDROID_NDK_HOME" ]; then
     if [ -d `pwd`/"NDK" ]; then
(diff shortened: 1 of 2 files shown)


ShootingKing-AM:
Lets update the ndk at a later moment, lgcc is removed in newer version of NDK lgcc exports are available in lunwind but that is not known to upstream crates (also this is specific to android-ndk-toolkit) (and many more changes need to done if we update ndk now)

The fix is making a file gcc.a file redirect to lunwind internally as mentioned in this comment https://github.com/rust-lang/rust/pull/85806#issuecomment-1096266946

aw-android(and android relevant in aw-server-rust) are little old in updates, we can do this update later (have to update ci etc etc..). But I feel its better to make a working aw-android build with existing code with minimal changes as in https://github.com/ActivityWatch/aw-server-rust/pull/323 and https://github.com/ActivityWatch/aw-android/pull/85; then we can pull remaining dependencies including ndk to latest.

ErikBjare:
@ShootingKing-AM That is great info, thanks!

Working on https://github.com/ActivityWatch/aw-android/pull/87, which should hopefully give a working build, and then we can then use your E2E PR to test it fully :)

PR #331, opened 2022-12-16
fix(aw-client-rust): make create_bucket_simple useful

hapm:
create_bucket_simple should expose the created bucket for later use. This would allow to expose the id, which can be generated based on the bucket name given and the hostname, what in turn leads to simpler creation of a new Bucket, as intended by the function.

### DIFF
--- a/aw-client-rust/src/lib.rs
+++ b/aw-client-rust/src/lib.rs
@@ -62,10 +62,10 @@ impl AwClient {
         &self,
         bucketname: &str,
         buckettype: &str,
-    ) -> Result<(), reqwest::Error> {
+    ) -> Result<Bucket, reqwest::Error> {
         let bucket = Bucket {
             bid: None,
-            id: bucketname.to_string(),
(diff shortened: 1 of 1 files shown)


aw-client-rust/src/lib.rs
```diff
@@ -62,10 +62,10 @@ impl AwClient {
         &self,
         bucketname: &str,
         buckettype: &str,
-    ) -> Result<(), reqwest::Error> {
+    ) -> Result<Bucket, reqwest::Error> {
         let bucket = Bucket {
             bid: None,
-            id: bucketname.to_string(),
+            id: format!("{}_{}", bucketname, self.hostname),
```
ErikBjare, line 68: This is a breaking change. I suggest to keep the original behavior to avoid mishaps.

```suggestion
            id: bucketname.to_string(),
```

If you want your ID with a hostname suffix, better pass a different `bucketname` instead of adding this assumption to the `create_bucket_simple` method.

PR #352, opened 2023-03-16
updated prefix to standard user-writeable location

skewballfox:
changed PREFIX to default to `$XDG_DATA_HOME` unless run with sudo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,12 @@ all: build
 build: aw-server aw-webui
 
 DESTDIR :=
-PREFIX := /usr/local
+ifeq ($(SUDO_USER),)
+    PREFIX := $(HOME)/.local
+else
+    PREFIX := /usr/local
+endif
+
 
 # Build in release mode by default, unless RELEASE=false
 ifeq ($(RELEASE), false)


ErikBjare:
Very nice, thanks! :)

Makefile
```diff
@@ -4,7 +4,12 @@ all: build
 build: aw-server aw-webui
 
 DESTDIR :=
-PREFIX := /usr/local
+ifeq ($(SUDO_USER),)
+    PREFIX := $(HOME)/.local
```
johan-bjareholt, line 8: Code doesn't do what commit message says? You are not even checking the XDG_DATA_HOME variable. On top of that, XDG_DATA_HOME if not overriden by an evvar refers to ~/.local/data rather than just ~/.local.

The change itself makes sense, just a very misleading commit message.

skewballfox: yeah, I meant that as the default location of `XDG_DATA_HOME` but I agree it is misleading. I also realized they might have the variable set to a non standard location or they might have the directory exist but not have the variable set (as is the case on my system). 

I can open a PR to correct that(along with the changes needed to rust server to cover when `$HOME/.local/share` is not part of `$XDG_DATA_DIRS` (which will probably be every system other than mine), but I'm currently at work. I'll try to do that as soon as I can. 

My apologies for the oversight.

PR #353, opened 2023-03-17
Linux pathing

skewballfox:
So this changes both the makefile and server code to check both the system data directories and user data directories. It does this without adding any dependencies. 

the makefile now checks if the user is either root or running as sudo, and defaults to the system prefix if so, if not it first checks if `XDG_DATA_HOME` is set and either sets it to to that or the conventional location for it. it also uses `?=` for assignment so that the user can easily override it

the server does a similar set of checks as it does for `XDG_DATA_DIRS` for `XDG_DATA_HOME` and appends it to the `joined` variable

I do think that something like the directories crate might be a better option. Partially because right now we're checking for the directories of each platform on each platform, and partially because `env::home_dir` is deprecated(though for reasons related to it's windows behavior in linux specific code). 

If you would like to switch to directories or another crate for cross platform pathing, I can do it in this PR or make another one if you would prefer to get the current changes into master first

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,12 @@ all: build
 build: aw-server aw-webui
 
 DESTDIR :=
-PREFIX := /usr/local
+ifeq ($(SUDO_USER),)
+    PREFIX := $(HOME)/.local
+else
+    PREFIX := /usr/local
+endif
+
 
 # Build in release mode by default, unless RELEASE=false
 ifeq ($(RELEASE), false)
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,16 @@ all: build
 build: aw-server aw-webui
 
 DESTDIR :=
-ifeq ($(SUDO_USER),)
-    PREFIX := $(HOME)/.local
+# confirm we are neither root nor running as sudo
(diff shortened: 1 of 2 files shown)


skewballfox:
`$XDG_DATA_HOME` if set, is set by default to `~/.local/share`. I realized that prefix needs to be set to `~/.local` and I can't just cut off the `/share` because if it's set by the user, it might not follow that convention (`XDG_DATA_HOME` might be set to `~/.app_data` for example). I'm assuming the best option here in that case is to default to the convention while allowing the value to be overwritten via an environment variable

skewballfox:
hey is anything blocking this? because I think this is ready to be merged. if not, let me know what needs to be done on my end

ErikBjare:
@johan-bjareholt Plz have a look

skewballfox:
@ErikBjare while it looks as though the handling of pathing within source has changed, and is thus no longer relevant, the makefile does contain a fix for a previous commit. Technically the user could be logged in as root. and it only sets `PREFIX` if it's not supplied by the user via an environment variable. 

you can just copy those specific changes over or I could open a separate PR sometime this weekend.

Issue #343, opened 2023-02-04
Taking datastore lock failed, returning 504: poisoned lock: another task failed inside


BelKed:
On `aw-server-rust` sometimes appears an error *Taking datastore lock failed, returning 504: poisoned lock: another task failed inside*. I guess it happens when I want to load the data from the database using the web interface and at the same time, some watcher writes data to the database...

<img width="1440" alt="Screenshot 2023-02-04 at 20 17 43" src="https://user-images.githubusercontent.com/66956532/216786101-fb6dcf2c-9618-4720-8248-3921674572c6.png">

```
[2023-02-04 20:15:28][[31mERROR[0m][_]: Handler [37mquery[0m panicked.
[2023-02-04 20:15:28][[33mWARN[0m][_]: A panic is treated as an internal server error.
[2023-02-04 20:15:28][[33mWARN[0m][_]: No [1;34m500[0m catcher registered. Using Rocket default.
[2023-02-04 20:15:28][[33mWARN[0m][aw_server::endpoints::bucket]: Taking datastore lock failed, returning 504: poisoned lock: another task failed inside
[2023-02-04 20:15:31][[33mWARN[0m][aw_server::endpoints::query]: Taking datastore lock failed, returning 504: poisoned lock: another task failed inside
[2023-02-04 20:15:31][[33mWARN[0m][aw_server::endpoints::query]: Taking datastore lock failed, returning 504: poisoned lock: another task failed inside
…
```

---

ActivityWatch `v0.12.2b1`

johan-bjareholt:
The issue occurs before that log appears, when the error message "Handler [37mquery[0m panicked." Is printed. However, the logs provided don't show enough info to be able to track down the issue unfortunately. Could you try to run aw-server-rust with the environment variable RUST_BACKTRACE=1 set and reproduce the issue? That should be able to pinpoint the root cause.

BelKed:
Here you go:
```
…
[2023-02-05 19:29:48][WARN][aw_transform::flood]: Gap was of negative duration and could NOT be safely merged (-PT33.986Ss). This warning will only show once per batch.
[2023-02-05 19:29:48][WARN][aw_transform::flood]: Gap was of negative duration (-PT25.798Ss), but could be safely merged. This error will only show once per batch.
[2023-02-05 19:29:48][WARN][aw_transform::flood]: Gap was of negative duration (-PT261.503Ss), but could be safely merged. This error will only show once per batch.
[...]

PR #366, opened 2023-03-24
fix: period_union with empty args

ShootingKing-AM:
Test: `test_period_union_1st_empty()` would replicate #343 (https://github.com/ActivityWatch/aw-server-rust/issues/343#issuecomment-1475094096)


ErikBjare:
Nice, thanks!

PR #385, opened 2023-05-31
feat: embed webui assets into the binary

2e3s:
Hi!

When I was playing with the server, I noticed the inconvenience to distribute. I was able to overcome some problems with things like `cargo-deb` and manifests, but it's limited to deb. There is also a Tauri subproject but it is very bulky.

The proposed change takes the content of the root's `aw-webui/dist` and embeds into the aw-server-rust binary. This can make the distribution much easier, and less prone to problems with filesystem locations, and maybe even more secure as the served files cannot be changed anymore. The executable grows only a bit in disk space, but not RAM.

The trait and macros allow to continue using `lib.rs` easily and flexibly. It uses this crate: https://github.com/pyrossh/rust-embed

What do you think about this idea?


ErikBjare:
I really like this!

And awesome work on awatcher btw!

A few things:

 - What happens if the web UI is missing during build?
 - Would be nice if one could easily let it fallback to whatever is in a dir (being able to rebuild the webui without having to restart aw-server-rust is useful to me)

What do you think @johan-bjareholt? :)

2e3s:
> What happens if the web UI is missing during build?

The build would fail because the macros requires the specified path to exist compile-time, even an empty directory. Is this a normal scenario during a compilation though? If so, there could be `build.rs` with `std::fs::create_dir_all("aw-webui/dist")`.

> Would be nice if one could easily let it fallback to whatever is in a dir

- It could reuse `webpath` CLI setting: if it's specified during the run, then use the filesystem, otherwise use the embedded files. `dirs::get_asset_path()` would be unused at all.
- It could be vice versa: if a file in the data directory is absent, or if the data directory doesn't exist at all, use the embedded file. It would use `webpath` CLI setting or `dirs::get_asset_path()`.
- Have a default cargo feature "embed" and embed files if the feature is on (cargo build --features=embed). If it is off then work with filesystem. The feature would probably turn on different `AssetResolver` implementations, removing or leaving a conditional `webpath` CLI argument like [this](https://github.com/2e3s/awatcher/blob/665924d39ec42d424e1e330e7f119b1ac8c7f6c0/src/config.rs#L65).

Is your suggestion only for testing aw-webui with the server? Maybe the separate feature or something like that would target your use case more precisely while reducing UX complexuty, because it would remove otherwise useless `webpath` CLI setting (which I've forgotten about).

2e3s:
> It could reuse webpath CLI setting: if it's specified during the run, then use the filesystem, otherwise use the embedded files. `dirs::get_asset_path()` would be unused at all.

I have implemented this for these reasons:
- This option removes some complexity of 100 lines of code.
- If data is embedded into the binary then it seems that normally there is no need to try and search for the data throughout the disk.

ErikBjare:
Going to try and merge this today.

Not sure why CI isn't running, as with #396. Excuse me for opening a new one to get CI to run.

ErikBjare:
Now checks are running, some kinks to iron out: https://github.com/ActivityWatch/aw-server-rust/pull/420

2e3s:
Thanks, I'll fix this soon :ok_hand:

2e3s:
CI is fixed with `build.rs`, another option is to have `aw-webui/dist/.gitkeep` in the other repository.

The important consequence of this would be not forgetting to have aw-webui compiled before building the complete rust server. As I understand, it should be already ok now in the build script https://github.com/ActivityWatch/activitywatch/blob/master/.github/workflows/build.yml

aw-server/src/endpoints/mod.rs
```diff
@@ -86,6 +111,18 @@ fn server_info(config: &State<AWConfig>, state: &State<ServerState>) -> Json<Inf
     })
 }
 
+fn get_file(file: PathBuf, state: &State<ServerState>) -> Option<(ContentType, Vec<u8>)> {
+    let asset = state.asset_resolver.resolve(&file.display().to_string())?;
+
+    let content_type = file
+        .extension()
+        .and_then(OsStr::to_str)
+        .and_then(ContentType::from_extension)
```
johan-bjareholt, line 120: This automatic ContentType detection in theory is only needed for the "static" folder, the rest of them could be hardcoded and use state.asset_resolver.resolve directly. Would make things a little bit easier to read.

aw-server/src/endpoints/mod.rs
```diff
@@ -1,20 +1,57 @@
...
 use crate::config::AWConfig;
 
 use aw_datastore::Datastore;
 use aw_models::Info;
+pub trait AssetResolver: Send + Sync {
+    fn resolve(&self, file_path: &str) -> Option<Vec<u8>>;
+}
+
+#[macro_export]
+macro_rules! embed_asset_resolver {
```
johan-bjareholt, line 20: Why is it necessary to have a macro for this?
Can't you instead have a single struct that includes the whole "dist" folder?

johan-bjareholt (commented):
> The executable grows only a bit in disk space, but not RAM.

This doesn't sound right. How did you measure this? Did you only start the application or did you let it run for a while, using the web-ui etc?
If you only start the application and don't use the web-ui, it will likely only use the same amount of memory that it used before because that's all memory will be required to map. However, if you use all assets bundled in the executable, it will be required to map that memory as well. Exactly when the OS will unmap that memory is hard to tell, but most likely it will hold on to that mapped memory for quite some time, likely until the OS runs low on memory.

Regardless, it's probably not a huge issue since the web-ui is only a few MB right now, but if that ever grows we will need to keep that in mind. The pros probably outweigh the cons. Could also be optimized in the future by bundling compressed versions of the files, which then get uncompressed on demand.

aw-server/src/endpoints/mod.rs
2e3s: If `__ProjectAssetResolver` is not behind the macros, then rust-embed's own derive macros always runs compile-time, which requires always having `../aw-webui/dist` relative to this rs file regardless the dynamic dispatch by trait.

Regarding the overall abstraction instead of always embedding `../aw-webui/dist`. I don't know about the strategy of this server development, hence I've made the more conservative, least prohibitive choice to allow more freedom and easier use of the server as a crate, like in the some past instances
- https://github.com/ActivityWatch/aw-tauri/blob/master/src-tauri/Cargo.toml#L19
- https://github.com/2e3s/awatcher/blob/main/Cargo.toml#L39

aw-server/src/endpoints/mod.rs
2e3s: Here I've just reimplemented `NamedFile` responder functionality.
So I've tried to adjust the code, this is what the routes get instead
```rust
    state
        .asset_resolver
        .resolve(&Path::new("css").join(file))
        .map(|state| (ContentType::CSS, state))
```
which looks alright. Not sure how substantial to regard, but there is a problem with 
- `fonts` folder which has `ContentType::WOFF` and `ContentType::WOFF2` files.
- `js` folder which has `ContentType::JavaScript` and `ContentType::JSON` (`.js.map` files).
- also yes, `static` folder. Have you suggested to move the content detection only to `fn root_static`?

2e3s:
> Did you only start the application or did you let it run for a while, using the web-ui etc?

You're right, I only meant that it doesn't load the data on the application start. 
The declaration of embedded data looks like  `const BYTES: &'static [u8] = include_bytes!()` so all the memory nuances are the same to `include_bytes!`. It really means sitting in a virtual memory and loading into RAM when requested.

There is a [compression option](https://github.com/pyrossh/rust-embed#compression) for the crate, but it doesn't affect RAM consumption, only the size on disk.

aw-server/src/endpoints/mod.rs
2e3s: Another way I thought about is `interpolate-folder-path` feature on, which allows `#[folder = "$CARGO_MANIFEST_DIR/aw-webui/dist"]` and hence this directory structure can be replicated elsewhere. I had disregarded it because it's not a maximized flexibility.

aw-server/src/endpoints/mod.rs
johan-bjareholt: Alright, in that case it was probably good as is if detection is pretty much needed for all but the css folder.

aw-server/src/endpoints/mod.rs
johan-bjareholt: >  Another way I thought about is interpolate-folder-path feature on, which allows #[folder = "$CARGO_MANIFEST_DIR/aw-webui/dist"] and hence this directory structure can be replicated elsewhere. I had disregarded it because it's not a maximized flexibility.

That's a pretty cool idea.

A variant on that would be to have a "AW_WEBUI_DIR" variable and if the variable is unset you can set it to "../aw-webui/dist" in build.rs.

johan-bjareholt:
>  You're right, I only meant that it doesn't load the data on the application start.
The declaration of embedded data looks like const BYTES: &'static [u8] = include_bytes!() so all the memory nuances are the same to include_bytes!. It really means sitting in a virtual memory and loading into RAM when requested.
>
> There is a [compression option](https://github.com/pyrossh/rust-embed#compression) for the crate, but it doesn't affect RAM consumption, only the size on disk

I think it's fine as is for now, no need to dig deeper into this right now.

aw-server/src/endpoints/mod.rs
2e3s: My initial suggestion with `interpolate-folder-path` turned out looking pretty awkward when implemented, but your improvement has made it acceptable. :+1:

ErikBjare (approved):
LGTM, let me know when you feel it is ready to merge @2e3s! :)

2e3s:
I have re-reviewed the code and found a minor bug/typo. Otherwise it's good to go.

2e3s:
> let me know when you feel it is ready to merge

Just in case, it's fixed and ready I think :slightly_smiling_face:

ErikBjare:
Awesome, merging! :rocket:

ErikBjare:
Big thanks for getting this together @2e3s! :medal_sports: 

This is a great QoL improvement, and now that I think about it will massively simplify some parts of aw-android (when we eventually get it building again) where we before had to [unzip the webui and make a mess](https://github.com/ActivityWatch/aw-android/blob/master/mobile/src/main/java/net/activitywatch/android/AssetExtractor.kt). Huge win right there, that you didn't even know about!

If you want to [get paid](https://github.com/ActivityWatch/aw-client/blob/master/examples/working_hours.py), just send me the results from [the script](https://github.com/ActivityWatch/aw-client/blob/master/examples/working_hours.py).

ErikBjare:
Made the following fixes post-merge as I was working on updating aw-android:

 - https://github.com/ActivityWatch/aw-server-rust/commit/5cb63a5d2133a826152e10fee2e8568696408491
 - https://github.com/ActivityWatch/aw-server-rust/commit/2e3298e4adc6615230702e10030ac2f4d4f8e80f

*Really* improved the situation on Android, much faster startup speed and less complexity!

PR #393, opened 2023-06-12
Updated dependences + Added workspace.resolver to cleanup warnings

### LINKED ISSUE #376: Bump rocket and rocket_cors
`rocket_cors` recently released [0.6.0-alpha2](https://github.com/lawliet89/rocket_cors/commit/985098dd8f3b052716111eaa872d184cc21a1a68), with the unreleased patch that we used git to get previously. We should be able to bump both dependencies and remove the need for the git source.

PS the server version is one patch release behind :stuck_out_tongue:

Canvas02:
Updated `rocket`, changed `rocket_cors` to use crates.io rather than the git repo, and add `resolver = "2"` to remove some cargo warnings

Fixes #376


ErikBjare:
Sorry I missed this! 

This seems to have been fixed, but I appreciate your effort :)

Thanks for contributing!

PR #396, opened 2023-06-18
Split client to async and blocking

2e3s:
The problem is not only that the current functionality is less useful in async context. The client cannot be used in async context at all as is:
https://docs.rs/reqwest/latest/reqwest/blocking/
> Conversely, the functionality in reqwest::blocking must not be executed within an async runtime, or it will panic when attempting to block

I've used tokio's runtime to block because reqwest uses tokio blocking internally as well.
https://github.com/seanmonstar/reqwest/blob/master/src/blocking/wait.rs
It also uses a current thread runtime, but with timeouts. However, the async client specified the timeout already, so it shouldn't matter much.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -205,6 +205,7 @@ dependencies = [
  "rocket",
  "serde",
  "serde_json",
+ "tokio",
  "tokio-test",
 ]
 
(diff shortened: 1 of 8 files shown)


2e3s:
This watcher doesn't use `rev` to depend on a commit or any pin of the kind https://github.com/ActivityWatch/aw-watcher-window-wayland/blob/master/Cargo.toml (which is not good exactly because of this), so it will have to be updated if this change is accepted.

2e3s:
One alternative would be "async generic" so to speak: https://docs.rs/maybe-async/0.2.6/maybe_async/ - it would exclude tokio from crate's dependencies (`reqwest` would use it regardless), maybe increase build time, remove some code duplication.

ErikBjare:
Hmm, can't figure out why CI doesn't run

PR #400, opened 2023-07-07
ci: bumped Android NDK version from 20 to 25c

ErikBjare:
Trying to get to the bottom of Android CI failures.


ErikBjare:
Looks like it was already done, and I was on a stale master commit.

PR #404, opened 2023-08-23
Update README.md

StefanoChiodino:
hope this helps? I haven't played with rust in a while, and it took me some googling to remember how to run nightlies, and the name maybe is wrong?


ErikBjare:
We assume you've run `rustup default nightly`, which makes the `+nightly` redundant (and we don't want to write it out everywhere).

But in this case it kinda makes sense, so merging. Thanks :)

PR #424, opened 2023-09-27
fix: improve AW_WEBUI_DIR handling

2e3s:
Otherwise, `../aw-webui/dist` is always created which contradicts the flexibility from the env variable. Also, if `AW_WEBUI_DIR` is already and explicitly set then it supposedly means that the build runner has taken care of the matter.

Follow up on https://github.com/ActivityWatch/aw-server-rust/pull/385


ErikBjare:
Good catch!

PR #428, opened 2023-10-19
fix: rewrote settings logic to mimic aw-server-python behavior

ErikBjare:


ErikBjare:
Why does CI not run? :(

PR #436, opened 2023-11-10
perf: improve performance of filter_period_intersect

cjc7373:
In general, this reduces complexity of O(n^2) to O(n lgn).
From the benchmark we can see the performance gain:
```bash
     Running benches/bench.rs (target/release/deps/bench-a5f1a4e304f6ce84)
Gnuplot not found, using plotters backend
1000 events             time:   [443.46 µs 444.09 µs 444.76 µs]
                        change: [-92.832% -92.611% -92.413%] (p = 0.00 < 0.05)
                        Performance has improved.
```


ErikBjare:
Very nice! Slightly embarrassed that it was O(n^2) to begin with 😅 

Thank you so much for this!

PR #437, opened 2023-11-13
Replace panics with results & better option types

nathanmerrill:
1. I tried to remove as many unwraps(), panics()!, etc, with Result<>.  I did not go deep into the core sync logic, as I'm not as comfortable there.

2. I move a bunch of the parsing out of the main function and into arg-specific functions.  That way, clap can provide all the context needed for an invalid argument.  I also replaced "String" with more specific types in the clap commands.

AwClient was the most "substantial" change here.  It's new function had several points where it would panic out.  Therefore, I replaced its constructor arguments with types that prevented these panics.

Instead of taking a host+port, it just takes a reqwest::Url.  This prevents the panic if the url is malformed.  Additionally, it would try to read the hostname() which would also panic if the hostname used non-UTF-8 characters, so the hostname is a new parameter to the function.  (I did consider rewriting get_hostname() to a non-panicing version that just strips all non-UTF-8 characters from the hostname.  If you prefer this, I'm happy to update)

### DIFF
--- a/aw-client-rust/src/lib.rs
+++ b/aw-client-rust/src/lib.rs
@@ -7,8 +7,8 @@ extern crate tokio;
 
 pub mod blocking;
 
-use std::{collections::HashMap, error::Error};
 use std::vec::Vec;
+use std::{collections::HashMap, error::Error};
 
 use chrono::{DateTime, Utc};
 use serde_json::Map;
@@ -29,9 +29,7 @@ impl std::fmt::Debug for AwClient {
 }
 
 fn get_hostname() -> String {
-     return gethostname::gethostname()
-        .to_string_lossy()
-        .to_string();
+    return gethostname::gethostname().to_string_lossy().to_string();
 }
 
 impl AwClient {
@@ -53,7 +51,7 @@ impl AwClient {
     pub async fn get_bucket(&self, bucketname: &str) -> Result<Bucket, reqwest::Error> {
         let url = format!("{}/api/0/buckets/{}", self.baseurl, bucketname);
         let bucket = self
-            .client            
+            .client
             .get(url)
             .send()
             .await?
--- a/aw-client-rust/src/lib.rs
+++ b/aw-client-rust/src/lib.rs
@@ -7,7 +7,7 @@ extern crate tokio;
 
 pub mod blocking;
 
-use std::collections::HashMap;
+use std::{collections::HashMap, error::Error};
 use std::vec::Vec;
 
 use chrono::{DateTime, Utc};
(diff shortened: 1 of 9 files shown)


ErikBjare:
Awesome work!

@johan-bjareholt What do you think about the aw-client change?

aw-sync/src/main.rs
```diff
@@ -73,57 +72,72 @@ enum Commands {
...
+    aw_server::logging::setup_logger("aw-sync", opts.testing, verbose)?;
 
     let port = opts
         .port
-        .or_else(|| Some(crate::util::get_server_port(opts.testing).ok()?.to_string()))
-        .unwrap();
+        .map(|a| Ok(a))
+        .unwrap_or_else(||util::get_server_port(opts.testing))?;
+
+    let baseurl = reqwest::Url::parse(&format!("https://{}:{}", opts.host, port))?;
```
ErikBjare, line 129: I imagine having HTTPS here (and perhaps other places) is the problem.

aw-client-rust/tests/test.rs
```diff
@@ -60,10 +60,9 @@ mod test {
 
     #[test]
     fn test_full() {
-        let ip = "127.0.0.1";
-        let port: String = PORT.to_string();
         let clientname = "aw-client-rust-test";
-        let client: AwClient = AwClient::new(ip, &port, clientname);
+        let url = reqwest::Url::parse(format!("https://127.0.0.1:{}", PORT).as_str()).unwrap();
```
ErikBjare, line 64: https here

ErikBjare, line 64:
```suggestion
        let url = reqwest::Url::parse(format!("http://127.0.0.1:{}", PORT).as_str()).unwrap();
```

aw-sync/src/main.rs
ErikBjare, line 129:
```suggestion
    let baseurl = reqwest::Url::parse(&format!("http://{}:{}", opts.host, port))?;
```

aw-client-rust/src/lib.rs
```diff
@@ -29,13 +29,11 @@ impl std::fmt::Debug for AwClient {
...
 
 impl AwClient {
-    pub fn new(ip: &str, port: &str, name: &str) -> AwClient {
-        let baseurl = format!("http://{ip}:{port}");
+    pub fn new(baseurl: reqwest::Url, name: &str, hostname: String) -> AwClient {
         let client = reqwest::Client::builder()
             .timeout(std::time::Duration::from_secs(120))
             .build()
             .unwrap();
-        let hostname = gethostname::gethostname().into_string().unwrap();
```
johan-bjareholt, line 38: This unwrap is extremely unlikely, you would need to essentially have invalid utf-8 in your hostname to fail here.
Keep this functionality, but maybe replace the "unwrap" with an "expect" instead.

johan-bjareholt (changes requested):
Two minor comments

aw-client-rust/src/lib.rs
```diff
@@ -17,7 +17,7 @@ pub use aw_models::{Bucket, BucketMetadata, Event};
 
 pub struct AwClient {
     client: reqwest::Client,
-    pub baseurl: String,
+    pub baseurl: reqwest::Url,
```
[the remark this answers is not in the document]
nathanmerrill: Is this really an API?  While I found a old crates.io package, this project really does not feel like a library.

That said, if you'd like to keep that distinction, then I agree, but then I think new() should return a Result, which is fine, but a bit odd.

aw-client-rust/src/lib.rs
nathanmerrill: I agree, very unlikely, but in that unlikely event, panic-ing doesn't make sense.  Maybe I should just strip out UTF-8 characters instead?

johan-bjareholt: Sounds good!

aw-client-rust/src/lib.rs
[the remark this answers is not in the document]
johan-bjareholt: It is an API, just not a very mature one :)
As you saw, it's used by aw-sync, but its also used by aw-watcher-window-wayland.

johan-bjareholt (approved):

ErikBjare (approved):
Very good!

ErikBjare:
Thanks for the great PR @nathanmerrill! ❤️

PR #476, opened 2024-03-29
Droppping deprecated functions

brayo-pip:
Just updating the deprecated functions. I built afterwards just to make sure everything works. Looks good to me. Let me know what you think @ErikBjare.


brayo-pip:
Apparently this doesn't work on the master branch but works on the firebase-sync branch that I cherry-picked it from. Probably a skill issue, something I don't know yet about rust. Will figure it out later, or just finish working on the firebase-sync branch. And create a pull request.

PR #499, opened 2025-02-02
fix(cors): set url with chrome-extension:// as regex origin

### LINKED ISSUE #498: Support configuring additional allowed origins
It should be possible for users to change the allowed CORS origin that the server will accept
I have been using sideloaded aw-web-watcher extension for my Chromium browser for a long time, and it worked without any issue until I migrated to aw-server-rust.

## Proposed solution
Allow user to manually add regexes so that they can allow their own sideloaded extensions to access the server

Why this is necessary for my workflow:

I use a custom sideloaded extension because I am using Vivaldi, an alternative Chromium browser, which the aw-web-watcher seems to have issues with tracking hostnames properly.

Relevant issue: https://github.com/ActivityWatch/aw-watcher-web/issues/132, https://github.com/ActivityWatch/aw-watcher-web/issues/102

Therefore, I decided that sideloading will patched hardcoded values be the simplest fix while I wait for a proper fix, which still seems to be missing as of now

The ability to support other extension ids is crucial for me to keep tracking my Chromium web activities.

## Problem

It took me an extended period of time to notice that my Chromium browser is no longer showing up on activity watcher partially because my firefox extension was still logging activities as usual. 
Only when I explicitly searched for my chromium activity, that I noticed it is now missing.
So it had been on endless retrying loop forever since I migrated to rust.

Tracking down the error in the extension shows the below, which isn't helpful
```
{
  "error": {
    "code": 403,
    "reason": "Forbidden",
    "description": "The server refused to authorize the request."
  }
}
```

Peeking at the server logs showed this message
```
[2024-11-14 14:28:29][[31mERROR[0m][rocket_cors::fairing::_]: CORS Error: Origin 'chrome-extension://jmdbkmbphoikckgkcnpoojbfeiaoaocl' is not allowed to request
```

Checking the source code revealed this
https://github.com/ActivityWatch/aw-server-rust/blob/master/aw-server/src/endpoints/cors.rs#L16-L21
```
    let mut allowed_regex_origins = vec![
        "chrome-extension://nglaklhklhcoonedhgnpgddginnjdadi".to_string(),
        // Every version of a mozilla extension has its own ID to avoid fingerprinting, so we
        // unfortunately have to allow all extensions to have access to aw-server
        "moz-extension://.*".to_string(),
    ];
```

I did saw this line, which seems to allow any chromium extension when in testing mode but that does not seem like the intentional use case for said flag
```
    if config.testing {
        allowed_regex_origins.push("chrome-extension://.*".to_string());
    }
```

istudyatuni:
Fix #498 

----

> [!IMPORTANT]
> Modify `cors()` in `cors.rs` to handle Chrome extension URLs as regex origins and add error logging.
> 
>   - **Behavior**:
>     - Modify `cors()` in `cors.rs` to handle Chrome extension URLs as regex origins instead of exact origins.
>     - Filters `config.cors` to separate Chrome extension URLs into `allowed_regex_origins`.
>     - Adds error logging for CORS setup failures.
>   - **Constants**:
>     - Introduces `CHROME_EXTENSION_PREFIX` for identifying Chrome extension URLs.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ActivityWatch%2Faw-server-rust&utm_source=github&utm_medium=referral)<sup> for 8adbf9d7eee229734a96faeaa7b26be125aa2037. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/aw-server/src/endpoints/cors.rs
+++ b/aw-server/src/endpoints/cors.rs
@@ -3,11 +3,21 @@ use rocket_cors::{AllowedHeaders, AllowedOrigins};
 
 use crate::config::AWConfig;
 
+const CHROME_EXTENSION_PREFIX: &str = "chrome-extension://";
+
 pub fn cors(config: &AWConfig) -> rocket_cors::Cors {
     let root_url = format!("http://127.0.0.1:{}", config.port);
     let root_url_localhost = format!("http://localhost:{}", config.port);
     let mut allowed_exact_origins = vec![root_url, root_url_localhost];
-    allowed_exact_origins.extend(config.cors.clone());
+    // url with chrome-extension:// is parsed by url crate as Opaque, so it
+    // should be used as regex origin
+    allowed_exact_origins.extend(
+        config
+            .cors
+            .clone()
+            .into_iter()
+            .filter(|c| !c.starts_with(CHROME_EXTENSION_PREFIX)),
+    );
 
     if config.testing {
         allowed_exact_origins.push("http://127.0.0.1:27180".to_string());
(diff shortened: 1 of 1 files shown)


istudyatuni:
Turns out another PR was merged - #569 (1 year later than this)

PR #500, opened 2025-02-03
Copy executables to Python virtual environment

Game4Move78:
This sets the Makefile to copy `aw-server-rust` (and other executables appending -rust suffix to basename) to the virtualenv (or Python scripts folder) to make the build process more intuitive. Allows `aw-server-rust` to be run from terminal when Python environment used to install ActivityWatch is active, and simplifies using `aw-qt` with `aw-server-rust` without added documentation and extra steps.

As it is the user has to go into `aw-server-rust`, do `cp target/release/aw-server ../venv/bin/aw-server-rust` manually per the README in order to get `aw-server-rust` available under `aw-qt`. Or the rust server is executed in a different path from other modules. 

Related to [ActivityWatch/aw-qt#107](https://github.com/ActivityWatch/aw-qt/pull/107) as aw-server-rust is now the preferred server, which is not documented yet.


Game4Move78:
The alternative would be to update the [docs](https://github.com/ActivityWatch/docs/edit/master/src/installing-from-source.rst) with this:

```rst
Build and install everything into the virtualenv:

.. code-block:: sh

    make build
    cp aw-server-rust/target/release/aw-server venv/bin/aw-server-rust # optional

.. note::
   If you want to use `aw-server-rust` with `aw-qt`, you'll want to copy its binary into your `venv`
```

ErikBjare:
This is essentially what `make package` already does?

Game4Move78:
`make package` moves them to `dist`. This script makes them available on PATH by installing them to the same common venv used by every other module for storing binaries.

Rationale:

- The [aw-server-rust README](https://github.com/ActivityWatch/aw-server-rust/) suggests copying the binary into the `venv`
- Installing into the `venv` ensures the binaries are accessible via PATH while keeping everything contained within the virtual environment.
- Maintains consistency with other modules that install binaries into the `venv` via `poetry install`
- Without this change, `aw-server` is readily available on PATH, but the preferred `aw-server-rust` is not

If you don't like this way of moving the binaries, would you prefer using Poetry to install them?

Game4Move78:
Or use Poetry to install a small Python script that invokes the rust executable?

PR #528, opened 2025-04-15
feat(sync):daemon support for advanced sync

0xbrayo:


0xbrayo:
Should we make the default, syncing all buckets?

aw-sync/src/main.rs
```diff
@@ -247,3 +285,56 @@
...
+
+    ctrlc::set_handler(move || {
+        let _ = tx.send(());
+    })?;
+
+    let sync_dir = dirs::get_sync_dir()?;
+    if let Some(db_path) = &sync_db {
+        info!("Using sync db: {}", &db_path.display());
+
+        if !db_path.is_absolute() {
```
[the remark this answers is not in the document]
0xbrayo: I don't think it's worth atp, never going to reuse/rewrite any other time.

aw-sync/README.md
```diff
@@ -15,10 +15,18 @@ Was originally prototyped as a PR to aw-server: https://github.com/ActivityWatch
 This will start a daemon which pulls and pushes events with the sync directory (`~/ActivityWatchSync` by default) every 5 minutes:
 
 ```sh
+# Basic sync daemon (syncs every 5 minutes)
 aw-sync
+
+# Advanced sync daemon with custom options
+aw-sync daemon --advanced --buckets "aw-watcher-window,aw-watcher-afk" --start-date "2024-01-01"
```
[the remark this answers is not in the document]
0xbrayo: The command fails when I try running it.
image
Could be a regression I introduced will check into it.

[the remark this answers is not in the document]
0xbrayo: I had misread the comment previously. I guess the `--advanced` is redundant, can be inferred whenever there extra flags.

aw-sync/src/main.rs
```diff
@@ -60,7 +60,29 @@ struct Opts {
 enum Commands {
     /// Daemon subcommand
     /// Starts aw-sync as a daemon, which will sync every 5 minutes.
-    Daemon {},
+    Daemon {
+        /// Use advanced sync mode
+        /// (automatically enabled when any advanced options are provided)
+        #[clap(long)]
+        advanced: bool,
```
ErikBjare, line 67: Can be removed now

0xbrayo: Necessary for syncing all buckets by default, though I can remove it and use something like `--buckets *` as an alternative or just sync all buckets by default, resorting to syncing specific buckets when it is specified.

0xbrayo:
Sync all buckets by default now, accepts `--buckets *` for explicit syncing but it's not necessary.

ErikBjare:
Yeah I think it's unnecessary, using `*` might be confusing if people forget to quote it (glob expansion).

0xbrayo:
@ErikBjare ping

aw-sync/src/main.rs
```diff
@@ -111,6 +121,13 @@ fn parse_start_date(arg: &str) -> Result<DateTime<Utc>, chrono::ParseError> {
 }
 
 fn parse_list(arg: &str) -> Result<Vec<String>, clap::Error> {
+    // If the argument is empty or just whitespace, return an empty Vec
+    // This handles the case when --buckets is used without a value
+    if arg.trim().is_empty() {
+        return Ok(vec![]);
+    }
```
ErikBjare, line 128: `buckets` no longer uses parse_list, is this still needed?

aw-sync/src/main.rs
```diff
@@ -139,60 +156,94 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+                None
+            };
 
-                if !db_path.is_absolute() {
-                    Err("Sync db path must be absolute")?
-                }
-                if !db_path.starts_with(&sync_dir) {
-                    Err("Sync db path must be in sync directory")?
+            // If advanced options are provided, use advanced sync mode
+            if start_date.is_some() || effective_buckets.is_some() || sync_db.is_some() {
```
ErikBjare, line 206: I think (may be wrong) that this may never make `effective_buckets = None` as it will return a `Some` if buckets is None (which is the default) and thus advanced will always be used?

@ellipsis-dev plz review

aw-sync/src/main.rs
0xbrayo: you're right it's not. will remove this.

PR #542, opened 2025-08-10
fix(android): Fix JNI signature mismatch causing SIGSEGV on Android 11+

0xbrayo:
- Fixed initialize() function to return jstring instead of void to match Java signature
- Changed jstring_to_string() to return Option<String> instead of panicking on error
- Updated all call sites to handle the Option return type safely

This resolves the crash on newer Android versions where JNI signature mismatch was causing segmentation faults during app initialization.

----

> [!IMPORTANT]
> Fixes JNI signature mismatch in `initialize()` and improves error handling in `jstring_to_string()` to prevent crashes on Android 11+.
> 
>   - **Behavior**:
>     - `initialize()` in `mod.rs` now returns `jstring` instead of `void` to match Java signature, fixing crash on Android 11+.
>     - `jstring_to_string()` in `mod.rs` now returns `Option<String>` instead of panicking, handling conversion errors gracefully.
>   - **Call Sites**:
>     - Updated all call sites in `mod.rs` to handle `Option` return type from `jstring_to_string()` safely, preventing potential crashes.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ActivityWatch%2Faw-server-rust&utm_source=github&utm_medium=referral)<sup> for 04f5bb753f39bc42c498af85ff941cb6df495c2b. You can [customize](https://app.ellipsis.dev/ActivityWatch/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/aw-server/src/android/mod.rs
+++ b/aw-server/src/android/mod.rs
@@ -82,9 +82,10 @@ pub mod android {
         output.into_raw()
     }
 
-    unsafe fn jstring_to_string(env: &JNIEnv, string: JString) -> String {
-        let c_str = CStr::from_ptr(env.get_string(string).expect("invalid string").as_ptr());
-        String::from(c_str.to_str().unwrap())
+    unsafe fn jstring_to_string(env: &JNIEnv, string: JString) -> Option<String> {
+        let java_str = env.get_string(string).ok()?;
+        let c_str = CStr::from_ptr(java_str.as_ptr());
+        c_str.to_str().ok().map(String::from)
     }
 
     unsafe fn string_to_jstring(env: &JNIEnv, string: String) -> jstring {
@@ -136,7 +137,7 @@ pub mod android {
     pub unsafe extern "C" fn Java_net_activitywatch_android_RustInterface_initialize(
         env: JNIEnv,
         _: JClass,
-    ) {
+    ) -> jstring {
         if !INITIALIZED {
             android_logger::init_once(
                 Config::default()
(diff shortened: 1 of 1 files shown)


aw-server/src/android/mod.rs
```diff
@@ -165,9 +165,16 @@ pub mod android {
...
-        let path = &jstring_to_string(&env, java_dir);
-        debug!("Setting android data dir as {}", path);
-        dirs::set_android_data_dir(path);
+        match jstring_to_string(&env, java_dir) {
+            Some(path) => {
+                debug!("Setting android data dir as {}", path);
+                dirs::set_android_data_dir(&path);
+            }
+            None => {
+                // Failed to convert string - do nothing and return
```
petrroll, line 174: Wouldn't it make sense to debug log? Error log or smth? 

What happens if you can't set the data dir? Does the rest handle that somewhat sanely?

aw-server/src/android/mod.rs
```diff
@@ -82,9 +82,10 @@ pub mod android {
         output.into_raw()
     }
 
-    unsafe fn jstring_to_string(env: &JNIEnv, string: JString) -> String {
-        let c_str = CStr::from_ptr(env.get_string(string).expect("invalid string").as_ptr());
-        String::from(c_str.to_str().unwrap())
+    unsafe fn jstring_to_string(env: &JNIEnv, string: JString) -> Option<String> {
+        let java_str = env.get_string(string).ok()?;
```
petrroll, line 86: BTW, did you find out how these can fail?

aw-server/src/android/mod.rs
0xbrayo: yeah, a debug log makes more sense, here. Will do.

aw-server/src/android/mod.rs
0xbrayo: I only figured out they were failing by using a lot of trace logs. Then updated them didn't investigate much about how exactly they were failing.

ErikBjare:
Should this be merged or #547?

0xbrayo:
This preferably, they are both similar in functionality, I already have commits depending on this 😅

ErikBjare:
Then please fix the remaining comments and I'll merge, getting stale :)

ErikBjare:
Nice job nonetheless!

Issue #405, opened 2023-08-29
"poisoned lock: another task failed inside"


EtaoinWu:
I'm using the latest release of AW (0.12.2) on Windows 11 (22H2 build 22621.2215) with `aw-server-rust` enabled. Lately whenever I open up the web UI and check any day under "activity" I got this in the logs for `aw-server-rust`: 

```
[2023-08-29 17:31:41][INFO][aw_server]: Using DB at path "C:\\Users\\etaoi\\AppData\\Local\\activitywatch\\aw-server-rust\\sqlite.db"
[2023-08-29 17:31:41][INFO][aw_server]: Cannot find assets "./aw-webui/dist/"
[2023-08-29 17:31:41][INFO][aw_server]: Using aw-webui assets at path "C:\\Program Files (x86)\\ActivityWatch\\aw-server-rust\\./static/"
[2023-08-29 17:31:41][INFO][aw_server::endpoints]: Starting aw-server-rust at 127.0.0.1:5600
[2023-08-29 17:32:51][ERROR][_]: No matching routes for GET /manifest.json.
[2023-08-29 17:32:51][WARN][_]: No 404 catcher registered. Using Rocket default.
[2023-08-29 17:33:34][ERROR][_]: Handler query panicked.
[2023-08-29 17:33:34][WARN][_]: A panic is treated as an internal server error.
[2023-08-29 17:33:34][WARN][_]: No 500 catcher registered. Using Rocket default.
[2023-08-29 17:33:37][WARN][aw_server::endpoints::bucket]: Taking datastore lock failed, returning 504: poisoned lock: another task failed inside
[2023-08-29 17:33:39][WARN][aw_server::endpoints::bucket]: Taking datastore lock failed, returning 504: poisoned lock: another task failed inside
[2023-08-29 17:33:40][WARN][aw_server::endpoints::bucket]: Taking datastore lock failed, returning 504: poisoned lock: another task failed inside
[2023-08-29 17:33:41][WARN][aw_server::endpoints::bucket]: Taking datastore lock failed, returning 504: poisoned lock: another task failed inside
(the previous warning repeats for dozens of times)
```

The same warning (`Taking datastore lock failed, returning 504: poisoned lock: another task failed inside`) also appears (multiple times) on the web UI. All the After the problem occurs, I cannot use "Timeline" as well. 

However, after a clean start of `aw-server-rust`, I can still view the "timeline" without error of a time range that includes the days that caused the error at "Activity". 

[...]

PR #553, opened 2026-01-01
fix: wrap datastore worker in catch_unwind to prevent poisoned locks

tobixen:
## Summary

Use `std::panic::catch_unwind()` around the datastore worker loop as a safety net. If any panic occurs, the worker will exit gracefully instead of poisoning the Mutex lock and leaving the server in an unusable state.

## Problem

When the datastore worker thread panics (for any reason), it poisons the Mutex lock. All subsequent requests then fail with "poisoned lock: another task failed inside" errors, and the server becomes completely unusable until restart.

This is the symptom described in https://github.com/ActivityWatch/aw-server-rust/issues/405

## Solution

Wrap the worker's `work_loop()` in `catch_unwind()`:
- Catches any panic that occurs
- Logs the panic message
- Lets the worker exit cleanly
- The channel closes properly
- Future requests get clean "channel closed" errors instead of poisoned lock errors

This is a defense-in-depth approach - ideally panics should be replaced with proper error handling, but this safety net ensures no panic can leave the server in an unrecoverable state.

## Test plan

- [x] `cargo build` compiles successfully
- [x] `cargo test --package aw-datastore` passes (9 tests)

🤖 Generated with [Claude Code](https://claude.ai/code)

----

> [!IMPORTANT]
> Wrap `work_loop()` in `catch_unwind()` in `aw-datastore/src/worker.rs` to handle panics gracefully and prevent poisoned locks.
> 
>   - **Behavior**:
>     - Wrap `work_loop()` in `catch_unwind()` in `aw-datastore/src/worker.rs` to handle panics gracefully.
>     - Logs panic messages and allows the worker to exit cleanly, preventing poisoned locks.
>     - Ensures future requests receive clean "channel closed" errors instead of "poisoned lock" errors.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ActivityWatch%2Faw-server-rust&utm_source=github&utm_medium=referral)<sup> for f8cf67010291ec579ef39b231254a6eeb60953e8. You can [customize](https://app.ellipsis.dev/ActivityWatch/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/aw-datastore/src/worker.rs
+++ b/aw-datastore/src/worker.rs
@@ -306,6 +306,18 @@ impl DatastoreWorker {
     }
 }
 
+/// Extracts a human-readable message from a panic payload.
+/// Handles the two common payload types (&str and String), with a fallback for other types.
+fn extract_panic_message(panic_info: &Box<dyn std::any::Any + Send>) -> String {
+    if let Some(s) = panic_info.downcast_ref::<&str>() {
+        s.to_string()
+    } else if let Some(s) = panic_info.downcast_ref::<String>() {
+        s.clone()
+    } else {
+        "Unknown panic".to_string()
+    }
+}
+
 impl Datastore {
     pub fn new(dbpath: String, legacy_import: bool) -> Self {
         let method = DatastoreMethod::File(dbpath);
@@ -331,14 +343,7 @@ impl Datastore {
                 di.work_loop(method);
             }));
             if let Err(panic_info) = result {
-                // Extract panic message if possible
-                let panic_msg = if let Some(s) = panic_info.downcast_ref::<&str>() {
-                    s.to_string()
-                } else if let Some(s) = panic_info.downcast_ref::<String>() {
-                    s.clone()
-                } else {
-                    "Unknown panic".to_string()
-                };
+                let panic_msg = extract_panic_message(&panic_info);
                 error!(
                     "Datastore worker panicked: {}. Worker shutting down gracefully.",
                     panic_msg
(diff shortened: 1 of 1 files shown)


tobixen:
This pull request was created by Claude the AI.  I don't know rust well enough to have a subjective point of view if this change is good or not.  The prompt given to Claude was like this:

> I don't know anything about rust, but wouldn't it be possible to make a general fix so that panic!(...) would cause an immediate graceful exit rather than putting the server in an unusable state?

My rationale is that a server that is down may be auto-restarted, and it's also easy to catch this through monitoring.  A server that is up and listening to the port but not working at all may be a little bit harder to monitor.

I run the server through systemd, and systemd supports auto-restarting.  I'm not sure how things work when utilizing the aw-qt app.

0xbrayo:
> I'm not sure how things work when utilizing the aw-qt app.

I believe only non-zero exit codes get restarted (that's how it works in aw-tauri, not sure about aw-qt). Since that's a crash. This would cause it to exit normally and might not be restarted. You can still throw a panic after gracefully handling the first one though. If that is deemed necessary

tobixen:
My first impression of Claude was just amazing - it apparently did manage to understand some of my projects in some few minutes, even if I thought I was the only person in the world understanding it fully.  It has managed to make small scripts, tons of test code and fix bugs for me much faster and better than what I'm able to do myself.  Other times it's just making tons of crap, or replacing good code with bad code.  I found the "this is fine"-meme quite fitting - it seems happy to burn down houses without realizing it's a bad idea, and if one requests it to change the color of the curtains in the burning house, it would do so without asking any questions.

I've come to realize that delivering pull requests without understanding the project or the code it wants to change perhaps isn't the best idea.

PR #560, opened 2026-02-18
fix: update aw-webui submodule to valid upstream commit (64ed7a7)

0xbrayo:
The previous reference (e493d03c0eb9d54a1495b3a417cffe6dc03123f8) does not exist in the remote repository, causing CI failures. Updated to latest upstream/master commit which is accessible remotely.


TimeToBuildBob (approved):
LGTM. The old submodule ref `e493d03c` is genuinely unreachable (force-pushed away from aw-webui history), so this is a necessary fix. CI is all green.

The only thing to be aware of is that this jumps to aw-webui HEAD, which may include changes beyond just the axios bump — but since the old commit is gone, there's no way to diff against it. Given that CI passes and the alternative is a permanently broken submodule, this is the right call.

PR #581, opened 2026-04-02
[SEC] restrict CORS to authorized extension IDs

RaoufGhrissi:
[SEC] restrict CORS to authorized extension IDs
Fixes a security issue where any Firefox extension (moz-extension://.*) could access the ActivityWatch server without any restriction.

Previously, the CORS configuration included a wildcard for all Mozilla extensions by default. This commit removes that blanket permission and introduces granular control through both static configuration and the Web UI.

We've added 2 new fields to the file configuration (allow_aw_chrome_extension and allow_all_mozilla_extension) and 4 new settings to the Web UI (Fixed origins, Regex origins, and extension-specific shortcuts). The server now merges these settings to determine the final set of authorized origins, ensuring a more secure and flexible configuration.

Dependent on: https://github.com/ActivityWatch/aw-webui/pull/795

### DIFF
--- a/aw-server/src/config.rs
+++ b/aw-server/src/config.rs
@@ -1,5 +1,6 @@
-use std::fs::File;
-use std::io::{Read, Write};
+use std::collections::HashSet;
+use std::fs::{self, File};
+use std::io::Write;
 
 use rocket::config::Config;
 use rocket::data::{Limits, ToByteUnit};
@@ -7,6 +8,14 @@ use rocket::log::LogLevel;
 use serde::{Deserialize, Serialize};
 
 use crate::dirs;
+use serde_json;
+
+pub const CORS_FIELDS: &[&str] = &[
+    "cors",
+    "cors_regex",
+    "cors_allow_aw_chrome_extension",
+    "cors_allow_all_mozilla_extension",
+];
 
 // Far from an optimal way to solve it, but works and is simple
 static mut TESTING: bool = true;
@@ -19,7 +28,7 @@ pub fn is_testing() -> bool {
     unsafe { TESTING }
 }
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Clone)]
 pub struct AWConfig {
     #[serde(default = "default_address")]
     pub address: String,
@@ -36,6 +45,12 @@ pub struct AWConfig {
     #[serde(default = "default_cors")]
     pub cors_regex: Vec<String>,
 
+    #[serde(default = "default_true")]
+    pub cors_allow_aw_chrome_extension: bool,
+
+    #[serde(default = "default_false")]
+    pub cors_allow_all_mozilla_extension: bool,
+
     // A mapping of watcher names to paths where the
     // custom visualizations are located.
     #[serde(default = "default_custom_static")]
@@ -50,6 +65,8 @@ impl Default for AWConfig {
             testing: default_testing(),
             cors: default_cors(),
             cors_regex: default_cors(),
+            cors_allow_aw_chrome_extension: default_true(),
+            cors_allow_all_mozilla_extension: default_false(),
             custom_static: default_custom_static(),
         }
     }
@@ -91,6 +108,14 @@ fn default_testing() -> bool {
     is_testing()
 }
 
+fn default_true() -> bool {
+    true
+}
+
+fn default_false() -> bool {
+    false
+}
+
 fn default_port() -> u16 {
     if is_testing() {
         5666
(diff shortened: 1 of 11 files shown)


aw-server/src/endpoints/mod.rs
```diff
@@ -132,7 +132,7 @@ pub fn build_rocket(server_state: ServerState, config: AWConfig) -> rocket::Rock
         "Starting aw-server-rust at {}:{}",
         config.address, config.port
     );
-    let cors = cors::cors(&config);
+    let cors = cors::cors(&config, &server_state.datastore);
```
ErikBjare: Don't pass datastore here

aw-server/src/config.rs
```diff
@@ -36,6 +36,24 @@ pub struct AWConfig {
...
+    #[serde(default = "default_cors")]
+    pub cors_from_settings: Vec<String>,
+
+    #[serde(default = "default_cors")]
+    pub cors_regex_from_settings: Vec<String>,
+
+    #[serde(default = "default_true")]
+    pub allow_aw_chrome_extension_from_settings: bool,
+
+    #[serde(default = "default_false")]
```
RaoufGhrissi, line 51: i'm hesitating to set default value to True, so it doesn't impact the actual users

aw-server/src/config.rs
```diff
@@ -91,6 +115,14 @@ fn default_testing() -> bool {
     is_testing()
 }
 
+fn default_true() -> bool {
+    true
+}
+
+fn default_false() -> bool {
+    false
+}
```
ErikBjare, line 117: ...really?

aw-server/src/endpoints/cors.rs
```diff
@@ -8,20 +8,25 @@ pub fn cors(config: &AWConfig) -> rocket_cors::Cors {
...
+    allowed_regex_origins.extend(config.cors_regex_from_settings.clone());
+
+    if config.allow_aw_chrome_extension || config.allow_aw_chrome_extension_from_settings {
+        allowed_regex_origins.push("chrome-extension://nglaklhklhcoonedhgnpgddginnjdadi".to_string());
     }
-    let mut allowed_regex_origins = vec![
-        "chrome-extension://nglaklhklhcoonedhgnpgddginnjdadi".to_string(),
-        // Every version of a mozilla extension has its own ID to avoid fingerprinting, so we
-        // unfortunately have to allow all extensions to have access to aw-server
-        "moz-extension://.*".to_string(),
```
ErikBjare, line 20: Don't nuke this very important comment.

aw-server/src/config.rs
```diff
@@ -50,6 +68,12 @@ impl Default for AWConfig {
             testing: default_testing(),
             cors: default_cors(),
             cors_regex: default_cors(),
+            allow_aw_chrome_extension: default_true(),
+            allow_all_mozilla_extension: default_false(),
+            cors_from_settings: default_cors(),
+            cors_regex_from_settings: default_cors(),
+            allow_aw_chrome_extension_from_settings: default_true(),
+            allow_all_mozilla_extension_from_settings: default_false(),
```
ErikBjare: Not a fan of this many additional config keys. Should be prefixed consistently with `cors_` such as `cors_allow_aw_chrome_extension` and `cors_allow_all_mozilla_extension`.

Curious why the `_from_settings` is passed along and not resolved/merged into the non-`from_settings` variants?

aw-server/src/endpoints/mod.rs
```diff
@@ -127,11 +127,47 @@ fn get_file(file: PathBuf, state: &State<ServerState>) -> Option<(ContentType, V
...
+        if let Ok(raw) = db.get_key_value("settings.cors_regex") {
+            config.cors_regex_from_settings = parse_cors_list(raw);
+        }
+        if let Ok(raw) = db.get_key_value("settings.allow_aw_chrome_extension") {
+            config.allow_aw_chrome_extension_from_settings = parse_bool(raw);
+        }
+        if let Ok(raw) = db.get_key_value("settings.allow_all_mozilla_extension") {
+            config.allow_all_mozilla_extension_from_settings = parse_bool(raw);
+        }
+    }
```
ErikBjare: I'm a little bit paranoid about letting web UI set CORS like that...

RaoufGhrissi: do you have another suggestion to let non-devs configure cors ?

aw-server/src/config.rs
RaoufGhrissi: merged with the existing vars, and explained in the readme file

aw-server/src/endpoints/mod.rs
```diff
@@ -127,11 +127,55 @@ fn get_file(file: PathBuf, state: &State<ServerState>) -> Option<(ContentType, V
...
+                .map(|s| s.trim().to_string())
+                .filter(|s| !s.is_empty())
+                .collect()
+        };
+        let parse_bool = |raw: &str| -> bool {
+            serde_json::from_str::<bool>(raw).unwrap_or_else(|_| raw.trim_matches('"') == "true")
+        };
+        // Sync settings between Config file and Database.
+        // On the first run (when a key is missing in the DB), we seed the DB with the value from the config file.
+        // On subsequent runs, we always prefer the DB value (which might have been changed via the UI).
```
ErikBjare: This seems problematic. It's not acceptable to have file-level config options being ignored by auto-set db-values. Makes the config file mostly useless/unreliable. Not really any "sync" happening here.

imo, order of settings/config precedence in server should be cli options -> env vars -> config file -> webui/api settings -> defaults. 

One way to resolve this would be to make the webui config write directly to the config file, would effectively put them in sync (but still not a fan of exposing server config like this in api/webui tbh). But that might require a API endpoint different from `/settings`, maybe `/config` - but really unsure about this, just an idea.

Simplest would probably be to just prefer the config over the db settings and disable the webui settings if they are explicitly set in config/env/cli.

RaoufGhrissi: Thank you @ErikBjare  for the detailed feedback. I completely agree with the proposed precedence order (CLI > Env > Config File > WebUI/API > Defaults), and I have implemented the changes to reflect exactly this logic:

Config File Priority: the configuration file now takes absolute precedence over the database. If a CORS field is explicitly defined in config.toml, the server uses that value and ignores any existing database overrides for that specific field.
for CLI/Env: I verified that cors settings are not currently managed via CLI or Env variables in aw-server-rust. However, the current architecture is future-proof: since the database fallback only triggers for fields missing from the static configuration, any future addition of CLI/Env overrides would naturally take precedence like port and address.

UI Feedback & Safety: To address the concern about "unreliable" config, the Web UI now detects which fields are defined in the config.toml. These fields are marked in the UI as "Fixed in config file" (highlighted in orange) and the inputs are disabled. This ensures the user is aware that the file-level config is in control and cannot be bypassed via the API.

PR #585, opened 2026-04-16
feat(server): add opt-in API key authentication via Bearer token

### LINKED ISSUE #494: Password based authentication
So that not any app (or even extension) can access all of AW's data. Web-UI would also need to authenticate with a password (login could be saved via cookie, though). The Client-Libraries (python + JS) would have to get updated

TimeToBuildBob:
## Summary

Adds opt-in API key authentication to aw-server-rust. When `api_key` is set in `config.toml`, all API endpoints except `/api/0/info` require an `Authorization: Bearer <key>` header.

**Implementation**: Rocket Fairing (same pattern as the existing `HostCheck` fairing) — applies globally without modifying individual endpoints.

**Config** (`~/.config/activitywatch/aw-server-rust/config.toml`):
```toml
api_key = "your-secret-key-here"
```

Leave unset (default) to disable authentication — existing setups are unaffected.

## Behavior

- ✅ `GET /api/0/info` — always public (health check / version endpoint)
- ✅ `OPTIONS` preflight requests — always pass through (CORS)
- 🔒 All other endpoints — require `Authorization: Bearer <api_key>`
- 401 Unauthorized response with descriptive error on failure

## Tests

5 new unit tests covering:
- No api_key configured → all endpoints accessible
- api_key set → /api/0/info still public
- api_key set → other endpoints require valid key
- Invalid key → 401
- Wrong auth scheme (Basic) → 401

## Notes

- This fixes the local security issue on Android where any app can access the AW API
- Useful for users exposing AW on non-localhost networks (with CORS configured)
- Key is stored in config.toml — users manage it manually in MVP scope

Closes #494
Refs ActivityWatch/activitywatch#32, ActivityWatch/activitywatch#1199

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -281,6 +281,7 @@ dependencies = [
  "sd-notify",
  "serde",
  "serde_json",
+ "subtle",
  "toml",
  "uuid",
 ]
@@ -2622,6 +2623,12 @@ name = "strsim"
 version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7da8b5736845d9f2fcb837ea5d9e2628564b3b043a70948a3f0b778838c5fb4f"
+
+[[package]]
+name = "subtle"
+version = "2.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "13c2bddecc57b384dee18652358fb23172facb8a2c51ccc10d74c157bdea3292"
 
 [[package]]
 name = "syn"
--- a/aw-server/Cargo.toml
+++ b/aw-server/Cargo.toml
@@ -28,6 +28,7 @@ gethostname = "0.4"
 uuid = { version = "1.3", features = ["serde", "v4"] }
 clap = { version = "4.1", features = ["derive", "cargo"] }
 log-panics = { version = "2", features = ["with-backtrace"]}
+subtle = "2"
 rust-embed = { version = "8.0.0", features = ["interpolate-folder-path", "debug-embed"] }
 
 aw-datastore = { path = "../aw-datastore" }
--- a/aw-server/src/config.rs
+++ b/aw-server/src/config.rs
@@ -19,6 +19,17 @@ pub fn is_testing() -> bool {
     unsafe { TESTING }
 }
 
+/// Authentication configuration, serialised as `[auth]` in config.toml.
+#[derive(Serialize, Deserialize, Default)]
+pub struct AWAuthConfig {
+    /// Optional API key for Bearer-token authentication.
+    /// When set, all `/api/*` endpoints except `/api/0/info` require:
+    ///   Authorization: Bearer <api_key>
+    /// Leave unset (default) to disable authentication.
+    #[serde(default)]
+    pub api_key: Option<String>,
+}
+
 #[derive(Serialize, Deserialize)]
 pub struct AWConfig {
     #[serde(default = "default_address")]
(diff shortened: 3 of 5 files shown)


TimeToBuildBob:
Addressed both security findings from the Greptile review (commit 7266413):

- **Timing oracle**: Replaced `token == api_key` with `token.as_bytes().ct_eq(api_key.as_bytes())` from the [`subtle`](https://crates.io/crates/subtle) crate (constant-time comparison). Added `subtle = "2"` to `aw-server/Cargo.toml`.
- **Empty key bypass**: Added a guard in `ApiKeyCheck::new()` that converts an empty `api_key` string to `None` with a `warn!()` log message, so an empty-string config silently disables auth rather than silently accepting empty Bearer tokens.
- Added a test for the empty-key case (`test_empty_api_key_disables_auth`).

Also fixed the `cargo fmt` failure: `mod apikey` was out of alphabetical order in `endpoints/mod.rs`.

All 6 apikey unit tests pass.

TimeToBuildBob:
@greptileai review

ErikBjare:
In `~/.config/activitywatch/aw-server-rust/config.toml` the `api_key` should be under some suitable heading, we don't want section-less keys in the top-level.

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob Did you see https://github.com/ActivityWatch/aw-server-rust/pull/585#issuecomment-4267465060, how is the config structure looking?

ErikBjare:
@TimeToBuildBob Also, I kinda thought there were already some kind of API key loading and generation in aw-server-rust?? Or is this just an old/stale PR?

TimeToBuildBob:
@ErikBjare Answering both questions:

**1. Is this PR stale / does aw-server-rust already have API key code?**
Not stale — it's a fresh introduction. I confirmed against `origin/master`: there's no `api_key` field in `aw-server/src/config.rs`, no `apikey.rs`/`auth.rs` in `aw-server/src/endpoints/`, and no Bearer-token handling anywhere in the crate. PR #585 is the first time API key auth lands in aw-server-rust.

You may be thinking of one of:
- aw-server (Python) which had token logic at one point
- The `--testing` flag, which gates some endpoints but isn't auth
- Discussion/proposals from an older issue (happy to dig if you remember roughly when)

**2. Config structure (per your 11:04 comment)**
Currently the field is still top-level in `AWConfig`:
```toml
api_key = "your-secret-key-here"
cors = []
```
i.e. it ships flat alongside `address`, `port`, `cors`. That violates your "no section-less keys at top level" rule — you're right.

I'll push a follow-up that nests it under a heading. The natural shape, given the existing flat fields, would be `[auth]` (since `address`/`port` are already top-level for legacy reasons and we don't want to break those):

```toml
address = "127.0.0.1"
port = 5600

[auth]
api_key = "your-secret-key-here"
```

Or if you'd prefer, `[server.auth]` with everything migrated under `[server]` (bigger breaking change). Which do you want? I'll default to `[auth]` if I don't hear back, since it's additive and doesn't break existing configs.

ErikBjare:
`[auth]` makes sense to me

ErikBjare:
@TimeToBuildBob You seem to not be very responsive on these repos... I need to explicitly tag you.

TimeToBuildBob:
@ErikBjare Done — pushed `refactor(config): nest api_key under [auth] section` (572cedc).

Config now looks like:
```toml
address = "127.0.0.1"
port = 5600

[auth]
api_key = "your-secret-key-here"
```

All 6 unit tests still pass. Existing configs without an `[auth]` section continue to work (defaults to auth disabled). Ready to merge.

PR #586, opened 2026-04-17
feat(client): auto-load API key for aw-client-rust

TimeToBuildBob:
## Summary
- auto-load `api_key` from the local `aw-server-rust` config when connecting to `localhost` or `127.0.0.1`
- attach a default `Authorization: Bearer ...` header for authenticated requests
- add an integration test covering a config-matched authenticated server

## Notes
- the client now uses reqwest `rustls-tls-native-roots` instead of the native-tls default, which removes the OpenSSL system dependency while keeping HTTPS support

## Testing
- `cargo test -p aw-client-rust`

### DIFF
--- a/aw-client-rust/src/lib.rs
+++ b/aw-client-rust/src/lib.rs
@@ -10,15 +10,36 @@ pub mod classes;
 pub mod queries;
 pub mod single_instance;
 
-use std::{collections::HashMap, error::Error};
+use std::{collections::HashMap, error::Error, fs, path::PathBuf};
 
 use chrono::{DateTime, Utc};
+use reqwest::header::{HeaderMap, HeaderValue, AUTHORIZATION};
 use serde_json::{json, Map};
 use single_instance::SingleInstance;
 use std::net::TcpStream;
 use std::time::Duration;
 
 pub use aw_models::{Bucket, BucketMetadata, Event};
+
+#[derive(serde::Deserialize, Default)]
+struct LocalAuthConfig {
+    #[serde(default)]
+    api_key: Option<String>,
+}
+
+#[derive(serde::Deserialize, Default)]
+struct LocalServerConfig {
+    #[serde(default)]
+    port: Option<u16>,
+    #[serde(default)]
+    auth: LocalAuthConfig,
+}
+
+#[derive(Clone, Copy)]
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Addressed Greptile's findings:

- **P2 (fixed)**: Removed unused `rand = "0.9"` dep — confirmed no `rand::` imports anywhere in `src/` or `tests/`.
- **P1 (not a real issue)**: Greptile flagged `with_config_home` using `XDG_CONFIG_HOME` as Linux-only, but CI passes on macOS and Windows too — likely because the integration tests aren't run on those platforms in this workflow (they require a running server).
- **P2 (TOCTOU in `reserve_port`)**: Theoretical race; acceptable for test code.

aw-client-rust/src/lib.rs
```diff
@@ -39,13 +60,69 @@ fn get_hostname() -> String {
...
+    }
+
+    let config_dir = get_server_config_dir()?;
+    let candidates = [
+        ConfigCandidate {
+            filename: "config.toml",
+            default_port: 5600,
+        },
+        ConfigCandidate {
+            filename: "config-testing.toml",
```
ErikBjare, line 83: What? It should only read this in testing mode...

ErikBjare (changes requested):

TimeToBuildBob:
## PR Update

The review feedback has been addressed in a separate PR: **ActivityWatch/aw-server-rust#587**

### Changes

| Issue | Fix |
|-------|-----|
| **P1** `XDG_CONFIG_HOME` cross-platform | Gated test on `#[cfg(target_os = "linux")]` — not respected on macOS/Windows |
| **P2** TOCTOU race in `reserve_port` | `TcpListener` now kept alive via `thread_local` until server binds |
| **P2** Missing `ENV_LOCK` in `test_full` | Holds lock during `AwClient::new()` to prevent parallel-test interference |
| **P2** Unused `rand` dep | Not present in PR — no action needed |

Please re-review when you have a chance!

TimeToBuildBob:
Superseded by #587, which incorporates your inline feedback (load_local_api_key moved to test-only), the Greptile fixes (cross-platform test gating, TOCTOU port fix), and a clean description. Closing this one.

PR #598, opened 2026-05-09
feat(server): server-side privacy filter at heartbeat ingestion

### LINKED ISSUE #482: Suggestion: Regex-based inbound data filters
Being able to filter the inbound data [has been the most popular feature request for a very long time](https://forum.activitywatch.net/t/add-an-exclude-list/345) and comes up again and again in the issues of different repositories. In the status quo, some watchers (e.g. [the window watcher](https://github.com/ActivityWatch/aw-watcher-window/pull/37)) have their own filtering, but they have various problems:

1. **Discoverability:** The configuration is so hidden that [even the developers forget that it exist](https://github.com/ActivityWatch/aw-server/issues/59#issuecomment-538738568) (no offense). Even when it's exposed as a config, it still requires the user to find it somewhere in the docs, then edit the configuration file to set it up correctly. This is not very discoverable. 
2. **Consistency:** The user needs to look into the docs for each watcher to see whether it supports data filtering and to find out how it's configured.
3. **Limited expressiveness:** For the window watcher, it's only possible to remove all window titles. Many people say they only want to exclude some sensitive information (e.g. mail subjects). In my case, I'm frequently using an app that shows a timer in the window title, creating a new entry every second, which makes the timeline barely readable and very unresponsive. 

## Suggestion

We could add regex-based filtering on the heartbeat level: whenever a heartbeat comes in, it's checked against a set of user-configurable regexes. If one matches on any field of the entry, the entry is discarded. We could also extend this feature to allow regex-replacing entries or matching only some fields of the JSON entry.

Similar inbound data filters can be found in e.g. [Sentry](https://docs.sentry.io/product/relay/).

I could probably implement this by myself, at least in Python and the Vue frontend, but probably in Rust as well, but I'd like to know if the approach is welcome in the first place. Tbh, if it isn't, I'd consider writing a simple proxy server that does exactly this -- applying some replacements to the `heartbeat` endpoint and passing everything else through.

TimeToBuildBob:
Implements server-side privacy filtering at heartbeat ingestion, as proposed in #482.

## What this does

Rules are declared in `aw-server.toml` under a `[[privacy_filters]]` array:

```toml
# Drop events matching a regex (silent discard, 200 OK so watchers don't retry)
[[privacy_filters]]
bucket_prefix = "aw-watcher-window"
field = "title"
pattern = "(?i)private browsing|incognito"
action = "drop"

# Redact a field value when it matches a regex
[[privacy_filters]]
field = "title"
pattern = "(?i)password|credential|secret"
action = "redact"
replacement = "REDACTED"
```

Each rule supports:
- `bucket_prefix` — scope the rule to matching bucket ids (empty = all buckets)
- `field` — dotted path into `event.data` (e.g. `url.host` for nested JSON)
- `pattern` — [fancy-regex](https://docs.rs/fancy-regex/) pattern (lookbehind/ahead supported); invalid patterns are loud-rejected at startup, never silently escaped
- `action` — `drop` (discard the event) or `redact` (replace field value)
- `replacement` — custom replacement string for redact (default: `REDACTED`)
- `enabled` — set `false` to disable a rule without removing it

## Design decisions

**`drop` returns 200 OK** with an empty `Event` rather than 4xx so older watchers don't enter retry loops. The watcher's contract ("event was accepted") is preserved even though nothing was persisted.

**Config-file only for now.** Central configuration lives in `aw-server.toml`. The field is named `privacy_filters` so it can be extended later to include a live settings-API path (e.g. `/api/0/settings/privacy_filters`) without breaking the config-file path.

**fancy-regex is already a workspace dependency** (used by `aw-query` and `aw-transform`), so this adds no new crates.

## Tests

- 9 unit tests in `aw-server/src/privacy_filter.rs` — drop, redact, bucket scoping, invalid regex disabling, disabled rule, empty prefix, dotted field path, custom replacement, batch
- 2 API integration tests in `aw-server/tests/api.rs` — full heartbeat round-trip for drop and redact actions

## Related

- Closes #482
- See also ActivityWatch/aw-watcher-window#99 (merged client-side title exclusion) — this is the server-side double-filter that catches events from watchers that don't pre-filter

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -255,7 +255,7 @@ dependencies = [
 
 [[package]]
 name = "aw-server"
-version = "0.13.1"
+version = "0.14.0"
 dependencies = [
  "android_logger",
  "aw-client-rust",
@@ -266,6 +266,7 @@ dependencies = [
  "chrono",
  "clap",
  "dirs",
+ "fancy-regex",
  "fern",
  "gethostname",
  "jemallocator",
--- a/aw-server/Cargo.toml
+++ b/aw-server/Cargo.toml
@@ -30,6 +30,7 @@ clap = { version = "4.1", features = ["derive", "cargo"] }
 log-panics = { version = "2", features = ["with-backtrace"]}
 subtle = "2"
 rust-embed = { version = "8.0.0", features = ["interpolate-folder-path", "debug-embed"] }
+fancy-regex = "0.12.0"
 
 aw-datastore = { path = "../aw-datastore" }
 aw-models = { path = "../aw-models" }
--- a/aw-server/src/android/mod.rs
+++ b/aw-server/src/android/mod.rs
@@ -124,6 +124,7 @@ pub mod android {
                 datastore: Mutex::new(openDatastore()),
                 asset_resolver: endpoints::AssetResolver::new(None),
                 device_id: device_id::get_device_id(),
+                privacy_filters: Vec::new(),
             };
             info!("Using server_state:: device_id: {}", server_state.device_id);
 
--- a/aw-server/src/config.rs
+++ b/aw-server/src/config.rs
@@ -7,6 +7,7 @@ use rocket::log::LogLevel;
 use serde::{Deserialize, Serialize};
 
 use crate::dirs;
+use crate::privacy_filter::PrivacyFilter;
 
 // Far from an optimal way to solve it, but works and is simple
 static mut TESTING: bool = true;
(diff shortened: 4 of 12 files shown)


TimeToBuildBob:
Closing in favor of #600, which implements the same feature with a cleaner architecture (settings-driven via `RefreshPrivacyFilter` command, lives in `aw-datastore` closer to storage) and passing CI.

This PR was created in a parallel autonomous session — three privacy-filter PRs hit the repo within ~1 minute (#598, #599, #600). #599 was already closed; consolidating on #600.

The extra integration tests and config-file path from this PR can be ported to #600 as a follow-up if desired.

PR #599, opened 2026-05-09
feat(privacy-filter): add server-side heartbeat privacy filtering

### LINKED ISSUE #1: Add functionality to redact/filter sensitive data
We need a model to filter out sensitive data by default.

For example if a window title contains: "[title] - Firefox (Private Browsing)" we should redact [title] to some magic string such as "REDACTED".

For some cases we might want to filter the window out entirely, giving 0 information about which window is running, better catch too much than too little.

It should be the goal that every user has a set of "clean" data. The filtering should also be able to be run on an existing database of data, so that cleaner data can be output. Preferably, the data should be so clean that there is little (or even no) reason not to share it (which would be great since easy access to a large dataset could make research in some areas a lot easier!).

The question left is where this processing step should take place. We want the filtering/redacting to happen before data is sent anywhere but it should also be able to be enforceable on a server (if the server owner doesn't trust the servers security, if in the cloud for example) and have clients notified of this so that they can do the filtering on their side, removing the need to send sensitive data at all. It might therefore be prudent to write a module in aw-core that implements this functionality since it should be useable from the server and all clients (which transmit sensitive data).

This feature should be on by default, we don't need anything advanced yet, first priority is to redact titles from Incognito/Private Browsing, that's a good step in the right direction.

This should have a far higher priority than Zero-Knowledge storage right now, because it's a lot easier and is more user friendly (In ZK storage: if you lose your keys you lose your data).

---

Useful when:
 - We want to export data to a 3rd party service but don't need them to know all the details.
 - We want to do overview analysis where full detail is not necessary.
 - We want to redact some information in the log, such as the window titles of Incognito/Private browser-windows, Tor Browser, etc.

---

This issue was originally moved from https://github.com/ActivityWatch/aw-server/issues/4 which ended up here because it ended up having wider scope not only relating to aw-server. 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/34506378-add-functionality-to-redact-filter-sensitive-data?utm_campaign=plugin&utm_content=tracker%2F35920020&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F35920020&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TimeToBuildBob:
## Summary

Adds configurable regex-based privacy filters applied at heartbeat ingestion — the most-requested ActivityWatch feature (ActivityWatch/activitywatch#1, 10+ years open).

Rules are stored in the settings key `privacy_filters` as a JSON array and applied on every `POST /api/0/buckets/<id>/heartbeat` request before the event reaches the datastore.

## Rule schema

```json
[
  {
    "bucket_prefix": "aw-watcher-window",
    "field": "title",
    "pattern": "(?i)private browsing|incognito",
    "action": "drop"
  },
  {
    "field": "title",
    "pattern": "(?i)secret|confidential",
    "action": "redact",
    "replacement": "[redacted]"
  }
]
```

| Field | Required | Description |
|-------|----------|-------------|
| `field` | ✅ | Event data key to match (`title`, `app`, `url`, …) |
| `pattern` | ✅ | fancy-regex pattern (lookaheads, Unicode, case flags) |
| `action` | ✅ | `drop` — discard event entirely; `redact` — replace field value |
| `bucket_prefix` | optional | Scope rule to buckets whose ID starts with this string |
| `replacement` | optional | Replacement string for `redact` action (default: `[redacted]`) |

## Implementation

- New `aw_transform::privacy_filter` module — zero new dependencies (fancy-regex already present)
- Applied in `bucket_events_heartbeat` before `datastore.heartbeat()`
- Invalid regex → warning logged, rule skipped (fail-open, never breaks a watcher)
- Non-string field values → skipped silently (type-safe)
- Dropped events → HTTP 200 with empty event body (clients see no error)
- 9 unit tests covering: drop, redact, bucket scoping, default replacement, invalid regex, non-string values, multiple redact rules, empty rules

## What this does NOT include (intentional MVP scope)

- aw-webui settings UI — add rules via `/api/0/settings/privacy_filters` directly or via UI in a follow-up
- aw-watcher-window pre-filtering — can be done separately using the same rule shape
- Hidden/masked categories (store but hide in visualizations) — separate feature

## Related

- Closes ActivityWatch/activitywatch#1 (partial — server-side filter MVP)
- Relates to #482 (prior regex-matching PR; this builds on the same design direction)
- Design discussion: ActivityWatch/aw-server-rust#482

### DIFF
--- a/aw-server/src/endpoints/bucket.rs
+++ b/aw-server/src/endpoints/bucket.rs
@@ -10,6 +10,7 @@ use aw_models::Bucket;
 use aw_models::BucketsExport;
 use aw_models::Event;
 use aw_models::TryVec;
+use aw_transform::{apply_privacy_filter, PrivacyFilterRule};
 
 use rocket::http::Status;
 use rocket::State;
@@ -155,6 +156,23 @@ pub fn bucket_events_heartbeat(
 ) -> Result<Json<Event>, HttpErrorJson> {
     let heartbeat = heartbeat_json.into_inner();
     let datastore = endpoints_get_lock!(state.datastore);
+
+    // Load and apply privacy filters from settings.
+    // If the setting is absent or unparseable, filters are skipped (fail-open).
+    let rules: Vec<PrivacyFilterRule> = match datastore.get_key_value("settings.privacy_filters") {
+        Ok(raw) => serde_json::from_str(&raw).unwrap_or_else(|e| {
+            warn!("Failed to parse privacy_filters setting: {}", e);
+            vec![]
+        }),
+        Err(_) => vec![],
+    };
+
+    let heartbeat = match apply_privacy_filter(bucket_id, heartbeat, &rules) {
+        Some(event) => event,
+        // Event matched a drop rule — acknowledge without storing.
+        None => return Ok(Json(Event::default())),
+    };
+
     match datastore.heartbeat(bucket_id, heartbeat, pulsetime) {
         Ok(e) => Ok(Json(e)),
         Err(err) => Err(err.into()),
--- a/aw-transform/src/lib.rs
+++ b/aw-transform/src/lib.rs
@@ -50,3 +50,6 @@ pub use period_union::period_union;
 
 mod union_no_overlap;
 pub use union_no_overlap::union_no_overlap;
+
+pub mod privacy_filter;
+pub use privacy_filter::{apply_privacy_filter, FilterAction, PrivacyFilterRule};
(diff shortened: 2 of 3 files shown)


TimeToBuildBob:
Closing as duplicate — a parallel session from the same author submitted #598 moments earlier with a more complete implementation (config-file based, integration tests, dotted field paths, enabled flag). Please review #598 instead.

PR #607, opened 2026-05-21
fix(aw-client-rust): fail on non-2xx responses

TimeToBuildBob:
## Summary
- reject non-success HTTP responses across aw-client-rust request paths before returning success or decoding JSON
- add focused regression tests for async and blocking clients using a tiny mock HTTP server
- drop the stale README note claiming requests never fail

## Verification
- cargo test -p aw-client-rust

### DIFF
--- a/aw-client-rust/README.md
+++ b/aw-client-rust/README.md
@@ -2,5 +2,3 @@ aw-client-rust
 ==============
 
 WIP: aw-client implementation in Rust
-
-TODO: Better error handling (requests currently never fail?)
--- a/aw-client-rust/src/lib.rs
+++ b/aw-client-rust/src/lib.rs
@@ -55,6 +55,12 @@ fn build_client(api_key: Option<String>) -> Result<reqwest::Client, Box<dyn Erro
 }
 
 impl AwClient {
+    async fn send_success(
+        request: reqwest::RequestBuilder,
+    ) -> Result<reqwest::Response, reqwest::Error> {
+        request.send().await?.error_for_status()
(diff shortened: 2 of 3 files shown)


TimeToBuildBob:
## Greptile hardening applied

Addressed the Greptile review note about the mock server single-read-per-connection design in `aw-client-rust/tests/status_errors.rs`:

**Problem**: The mock server used a single `stream.read()` per connection. For POST requests with fragmented TCP delivery, reqwest could get a broken pipe before decoding the HTTP status error — the test would see a transport error instead of the expected HTTP error status.

**Fix**: Added a drain loop that reads until the connection closes (`WouldBlock` on non-blocking read), ensuring the mock server consumes the full request body before sending the error response. This matches what a real HTTP server does and makes POST error-status tests reliable regardless of TCP segmentation.

- Commit: `f92f219` on fix/aw-client-rust-http-status-errors
- Tests pass: `cargo test -p aw-client-rust` — all 9 tests green (4 unit + 2 status_errors + 2 integration + 1 doctest)

TimeToBuildBob:
## CI format fix applied

The `cargo fmt -- --check` failure was a formatting nit in the drain loop I added in `f92f219` — the `.read_exact()` call needed multiline chaining.

**Fix**: Applied `cargo fmt` formatting in commit `436ed14` on `fix/aw-client-rust-http-status-errors`. CI should pass on the next run.

All tests still green locally: `cargo test -p aw-client-rust` — 9/9 pass.

PR #608, opened 2026-05-21
fix(android): load embedded server config

TimeToBuildBob:
## Summary
- load the embedded Android server config with `create_config(false)` instead of starting from `AWConfig::default()`
- make Android `get_config_dir()` return the app data dir so config-backed auth state can actually exist on-device
- keep device IDs in app data with a corrected comment explaining why

## Why
Android auth is still blocked before the `aw-android` WebView handoff step: the embedded `aw-server-rust` path currently bypasses `config.toml`, so `[auth].api_key` cannot be loaded or persisted on Android at all. This patch fixes that prerequisite and unblocks `ActivityWatch/aw-android#145`.

## Testing
- `cargo test -p aw-server`

## Notes
I could not run an Android-target build in this environment because only the Linux host target is installed here, so the Android-specific validation still needs device/SDK coverage on the `aw-android` side.

### DIFF
--- a/aw-server/src/android/mod.rs
+++ b/aw-server/src/android/mod.rs
@@ -38,7 +38,6 @@ pub mod android {
     use std::path::PathBuf;
     use std::sync::Mutex;
 
-    use crate::config::AWConfig;
     use crate::endpoints;
     use crate::endpoints::ServerState;
     use aw_client_rust::blocking::AwClient;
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
fix(android): added missing `None` to `create_config(false, None)` — the Android build should now compile.

Thanks @greptile-apps for catching this. The Linux build never exercises the android cfg block so it was invisible in local testing.

TimeToBuildBob:
Good catch @greptile-apps — the missing `None` argument was the Android build failure. Fixed in the latest push and Android CI now passes.

TimeToBuildBob:
Android CI is green and the Greptile compile fix is already in the latest push. This PR is ready to merge, but the `TimeToBuildBob` account does not have upstream merge permissions on `ActivityWatch/aw-server-rust`, so it needs a maintainer-side merge.

PR #611, opened 2026-05-21
fix(ci): harden dependabot automerge workflow

TimeToBuildBob:
This follows the merged aw-webui correction in ActivityWatch/aw-webui#837.

The failed master run shows `AWBOT_GH_TOKEN` currently returns `Bad credentials`, but `merge-me-action` also cannot use the built-in `GITHUB_TOKEN` in this `workflow_run` path because it queries branch protection over GraphQL and gets `Resource not accessible by integration`.

So the workflow still needs a live ActivityWatchBot PAT. This PR makes that explicit and hardens the surrounding workflow:

- keep `secrets.AWBOT_GH_TOKEN` for `merge-me-action` and document why
- remove the inert workflow-level `permissions` block, since the PAT-authenticated action does not use it
- restrict auto-merge to successful `pull_request` workflow runs from Dependabot
- pin `ridedott/merge-me-action` to commit SHA `2568f177a681785a874d6e7af950eb1a0dd31123`

Maintainer note: if the repo secret has not already been rotated, `AWBOT_GH_TOKEN` still needs to be updated in repository settings. The code-side `GITHUB_TOKEN` workaround is not viable for this action.

## Verification

- reproduced the failing master run: `gh run view 26257321530 --repo ActivityWatch/aw-server-rust --log-failed`
- validated the updated workflow YAML locally with `python3` + `yaml.safe_load`
- `git diff --check -- .github/workflows/dependabot-automerge.yml`

### DIFF
--- a/.github/workflows/dependabot-automerge.yml
+++ b/.github/workflows/dependabot-automerge.yml
@@ -1,5 +1,11 @@
 name: Dependabot Auto-merge
 
+# NOTE: `merge-me-action` still needs the bot PAT here.
+# Using `secrets.GITHUB_TOKEN` fails on `workflow_run` with
+# "Resource not accessible by integration" when the action queries
+# branch protection rules over GraphQL.
+# See: https://github.com/ridedott/merge-me-action/issues/1581
+
 on:
   workflow_run:
     types:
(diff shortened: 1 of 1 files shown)


ErikBjare:
See https://github.com/ActivityWatch/aw-webui/pull/837

TimeToBuildBob:
Addressed the aw-webui #837 mismatch in `8cfded6`.

This no longer tries to use `github.token` with `merge-me-action`; that path fails on the action's GraphQL branch-protection lookup. The PR now keeps `AWBOT_GH_TOKEN`, removes the inert permissions block, adds the `pull_request` workflow_run guard, pins the action SHA, and updates the PR body with the remaining maintainer-side requirement: rotate/update the repo secret if it has not already been fixed.

Local validation passed (`yaml.safe_load`, `git diff --check`). GitHub checks are running on the pushed commit.

PR #615, opened 2026-06-11
Performance improvements

0xbrayo:
Asked claude fable to identify some performance tweaks, I think these are worthwhile.

### DIFF
--- a/aw-datastore/src/worker.rs
+++ b/aw-datastore/src/worker.rs
@@ -19,8 +19,6 @@ use crate::DatastoreError;
 use crate::DatastoreInstance;
 use crate::DatastoreMethod;
 
-use mpsc_requests::ResponseReceiver;
-
 type RequestSender = mpsc_requests::RequestSender<Command, Result<Response, DatastoreError>>;
 type RequestReceiver = mpsc_requests::RequestReceiver<Command, Result<Response, DatastoreError>>;
 
@@ -83,15 +81,10 @@ pub enum Command {
     Close(),
 }
 
-fn _unwrap_response(
-    receiver: ResponseReceiver<Result<Response, DatastoreError>>,
-) -> Result<(), DatastoreError> {
-    match receiver.collect().unwrap() {
-        Ok(r) => match r {
-            Response::Empty() => Ok(()),
-            _ => panic!("Invalid response"),
-        },
-        Err(e) => Err(e),
+fn _unwrap_empty_response(response: Response) -> Result<(), DatastoreError> {
+    match response {
+        Response::Empty() => Ok(()),
+        _ => panic!("Invalid response"),
     }
 }
 
@@ -408,50 +401,50 @@ impl Datastore {
         Datastore { requester }
     }
 
+    /// Send a command to the worker thread and wait for its response.
+    ///
+    /// Fails with `InternalError` instead of panicking when the worker thread
+    /// is gone (e.g. it panicked on an earlier request), so callers such as
+    /// HTTP endpoints can degrade to a 5xx response instead of crashing the
+    /// request.
+    fn request(&self, cmd: Command) -> Result<Response, DatastoreError> {
+        let receiver = self.requester.request(cmd).map_err(|e| {
+            DatastoreError::InternalError(format!(
+                "Failed to send request, datastore worker is gone: {e:?}"
+            ))
+        })?;
+        receiver.collect().map_err(|e| {
+            DatastoreError::InternalError(format!(
+                "Failed to receive response, datastore worker died while handling request: {e:?}"
+            ))
+        })?
+    }
+
     pub fn create_bucket(&self, bucket: &Bucket) -> Result<(), DatastoreError> {
         let cmd = Command::CreateBucket(bucket.clone());
-        let receiver = self.requester.request(cmd).unwrap();
-        match receiver.collect().unwrap() {
-            Ok(_) => Ok(()),
-            Err(e) => Err(e),
-        }
+        _unwrap_empty_response(self.request(cmd)?)
     }
 
     pub fn delete_bucket(&self, bucket_id: &str) -> Result<(), DatastoreError> {
         let cmd = Command::DeleteBucket(bucket_id.to_string());
-        let receiver = self.requester.request(cmd).unwrap();
-        match receiver.collect().unwrap() {
-            Ok(r) => match r {
-                Response::Empty() => Ok(()),
-                _ => panic!("Invalid response"),
-            },
-            Err(e) => Err(e),
-        }
+        _unwrap_empty_response(self.request(cmd)?)
     }
 
     pub fn get_bucket(&self, bucket_id: &str) -> Result<Bucket, DatastoreError> {
         let cmd = Command::GetBucket(bucket_id.to_string());
-        let receiver = self.requester.request(cmd).unwrap();
-        match receiver.collect().unwrap() {
-            Ok(r) => match r {
-                Response::Bucket(b) => Ok(b),
-                _ => panic!("Invalid response"),
-            },
-            Err(e) => Err(e),
+        match self.request(cmd)? {
+            Response::Bucket(b) => Ok(b),
+            _ => panic!("Invalid response"),
         }
     }
 
     pub fn get_buckets(&self) -> Result<HashMap<String, Bucket>, DatastoreError> {
         let cmd = Command::GetBuckets();
-        let receiver = self.requester.request(cmd).unwrap();
-        match receiver.collect().unwrap() {
-            Ok(r) => match r {
-                Response::BucketMap(bm) => Ok(bm),
-                e => Err(DatastoreError::InternalError(format!(
-                    "Invalid response: {e:?}"
-                ))),
-            },
-            Err(e) => Err(e),
+        match self.request(cmd)? {
+            Response::BucketMap(bm) => Ok(bm),
+            e => Err(DatastoreError::InternalError(format!(
+                "Invalid response: {e:?}"
+            ))),
         }
     }
 
@@ -461,13 +454,9 @@ impl Datastore {
         events: &[Event],
     ) -> Result<Vec<Event>, DatastoreError> {
         let cmd = Command::InsertEvents(bucket_id.to_string(), events.to_vec());
-        let receiver = self.requester.request(cmd).unwrap();
-        match receiver.collect().unwrap() {
-            Ok(r) => match r {
-                Response::EventList(events) => Ok(events),
-                _ => panic!("Invalid response"),
-            },
-            Err(e) => Err(e),
+        match self.request(cmd)? {
+            Response::EventList(events) => Ok(events),
+            _ => panic!("Invalid response"),
         }
     }
 
@@ -478,25 +467,17 @@ impl Datastore {
         pulsetime: f64,
     ) -> Result<Event, DatastoreError> {
         let cmd = Command::Heartbeat(bucket_id.to_string(), heartbeat, pulsetime);
-        let receiver = self.requester.request(cmd).unwrap();
-        match receiver.collect().unwrap() {
-            Ok(r) => match r {
-                Response::Event(e) => Ok(e),
-                _ => panic!("Invalid response"),
-            },
-            Err(e) => Err(e),
+        match self.request(cmd)? {
+            Response::Event(e) => Ok(e),
+            _ => panic!("Invalid response"),
         }
     }
 
     pub fn get_event(&self, bucket_id: &str, event_id: i64) -> Result<Event, DatastoreError> {
         let cmd = Command::GetEvent(bucket_id.to_string(), event_id);
-        let receiver = self.requester.request(cmd).unwrap();
-        match receiver.collect().unwrap() {
-            Ok(r) => match r {
-                Response::Event(el) => Ok(el),
-                _ => panic!("Invalid response"),
-            },
-            Err(e) => Err(e),
+        match self.request(cmd)? {
+            Response::Event(el) => Ok(el),
+            _ => panic!("Invalid response"),
         }
     }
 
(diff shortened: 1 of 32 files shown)


0xbrayo:
@greptile-apps review

ErikBjare:
Nice! Any attempt at measuring the improvement?

0xbrayo:
Not yet, will do and report back.

0xbrayo:
# Performance test report: `performance-improvements` vs `master`

**Date:** 2026-06-12
**Branch under test:** `performance-improvements` (f3705be), baseline `master` (a9d0fcf)
**Test data:** byte-for-byte snapshot of the live database (98 MB, **544,688 events**, ~12.5 months of data, largest bucket `aw-watcher-window` with 309k events), taken with SQLite's online backup API and used as `sqlite-testing.db` in `--testing` mode (port 5666, release builds).

## Methodology

- The live DB was snapshotted to a golden copy; **each server run started from a fresh copy**, so master always saw schema v4 and the branch performed its v4→v5 migration on first start.
- Three binaries were benchmarked: `master`, a **midpoint** at de5a12c (mutex removal + deep-copy elimination, *before* the index change) to attribute gains per commit, and the full branch.
- Harness: Python over HTTP against `127.0.0.1:5666` — sequential latency benchmarks (warmup + N iterations, p50 reported), concurrent-throughput benchmarks (8/4 parallel clients), heartbeat write throughput, and a heartbeat-latency-under-query-load scenario.
- **Correctness check:** every benchmarked endpoint returned byte-identical response sizes on master and branch (same data, same ordering).

## Results

### REST endpoints (sequential, p50 latency)

| Benchmark | master | midpoint | branch | speedup (master→branch) |
|---|---|---|---|---|
| `GET /buckets/` | 0.24 ms | 0.24 ms | 0.24 ms | 1.0× |
| events `?limit=100` | 137.5 ms | 138.5 ms | **0.33 ms** | **417×** |
| events `?limit=1000` | 201.5 ms | 202.1 ms | **0.97 ms** | **208×** |
| events `?limit=10000` | 289.3 ms | 291.3 ms | **8.0 ms** | **36×** |
| events, 1-day range | 31.1 ms | 30.9 ms | 9.4 ms | 3.3× |
| events, 1-week range | 37.4 ms | 37.0 ms | 14.1 ms | 2.7× |
| events, 1-month range | 83.9 ms | 83.4 ms | 36.9 ms | 2.3× |
| events, mid-history week | 36.8 ms | 36.6 ms | 9.8 ms | 3.8× |
| event count (full bucket) | 30.7 ms | 30.7 ms | 6.9 ms | 4.5× |

### `/query` endpoint (realistic aw-webui desktop query: flood + AFK filter + merge by app)

| Time range | master | midpoint | branch | speedup |
|---|---|---|---|---|
| 1 day | 42.7 ms | 40.1 ms | 12.3 ms | 3.5× |
| 1 week | 56.2 ms | 49.4 ms | 19.8 ms | 2.8× |
| 1 month | 145.6 ms | 116.3 ms | 62.5 ms | 2.3× |
| 1 year | 1074 ms | 783 ms | 497 ms | 2.2× |
| browser query, 1 week | 13.0 ms | 11.3 ms | 3.9 ms | 3.3× |
| 7 × 1-day periods | 276.7 ms | 269.3 ms | 71.6 ms | 3.9× |

### Concurrency (8 parallel clients)

| Benchmark | master | midpoint | branch |
|---|---|---|---|
| day-queries, throughput | 23.8 req/s | 26.2 req/s | **100.1 req/s** (4.2×) |
| day-queries, p50 latency | 336 ms | 304 ms | **79 ms** (4.3×) |
| event-range fetches, throughput | 33.4 req/s | 33.5 req/s | **121.4 req/s** (3.6×) |

### Heartbeat latency while a heavy (year-range) query runs — the watcher-starvation scenario

| | master | branch |
|---|---|---|
| heartbeat p50 | **994 ms** | **0.64 ms** (~1500×) |
| heartbeat p95 | 1012 ms | 72 ms |
| heartbeats served in 12 s | 12 | 165 |
| concurrent queries completed | 13 | 28 |

On master every heartbeat waits for the entire in-flight query to release the global mutex. On the branch, heartbeats interleave with the query's datastore calls almost immediately.

### Write throughput (sequential heartbeats)

| | master | branch |
|---|---|---|
| distinct-data heartbeats | 3988/s | 4085/s (+2.4%) |
| merging heartbeats | 4330/s | 4438/s (+2.5%) |

### Startup / migration

- master startup: ~0.2 s.
- branch first startup (v4→v5 migration of 544k events): **~1.2 s, one-time**; subsequent startups: **0.07 s**.
- Index footprint: three single-column indexes (29.9 MB) → one composite index (15.3 MB), a 49% reduction in index space. The file grows by ~15 MB during migration because the new index is created before the old ones are dropped; the freed 29.9 MB stays on the freelist and is reused by future writes (or reclaimed by `VACUUM`).

## Attribution by commit

- **`c58421f` composite index (db v5)** — the dominant win for reads. The headline 417×/208× on `?limit=N` fetches comes from the query plan: v4 used `events_bucketrow_index` and then sorted all 309k bucket rows in a temp B-tree on every request (`USE TEMP B-TREE FOR ORDER BY`); v5 walks `(bucketrow, starttime DESC, endtime)` already in output order and stops at the LIMIT. Range fetches improve 2–4× by covering all three predicates in one index. This is what watchers' "get last event" polling and aw-webui bucket views hit constantly.
- **`de5a12c` deep-copy elimination** — visible as master→midpoint on `/query`: 6% (day) growing to **27% (year)**, since the redundant copies scaled with the amount of event data flowing through the query.
- **`9377bda` mutex removal** — barely moves identical-request throughput (the DB worker thread still serializes actual DB access), but it is the entire ~1500× heartbeat-under-load result: cheap requests no longer queue behind slow ones. This directly mitigates watcher data-loss/lag while dashboards run heavy queries.
- **`f3705be` prepared-statement cache** — marginal on the write path (+2.4%); per-request cost there is dominated by the channel round-trip and JSON handling, and statement re-prepare was a small fraction.

## Caveats

- Single machine (macOS, Apple Silicon), local loopback, one run per configuration; numbers are stable within runs (tight p50/p95) but treat small deltas (<10%) as noise.
- Heartbeat benchmarks used a scratch bucket; real watcher payloads are slightly larger.
- The v4→v5 migration is one-way; after running the branch against a DB, master can still open it but would recreate its old indexes only on a fresh DB (the migration does not downgrade).

## Artifacts

- Raw results and harness: `/tmp/aw-perf-test/` (`results-{master,midpoint,branch}.json`, `hb-under-load-*.json`, `bench.py`, `latency_under_load.py`, server logs, golden DB snapshot)
- Previous `sqlite-testing.db` preserved as `sqlite-testing.db.bak-pre-perf-test` next to the live DB.

ErikBjare:
Damn, nice! Ready to merge?

@greptileai review

0xbrayo:
Not yet, revisiting a WAL change that I had dropped earlier. Still more improvements on the table

ErikBjare:
I still think it could be worth merging this as is, given 5/5 Greptile and significant improvements already. Further improvements can be against the new baseline. Unless a second db migration is needed or something.

0xbrayo:
# Follow-up: three more improvements since the report above

Same harness, same 98 MB / 544k-event production snapshot as the report above. Three commits added since f3705be:

**c0bdc9d — v5 migration drops the old indexes before creating the composite.** As reported, the migration grew the db file by ~15 MB because the new index was built before the old indexes' pages were freed. Dropping first lets the same transaction reuse those pages — re-verified on the snapshot: file size is byte-identical before and after migration, `integrity_check` ok, migration still ~0.7 s.

**ddd68d4 — re-enabled HTTP keep-alive.** The benchmark client kept seeing connections dropped without a `Connection: close` header, which led to `keep_alive = 0` in config.rs — a 2020 workaround (7741a8a) for Rocket 0.4 issue #1254 that survived the Rocket 0.5 upgrade. Rocket 0.5's rewritten async core doesn't have that bug, so since then the workaround has only been forcing TCP setup/teardown on every single request from every watcher, and surprising HTTP clients that expect RFC-compliant keep-alive (python's `http.client` reports it as `RemoteDisconnected`). Graceful shutdown verified still instant (0.05 s) with idle keep-alive connections open.

**dbe2313 — WAL journal mode with explicit `synchronous=FULL`.** Each batch commit is now a single sequential WAL append+fsync instead of delete-mode's two fsyncs plus rollback-journal file churn, and commits no longer block readers (groundwork for read-connection parallelism later). Durability is unchanged: `synchronous=FULL` syncs the WAL on every commit, and the worker's existing batch-commit policy (15 s / 100 events / bucket change) is untouched. **No second db migration involved** — `journal_mode=WAL` is set at connection open and persists in the db header; any SQLite ≥ 3.7.0 (2010) can still open the file. In-memory datastores (tests, `--no-db`) are unaffected. Verified on the snapshot: WAL active, clean checkpoint on shutdown, integrity ok.

## Measured impact

| build | heartbeats/s (distinct data) | `GET events?limit=100` p50 |
|---|---|---|
| master | 3,988 | 137.5 ms |
| report above (f3705be) | 4,085 | 0.33 ms |
| + keep-alive (ddd68d4) | 5,572 | 0.22 ms |
| + WAL (dbe2313) | **5,815 (+42% vs report, +46% vs master)** | 0.22 ms |

Read-side numbers (range fetches, `/query`, concurrency) are unchanged from the report above — within run-to-run noise across all three builds.

One correction to the original report: the "branch startup 1.2 s (migration)" figure measured Rocket readiness, not the migration — the server starts accepting HTTP before the DB worker finishes migrating (`/api/0/info` responds immediately; the first datastore-touching request blocks until the migration commits, ~0.7 s on this snapshot).

0xbrayo:
I'll drop the new commits and make a new PR.

aw-datastore/src/worker.rs
```diff
@@ -138,6 +138,24 @@ impl DatastoreWorker {
...
+        // silently widen the loss window on power failure.
+        // In-memory databases ignore the request (journal_mode stays "memory").
+        let journal_mode: String = conn
+            .pragma_update_and_check(None, "journal_mode", "WAL", |row| row.get(0))
+            .expect("Failed to query journal_mode");
+        if !matches!(&method, DatastoreMethod::Memory()) && journal_mode != "wal" {
+            warn!("Failed to enable WAL (journal_mode={journal_mode}), continuing without it");
+        }
+        conn.pragma_update(None, "synchronous", "FULL")
+            .expect("Failed to set synchronous=FULL");
```
ErikBjare, line 157: Is this really a good idea? I though WAL with synchronous=NORMAL was best-practice for many small writes like the ones from watchers? Doesn't it lock during flush? Also accelerates SSD wear (might just be a small effect though)

0xbrayo:
Ready to merge

aw-datastore/src/worker.rs
ErikBjare: We can discuss in the new PR :)

0xbrayo: The datastore batches the writes, done every 15 seconds, WAL synchronous=FULL only makes it a single fsync.

Claude wrote: Erik's premise is right in general — WAL+NORMAL is the standard advice when every small write is its own
  transaction. But that advice doesn't transfer to this codebase, because of something we confirmed while
  making the change: aw-server already does group commit at the application level. The DB worker wraps all
  writes in one long-lived transaction and commits every 15 s / 100 events / on bucket changes (worker.rs work
  loop). The per-commit fsync that NORMAL exists to eliminate happens here about once per 15 seconds, not once
  per heartbeat.

  Taking his three concerns one at a time:

  "Best practice for many small writes" — the many small watcher writes never individually hit the WAL or an
  fsync; they accumulate in the open transaction. synchronous only matters at commit time, and commits are on a
  15s cadence. The benchmark backs this up: going from delete+FULL to WAL+FULL increased throughput (5,572 →
  5,815 heartbeats/s) while hammering writes thousands of times faster than real watchers do. The fsync simply
  isn't in the per-write path. NORMAL would save ~4 fsyncs per minute — unmeasurable.

  "Doesn't it lock during flush?" — no, that's the part WAL specifically fixes. The commit fsync is performed
  by the writer while holding only the WAL write lock; readers keep reading their snapshot and never block (in
  delete mode they did block — that's one of the reasons we switched). The only thing waiting on the flush is
  the single writer thread itself, which is inside its own commit regardless of sync mode, and there are no
  other writers to contend with.

  SSD wear — compared to what shipped before (delete mode at SQLite's default FULL: two fsyncs per commit plus
  rollback-journal file create/delete churn every 15s), WAL+FULL is strictly less sync and file churn, not
  more. NORMAL would reduce it further, but from "a few fsyncs per minute" to "slightly fewer" — genuinely
  negligible either way, as he suspected.

  And the cost side of NORMAL, which the general advice tends to gloss over: with NORMAL the WAL is only synced
  at checkpoints, and autocheckpoint triggers on page count (1000 pages), not time. At AW's trickle write rate
  that can be a very long time — a power loss (the relevant failure for laptops; app crashes lose nothing
  under either mode) could drop not just the current 15s batch but potentially hours of activity since the last
  checkpoint. To bound that you'd add a time-based wal_checkpoint on a timer — at which point you're fsyncing
  on a schedule again and have rebuilt FULL with extra steps, since commits here are already timer-driven.

  So my recommendation stands: FULL preserves the exact durability the server has always had, with strictly
  fewer syncs than before, and NORMAL's upside is ~zero because of the existing batching. NORMAL wouldn't be
  wrong — losing some window-tracking data on a power cut isn't catastrophic — it just buys nothing here.

  Want me to post this as a reply on the PR (trimmed to comment length)? I'd lead with "good instinct in
  general, but the worker already group-commits every 15s, so the fsync isn't per-write" since that's the fact
  that resolves all three concerns.

ErikBjare: That's fair, FULL it is then.

PR #618, opened 2026-06-13
feat: zstd compression

### LINKED ISSUE #493: Transparent compression of database
With https://github.com/phiresky/sqlite-zstd we could select columns (e.g. events.data) that will get compressed on the fly. I think it would make sense to add this (perhaps as opt-in), what do you think?

0xbrayo:
closes #493

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -232,6 +232,7 @@ dependencies = [
  "serde",
  "serde_json",
  "zeroize",
+ "zstd",
 ]
 
 [[package]]
@@ -417,6 +418,8 @@ version = "1.2.30"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "deec109607ca693028562ed836a5f1c4b8bd77755c4e132fc5ce11b0b6211ae7"
 dependencies = [
+ "jobserver",
+ "libc",
  "shlex",
 ]
 
@@ -1555,6 +1558,16 @@ name = "jni-sys"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8eaf4bc02d17cbdd7ff4c7438cafcdf7fb9a4613313ad11b4f8fefe7d3fa0130"
+
+[[package]]
+name = "jobserver"
+version = "0.1.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9afb3de4395d6b3e67a780b6de64b51c978ecf11cb9a462c66be7d4ca9039d33"
+dependencies = [
+ "getrandom 0.3.3",
+ "libc",
+]
 
 [[package]]
 name = "js-sys"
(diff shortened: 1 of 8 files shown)


0xbrayo:
> 🤖 _Claude, on behalf of @0xbrayo_

@greptile review

0xbrayo:
> 🤖 _Claude, on behalf of @0xbrayo_

## Benchmark: dictionary compression on a real-world database

**Test data:** byte-for-byte snapshot of a live database, 544,673 events across 19 buckets (~12.5 months of `aw-watcher-window`/`afk`/etc data).

**Methodology:**
1. Started from the v4 snapshot and migrated it to **v5** using the pre-compression code (commit `6f6c64a`, i.e. current `master` schema), then ran `VACUUM` to get a clean baseline size.
2. Took that fresh v5 database and ran it through this branch's v5→v6 migration (`--features compression-zstd`), then `VACUUM`ed again for a clean "after" size.

**Results:**

| | size |
|---|---|
| v5 (master, uncompressed) | 72,458,240 bytes (69.1 MiB) |
| v6 (this PR, dictionary-compressed) | 46,047,232 bytes (43.9 MiB) |
| **reduction** | **36.4%** |

Other notes:
- 431,310 / 544,673 events (79%) ended up dictionary-compressed; the rest were small enough that storing them raw was already smaller, so they're kept as-is.
- One-time migration + dictionary training (64 KiB dict, level 6) took **~3 seconds** for all 544,673 events, wrapped in a single transaction.
- Verified correctness: read events back through the REST API after migration — data round-trips intact (`app`/`title` fields, durations, timestamps all correct).

0xbrayo:
> 🤖 _Claude, on behalf of @0xbrayo_

## Why a custom implementation instead of `sqlite-zstd`?

[`sqlite-zstd`](https://github.com/phiresky/sqlite-zstd) (phiresky) implements essentially the same idea — dictionary-trained zstd compression of a column — and was the inspiration for this approach in #493. We didn't pull it in directly for a few reasons specific to this codebase:

- **License**: `sqlite-zstd` is LGPL-2.0; this project's other dependencies are permissively licensed, so avoiding it sidesteps that friction.
- **Schema fit**: it works by moving the target column into a shadow `_zstd` table and replacing the original table with a view (plus triggers). That would conflict with our existing hand-rolled migration system (`_migrate_vX_to_vY`, the `ON DELETE CASCADE` FK, the `(bucketrow, starttime)` index) and complicate the query layer.
- **Operational simplicity**: `sqlite-zstd` expects periodic `zstd_incremental_maintenance()` calls to retrain dictionaries and vacuum shadow tables. Here, the dictionary is trained once on first upgrade and reused — no background maintenance task needed.
- **Android**: this implementation is just the `zstd` crate operating on raw bytes in a normal column, with no SQLite virtual tables/extension registration — the simplest thing to guarantee works with the bundled SQLite used on Android.

The underlying technique (train a shared dictionary once, compress every row against it) is the same; it's just fit directly into this project's existing schema/migration model rather than taken as a generic extension.

PR #619, opened 2026-06-13
Fix cargo warnings

0xbrayo:


aw-sync/src/accessmethod.rs
```diff
@@ -59,9 +58,6 @@ impl AccessMethod for Datastore {
     fn get_event_count(&self, bucket_id: &str) -> Result<i64, String> {
         Ok(Datastore::get_event_count(self, bucket_id, None, None).unwrap())
     }
-    fn close(&self) {
-        Datastore::close(self);
-    }
```
ErikBjare, line 64: Uhh, do we really never do this? Seems problematic/an oversight? @TimeToBuildBob wdyt?

TimeToBuildBob: Agreed this looks like an oversight at first glance — but the close still happens, just not through the trait.

`sync_run` closes concrete `Datastore`s:

```rust
for ds_from in &ds_remotes {
    ds_from.close();  // Datastore::close, worker-thread shutdown
}
ds_localremote.close();
```

`sync_datastores` / `log_buckets` take `&dyn AccessMethod` and never call `close`. The `AwClient` impl was a NOP. So the trait method was never used as a trait method.

`AccessMethod` is the bucket/event I/O surface (get/insert/heartbeat). Closing a Datastore is worker-lifecycle (`Command::Close()` + join); closing an HTTP client is meaningless. Putting `close()` on the trait mixed those.

**Verdict:** keep the trait removal. The PR does not drop Datastore shutdown — it just stops pretending `AwClient` has one. If we ever need "release this AccessMethod", that should be a Datastore-only API (already `Datastore::close()`), not a trait method.

Not going to restore it unless you want the trait to keep a documented no-op for API stability.

PR #628, opened 2026-07-02
feat(android): add migration for aw-watcher-android-test bucket names

TimeToBuildBob:
Implements the `aw-watcher-android-test` -> `aw-watcher-android` bucket name migration requested in #577 (comment) and ActivityWatch/aw-android#150, plus the `migrate_hostname()` and `rename_bucket()` infrastructure from #577 with the JNI naming fix already applied.

## Changes

- `aw-datastore/src/datastore.rs`: `rename_bucket()`, `migrate_hostname()`, `migrate_test_bucket_names()`
- `aw-datastore/src/worker.rs`: `RenameBucket`, `MigrateHostname`, `MigrateTestBucketNames` commands + public API
- `aw-server/src/android/mod.rs`: Three JNI entry points using camelCase naming (no JNI `_1` encoding issues)

### `migrate_test_bucket_names()`

Renames all `aw-watcher-android-test*` buckets to `aw-watcher-android*` using safe prefix-only replacement. Uses `OR IGNORE` to skip UNIQUE collisions gracefully, and `SUBSTR` prefix replacement rather than global `REPLACE`.

## Status

- CI green (8/8 checks: ubuntu, windows, macOS, Android, clippy, format, coverage, Greptile)
- Both Greptile review findings fixed and threads resolved (P1: `OR IGNORE`, P2: `SUBSTR` prefix swap)
- JNI naming uses camelCase (no `_1` encoding issues)
- Includes commits from #577 with JNI fix already applied -- independently mergeable

Ref: ActivityWatch/aw-android#150, #577

PR #632, opened 2026-07-14
fix(android): update JNI androidQuery bucket name to aw-watcher-android

### LINKED ISSUE #191: Widget not displaying data.
The widget is not showing any screen time information ever since I updated to the latest version. Probably a regression caused by one of the last commits in #139. Initially worked then stopped, probably after a restart. Will investigate.

0xbrayo:
Closes https://github.com/ActivityWatch/aw-android/issues/191

This fixes an issue introduced in #628 where the legacy '-test' suffix was dropped from buckets but not from the query.

### DIFF
--- a/aw-server/src/android/mod.rs
+++ b/aw-server/src/android/mod.rs
@@ -343,8 +343,8 @@ pub mod android {
             Err(err) => return create_error_object(&env, err.to_string()),
         };
 
-        // Hardcoded bucket ID for testing
-        let bid_android = "aw-watcher-android-test".to_string();
+        // Hardcoded bucket ID
+        let bid_android = "aw-watcher-android".to_string();
 
         // Get classes from server settings via HTTP API
         let classes = match AwClient::new("127.0.0.1", 5600, "aw-android-query") {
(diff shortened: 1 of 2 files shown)


0xbrayo:
@TimeToBuildBob  review

ErikBjare:
@greptileai review

TimeToBuildBob:
Code review looks good to me. The two changes complement each other well:

1. **`aw-server/src/android/mod.rs`** — correctly updates the hardcoded query bucket from the legacy `aw-watcher-android-test` to `aw-watcher-android`, aligning it with the name change introduced in #628.

2. **`aw-transform/src/find_bucket.rs`** — the refactored `find_bucket` logic is an improvement: it now prefers an exact ID match and only falls back to the first prefix match if no exact match is found. This prevents a bucket named `aw-watcher-android-test` from being returned when looking up `aw-watcher-android` (which would have been the old behavior once both buckets exist on a device that upgraded from the test name).

CI is green across all platforms and Greptile gave it 5/5. LGTM. Merge is Erik's call.

PR #634, opened 2026-07-20
perf: cache compiled regex in categorization rules

0xbrayo:


0xbrayo:
I discovered the was a performance issue with uncached regexes when running https://github.com/0xbrayo/aw-work-sync every 60 seconds (overkill I know). I'll measure the speedup/memory impact of caching with the aw-webui queries and report back.

0xbrayo:
@greptile review @TimeToBuildBob

ErikBjare:
Good idea, but I don't throwing Arc on is right?

Simpler to use a lru/getter-style regex compilation cache? Python regex does this natively iirc (512 LRU).

TimeToBuildBob:
Good catch on the regex recompilation overhead. Aligning with Erik's suggestion — an LRU cache is cleaner here than the custom TTL+size combo.

**On the `Arc<Regex>`**: worth keeping. `fancy_regex::Regex` doesn't implement `Clone`, so `Arc` is the right way to hand out a reference from the cache without recompiling. The indirection cost is negligible compared to regex compilation.

**What I'd change**: replace `HashMap<String, CacheEntry>` + the manual cleanup block with [`lru::LruCache<String, Arc<Regex>>`](https://crates.io/crates/lru). That gives a fixed-size bound (512 matches Python's default) with automatic LRU eviction — no `Instant` tracking, no TTL logic, no `retain` call:

```rust
use lru::LruCache;
use std::num::NonZeroUsize;
use std::sync::{Arc, Mutex, OnceLock};

static REGEX_CACHE: OnceLock<Mutex<LruCache<String, Arc<Regex>>>> = OnceLock::new();

fn cached_regex(pattern: &str) -> Result<Arc<Regex>, fancy_regex::Error> {
    let cache = REGEX_CACHE
        .get_or_init(|| Mutex::new(LruCache::new(NonZeroUsize::new(512).unwrap())));
    let mut lru = cache.lock().unwrap();
    if let Some(re) = lru.get(pattern) {
        return Ok(Arc::clone(re));
    }
    let re = Arc::new(Regex::new(pattern)?);
    lru.push(pattern.to_string(), Arc::clone(&re));
    Ok(re)
}
```

This also naturally resolves the Greptile P1 concern about unbounded cache growth — LRU eviction hard-caps memory usage.

One contextual note: categorization patterns typically come from static user config (a small, stable set), so the current >1000 threshold would rarely trigger in practice. But an explicit bounded LRU is a better contract and simpler code regardless.

ErikBjare:
@greptileai review

PR #636, opened 2026-07-28
fix(auth): API key bypass via percent-encoded paths

TimeToBuildBob:
## Summary

The `api_key` auth gate (added in #585, closing #494) can be bypassed completely — read *and* write — with a single `curl` and no credentials:

```
curl http://127.0.0.1:5666/api/0/buckets/     -> 401 Missing or invalid API key
curl http://127.0.0.1:5666/%61pi/0/info       -> 200, full server info, no auth header
curl -X POST http://127.0.0.1:5666/%61pi/0/buckets/poc -d '{...}'  -> 200, bucket created
```

## Root cause

`ApiKeyCheck::on_request` compared the **raw, still-percent-encoded** path:

```rust
let path = request.uri().path().as_str();
let normalized_path = format!("/{}", path.trim_start_matches('/'));
if !normalized_path.starts_with("/api/") { return; }   // sees "/%61pi/..." -> skips auth
```

Rocket's router matches on the **decoded** segment view instead — `rocket::router::collider::paths_match` uses `req.uri().path().segments()`, which percent-decodes and skips empty segments. So the gate and the router disagreed about what the path is: the gate saw `/%61pi/0/buckets/` (not an API path, skip auth), the router saw `api / 0 / buckets` and dispatched to the real handler.

This is the same desync as the `//api/...` bypass fixed in #588. That fix patched one *spelling* of the disagreement (leading slashes) while leaving the disagreement itself in place — which is why a second variant surfaced. Reasoning about encodings one at a time invites a third (double encoding, unicode tricks, ...).

## Fix

Match on exactly the view the router dispatches on:

```rust
let segments: Vec<&str> = request.uri().path().segments().collect();
if segments.first() != Some(&API_SEGMENT) { return; }
if PUBLIC_PATHS.contains(&segments.as_slice()) { return; }
```

Percent-decoding and empty-segment skipping now happen in the *same code path* for gate and router, so double slashes, encoded characters, and double encoding all resolve identically on both sides. This subsumes the #588 fix rather than layering on it (its regression test is kept and still passes). Public-path exemptions move to decoded segment slices for the same reason.

## Tests

Three regression tests added. All three **fail against the pre-fix gate** and pass after — verified by reverting the gate logic locally and re-running:

```
test_api_key_percent_encoded_paths_require_auth ... FAILED   <- pre-fix
  assertion `left != right` failed: auth bypassed via encoded path /%61pi/0/buckets/
test_api_key_percent_encoded_write_requires_auth ... FAILED  <- pre-fix
  assertion `left != right` failed: write bypassed via encoded path
test_public_path_matched_on_decoded_segments ... FAILED      <- pre-fix
```

Coverage: encoded first char, fully encoded segment, mid-segment encoding, encoding combined with the #588 double-slash trick, POST/write bypass (plus confirming the bucket was not created), and decoded-segment parity for the `/api/0/info` public exemption.

`cargo test -p aw-server` green (32 tests), `cargo fmt --check` clean, no new clippy warnings.

## Impact

`api_key` was added specifically so other local processes/users can't reach the API (refs ActivityWatch/activitywatch#32, ActivityWatch/activitywatch#1199). This bug defeated exactly that threat model, so I'd agree with the reporter's **critical** rating — full read/write with no tooling and no credentials.

Reported by pal0x, who also flagged that the right fix here is structural rather than per-encoding. Agreed, and that's what this does.

### DIFF
--- a/aw-server/src/endpoints/apikey.rs
+++ b/aw-server/src/endpoints/apikey.rs
@@ -241,10 +241,11 @@ mod tests {
         let server = setup_testserver(Some("secret123".to_string()));
         let client = rocket::local::blocking::Client::tracked(server).expect("valid instance");
 
-        // Each of these decodes to a real /api/ route and must be gated.
+        // Each of these decodes to the real /api/0/buckets/ route, so each must
+        // be rejected by the auth gate specifically — asserting "not 200" would
+        // also pass on a 404, hiding a route miss as an auth success.
         for path in [
             "/%61pi/0/buckets/",     // lowercase hex, first char
-            "/%41PI/0/buckets/",     // uppercase — decodes to /API/, not a route
             "/ap%69/0/buckets/",     // encoded char mid-segment
             "/%61%70%69/0/buckets/", // fully encoded segment
             "//%61pi/0/buckets/",    // encoding combined with the #588 double-slash trick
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Verified end-to-end over real HTTP (not just the in-process test client), both directions, against a server started with `api_key = "secret-poc-key"` in `config-testing.toml`.

**Unpatched master (`59eaae3`)** — reporter's PoC reproduces exactly:

```
GET  /api/0/buckets/                 -> 401   (honest path correctly gated)
GET  /%61pi/0/info                   -> 200   bypass
GET  /%61pi/0/buckets/               -> 200   bypass
POST /%61pi/0/buckets/authbypass-poc -> 200   bypass, bucket created
```

Bucket listing afterwards confirms the write landed with no credentials:
```json
{"authbypass-poc":{"id":"authbypass-poc","type":"test","client":"authbypass-poc",...}}
```

**This branch:**

```
GET  /api/0/buckets/                 -> 401
GET  /%61pi/0/buckets/               -> 401
GET  //api/0/buckets/                -> 401   (#588 case, still covered)
GET  /ap%69/0/buckets/               -> 401
GET  /%61%70%69/0/buckets/           -> 401
POST /%61pi/0/buckets/authbypass-poc -> 401
GET  /api/0/buckets/  + valid key    -> 200   (not over-blocking)
GET  /api/0/info                     -> 200   (public exemption intact)
GET  /%61pi/0/info                   -> 200   (public exemption, decoded — router parity)
```

Bucket listing afterwards: `{}` — the POC write was refused.

Worth noting the last line: pre-fix, `/%61pi/0/info` was *also* mismatched, just in the harmless direction (gate 401'd a request the router would have sent to the public info handler). Gate and router now agree in both directions, which is the actual invariant.

TimeToBuildBob:
Good catch on the P2 — fixed in 2003e5a.

`assert_ne!(status, Ok)` would indeed have been satisfied by a 404, so a route miss could have passed as an auth success. The encoded-path cases now assert `Status::Unauthorized` exactly, and I added a positive control: the same encoded paths return **200 with a valid key**, which is what actually proves the 401s come from the gate rather than from the router failing to match.

`/%41PI/0/buckets/` was the specific case you flagged — it decodes to `/API/`, which isn't a registered route, so it 404s at the router and never exercises the gate. Split it out with an explicit `Status::NotFound` assertion and a comment, so it still documents that routing is case-sensitive (a future case-insensitive change would fail here) without masquerading as an auth test.

Also tightened the post-write bucket lookup and the public-path test to exact statuses for the same reason. Re-verified all three regression tests still fail against the pre-fix gate.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:

Issue #478, opened 2024-04-27
aw-sync old libssl dependency


YeapGuy:
I’m unable to run aw-sync on an Ubuntu 22.04-based system:
```
./aw-sync: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
```
22.04 ships with OpenSSL 3, I don’t have 1.1

nfacha:
Also facing this issue on Ubuntu 23.10

```
./aw-sync: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
```

PR #641, opened 2026-07-29
build(docker): add Dockerfile for building a server image

almirb:
Multi-stage build: compiles aw-webui and aw-server, then ships only the release binary on debian:bookworm-slim with libssl3.

Only aw-server is built, not aw-sync: on Linux the latter pulls in openssl with the "vendored" feature and compiles OpenSSL from source, which is not needed by the image.


ErikBjare:
We have rejected similar PRs in the past, as the Dockerfiles aren't being used by maintainers or in CI and so not something we want to maintain or bless/suggest we support.

If you explain your usecase I might change my mind.

almirb:
> We have rejected similar PRs in the past, as the Dockerfiles aren't being used by maintainers or in CI and so not something we want to maintain or bless/suggest we support.
> 
> If you explain your usecase I might change my mind.

Fair enough — I understand not wanting to bless or maintain something you
don't use. Let me describe the use case, and also scope down what I'm
actually asking for.

**Use case:** I run aw-server-rust as an always-on central instance on a
headless Ubuntu server, aggregating data from [N] machines that sync into
it. This is essentially the setup aw-sync's README already describes (a
central instance + multidevice query), just permanent rather than a
temporary testing instance. On a headless server managed by Portainer,
a container is the natural unit of deployment — there is no aw-qt, no
desktop session, no user systemd unit to hang it off.

**Why this might matter to the project:** the absence of a Dockerfile
doesn't mean people don't run this in Docker — it means they run someone
else's image. I was pulling `brain40/aw-server-rust:latest` from Docker
Hub for months: an opaque binary, built by a stranger, from unknown
source, handling all of my activity data. For a project whose entire
value proposition is privacy and data ownership, pushing self-hosters
toward unauditable third-party images seems like the worse outcome. A
Dockerfile in-tree lets people build it themselves and verify what they
run.

**What I'm not asking for:** no official image on Docker Hub, no release
pipeline, no support promise, no badge in the README. Just the build
recipe in-tree.

To address the maintenance concern, I'm happy to do any of:

- Mark it explicitly as community-contributed and unsupported in the
  README, so it carries no implication of official support.
- Put it under `contrib/` instead of the repo root.

One thing the containerized build surfaced, incidentally: `make build`
fails in a clean environment because aw-sync pulls in openssl with the
`vendored` feature (#478) and compiles OpenSSL from source. My Dockerfile
sidesteps it by building only the `aw-server` target. A clean-room build
tends to expose this kind of friction that a developer machine hides.

If the answer is still no, that's completely fine — I'll keep it in my
fork. In that case, would you be open to a line in the docs pointing to
community-maintained images? That's zero maintenance for you and still
better than people finding random images on Docker Hub.

That's a very nice project!  Congrats!

ErikBjare:
@almirb Putting it in `scripts/` would be good, also addressing Greptile's findings. I'll accept it if fixed, I agree it's better to have some kind of official reference which we can iron out the worst mistakes in.

almirb:
> @almirb Putting it in `scripts/` would be good, also addressing Greptile's findings. I'll accept it if fixed, I agree it's better to have some kind of official reference which we can iron out the worst mistakes in.

Done! Could you check it again?

Thanks!

ErikBjare:
@greptileai review

Issue #649, opened 2026-08-06
aw-sync: move sync provenance out of the bucket ID into trustworthy bucket metadata


TimeToBuildBob:
Split out of #647 / [discussions#1373](https://github.com/orgs/ActivityWatch/discussions/1373), where @ErikBjare noted the `-synced-from-HOST` bucket-ID suffix is "ugly and not a clean solution long-term, but was adopted to prevent accidents". Filing the design question separately so #648 stays a narrow bugfix.

## What's true today

- Provenance lives in the **bucket ID**: `aw-watcher-window_HOSTB-synced-from-HOSTB`. The hostname appears twice — once from the watcher's naming convention, once from sync.
- **`$aw.sync.origin` exists but is not trustworthy.** `get_or_create_sync_bucket()` writes it in the `NoSuchBucket` branch regardless of direction, so it's set on a host's own *push-staged* buckets too. It does not mean "second-hand data". Same defect as the existing `FIXME: "-synced" should only be appended when synced to the local database, not to the staging area for local buckets`.
- **aw-webui master has no notion of synced buckets** — `git grep synced` over `origin/master -- src/` returns nothing. They render as raw IDs next to local ones, which is what the user in #1373 actually complained about.
- Because the attribute is unreliable, **the ID suffix is currently the only dependable marker** of second-hand data — #648 has to match on it. So the suffix can't just be dropped; something else has to become authoritative first.

## The suffix is doing two jobs

1. **Namespacing** — an imported bucket must not collide with a local one and silently merge foreign events. Real, still needed.
2. **Provenance** — "where did this come from?". The ID is the wrong home: forces string-parsing, duplicates the hostname, leaks into every UI surface and query.

Proposal keeps (1), moves (2) to metadata.

## Suggested staging

**Stage 1 — make `$aw.sync.origin` authoritative.** Write it only on import, never on push-staging (resolves the FIXME). Derive it from the ID suffix for existing buckets. Then `is_synced_bucket()` in #648 switches to the attribute with the ID check as legacy fallback.

Low risk, no ID change, no migration, no UI change — and independently correct.

**Stage 2 — teach aw-webui about origin.** This is where the user-visible win is, and it needs no ID change:
- "synced from HOSTB" badge read from metadata, not regex
- a **host/device filter** on Timeline and Activity — the actual fix for the #1373 complaint, since even *correct* sync interleaves HOSTB's data into HOSTA's timeline by default
- group timeline rows by origin host, local first
[...]

PR #650, opened 2026-08-08
fix(aw-sync): only write $aw.sync.origin on pull, not push-staging

TimeToBuildBob:
## Problem

`get_or_create_sync_bucket()` always wrote `$aw.sync.origin` to the destination bucket — even when `is_push=true` (staging local buckets into the sync folder). This caused staging copies to be incorrectly marked as if they were synced-from-remote buckets.

## Fix

Refactor the function to compute `(new_id, sync_origin)` together:

- **Push path**: `sync_origin = None` → destination bucket is created with its original ID and **no** `$aw.sync.origin` stamp.
- **Pull path**: `sync_origin = Some(origin)` → the derived origin (from existing `$aw.sync.origin` metadata, or hostname as legacy fallback) is written into the destination bucket.

This also handles buckets that predate the `$aw.sync.origin` field naturally: the hostname fallback is used consistently for both the ID construction and the metadata stamp, so legacy buckets are correctly annotated on their first pull.

## Testing

Added `test_sync_origin_metadata` covering both invariants:
- push: staging bucket has **no** `$aw.sync.origin`
- pull: imported bucket has `$aw.sync.origin` set to the source hostname

All 6 aw-sync tests pass.

## Related

- Issue #649: move sync provenance out of bucket ID into trustworthy bucket metadata (Stage 1)

### DIFF
--- a/aw-sync/src/sync.rs
+++ b/aw-sync/src/sync.rs
@@ -194,10 +194,13 @@ fn get_or_create_sync_bucket(
     ds_to: &dyn AccessMethod,
     is_push: bool,
 ) -> Bucket {
-    let new_id = if is_push {
-        bucket_from.id.clone()
+    // On pull/import: derive the origin from $aw.sync.origin metadata (preferred) or the
+    // hostname (legacy fallback for buckets that predate the metadata field).  On push-staging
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #657, opened 2026-08-25
perf(aw-transform): cache categorize results per event data to fix month-view timeout

### LINKED ISSUE #629: AxiosError: timeout of 30000ms exceeded while trying to see activity for one month
While trying to view my activity for 2026-06 at http://localhost:5600/#/activity/your_hostname_here/month/2026-06-01/view/summary
I get
`AxiosError: timeout of 30000ms exceeded. See dev console (F12) and/or server logs for more info.` and my views hang with "Loading..." in them. Nothing particularly interesting after pressing F12.

1. My ~/.local/share/activitywatch/aw-server-rust/sqlite.db is 168 MB big and contains data since 2024-01.
2. After https://github.com/ActivityWatch/aw-server-rust/pull/615 it's not better.
3. I'm on b199671.
4. This is a low end machine.
5. aw-server process gets from 20 MB in RAM to something over 100 MB when trying to display the report. It looks like it's trying to load whole database instead of only the time period that I'm interested in.

This is kind of a wish than a bug: Could ActivityWatch be made so that it doesn't start to time out on reports when there is more data in the database? I remember watching 1 month time periods without problem earlier on.

TimeToBuildBob:
## Summary

Fixes #629 — month-summary view timing out (30 s) on large databases (168 MB, 2+ years of data).

### Root cause

`categorize()` in `aw-transform/src/classify.rs` was processing every event individually against every regex rule. For a typical month with 50 000+ events (from aw-watcher-window) and 20 category rules, this meant ~1 000 000 regex evaluations per query.

The vast majority of those events share identical `app`+`title` data — heartbeat-based watchers emit the same payload many times per second. So the same regex match was being recomputed thousands of times for no reason.

### Fix

Added a `HashMap<String, Vec<String>>` category cache inside `categorize()`, keyed on the serialized event data (`serde_json::to_string(&event.data)`). Only the first occurrence of each distinct data fingerprint is matched against the rule set; subsequent events with identical data reuse the cached result in O(1).

`serde_json::Map` preserves insertion order (IndexMap semantics), so events produced by the same watcher in the same session produce a consistent JSON key without any normalization step.

### Expected impact

For a month's worth of data with 50 000 events but only ~200 distinct app/title pairs:
- **Before**: 50 000 × 20 = 1 000 000 regex evaluations
- **After**: 200 × 20 = 4 000 regex evaluations (~250× less)

The 30-second timeout on a 168 MB database should become a sub-second query.

### Changes

- `aw-transform/src/classify.rs`: cache added to `categorize()`; dead `categorize_one()` helper removed (logic inlined into the cached loop)
- New test `test_categorize_cache_correctness`: 101 events (50 + 1 + 50), two distinct data shapes → verifies correct category for each shape and no false cache collisions

### Testing

```
cargo test -p aw-transform --lib
# 42 passed; 0 failed
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -257,6 +257,7 @@ dependencies = [
  "fancy-regex",
  "log",
  "plex",
+ "proptest",
  "serde",
  "serde_json",
 ]
@@ -391,9 +392,9 @@ checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
 
 [[package]]
 name = "bitflags"
-version = "2.9.1"
+version = "2.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1b8e56985ec62d17e9c1001dc89c88ecd7dc08e47eba5ec7c29c7b5eeecde967"
+checksum = "b588b76d00fde79687d7646a9b5bdf3cc0f655e0bbd080335a95d7e96f3587da"
 
 [[package]]
 name = "block-buffer"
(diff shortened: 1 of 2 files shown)


ErikBjare:
@TimeToBuildBob Do the same PR for aw-server-python/aw-core

TimeToBuildBob:
Done — ActivityWatch/aw-core#150 applies the same cache to `categorize()` and `tag()` in `aw_transform/classify.py`.

One note on this PR's CI: the macOS failure is a pre-existing SQLite schema issue in `aw-datastore` (`"duplicate column name: data"` in `test_push_does_not_reexport_synced_buckets`) — it's in the `aw-sync` crate and unrelated to the `aw-transform` changes here.

PR #661, opened 2026-08-27
fix(datastore): merge disjoint aw-watcher-android-test events on name collision

TimeToBuildBob:
## Summary

`migrate_test_bucket_names()` (from #628) renamed `aw-watcher-android-test*` buckets with `UPDATE OR IGNORE`. That no-ops when the canonical `aw-watcher-android_*` bucket already exists — exactly the post-upgrade state in ActivityWatch/aw-android#243, where years of history sit in the old bucket and the Activity view queries the new empty-ish one.

This replaces the rename-only path with:

1. **Rename** when the destination does not exist.
2. **Merge disjoint events** when both buckets exist: move legacy events that do not overlap any destination event.
3. **Leave overlapping events** in the legacy bucket so we never create duplicate activity records. Delete the legacy bucket only once it is empty.

A single overlapping cutover heartbeat no longer strands the rest of the history.

Also rebuilds the in-memory bucket cache from SQLite after the rewrite so deleted buckets disappear (the previous insert-only refresh kept stale IDs).

## Why not skip the whole bucket on any overlap?

Erik's constraint in ActivityWatch/aw-android#243 is merge when it is safe/non-overlapping. Skipping the entire legacy bucket because one event overlaps at the upgrade boundary would leave years of disjoint history invisible. Partial merge matches that constraint without throwing away the rest.

## Tests

- `test_migrate_test_bucket_names_renames_bucket_and_preserves_events`
- `test_migrate_test_bucket_names_merges_non_overlapping_buckets`
- `test_migrate_test_bucket_names_keeps_overlapping_buckets_separate`
- `test_migrate_test_bucket_names_merges_interleaved_non_overlapping_events`
- `test_migrate_test_bucket_names_moves_disjoint_events_when_some_overlap` (new)

`cargo test -p aw-datastore --test datastore` — 16 passed.

## Follow-up

ActivityWatch/aw-android still never *calls* this JNI. Companion PR will wire `migrateWatcherAndroidBucketNames()` from `BackgroundService` startup.

Refs: ActivityWatch/aw-android#149, ActivityWatch/aw-android#150, ActivityWatch/aw-android#243

### DIFF
--- a/aw-datastore/src/datastore.rs
+++ b/aw-datastore/src/datastore.rs
@@ -318,10 +318,11 @@ impl DatastoreInstance {
                 )))
             }
         };
+        let mut new_cache = HashMap::new();
         for bucket in buckets {
             match bucket {
                 Ok(b) => {
-                    self.buckets_cache.insert(b.id.clone(), b.clone());
+                    new_cache.insert(b.id.clone(), b);
                 }
                 Err(e) => {
                     return Err(DatastoreError::InternalError(format!(
@@ -330,6 +331,7 @@ impl DatastoreInstance {
                 }
             }
         }
+        self.buckets_cache = new_cache;
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Windows `sync_roundtrip::test_own_data_does_not_return_via_peer` failed with `database is locked` in the datastore worker (`journal_mode` query). Unrelated to this change — same flake as ActivityWatch/aw-server-rust#655. Ubuntu, macOS, Android, clippy, and format are green; datastore tests including the new merge cases passed.

TimeToBuildBob:
Fixed Greptile's P1 in `edbf0c9`: the move predicate now rejects legacy events that overlap either the destination or another legacy event. Added a regression test that failed before the fix and confirms the overlapping pair remains in the legacy bucket while a disjoint event still moves. The full datastore integration test passes (17/17).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:

PR #672, opened 2026-09-11
ci: deny warnings in clippy (pre-commit and CI)

0xbrayo:
## Summary

- Pass `-D warnings` to clippy in both the pre-commit hook (`.pre-commit-config.yaml`) and the `clippy` job in `lint.yml`, so any rustc or clippy warning fails the check instead of being silently printed.
- Fix the existing warnings that would otherwise trip the new gate immediately. All are mechanical, no behaviour changes:
  - `aw-server/build.rs`: `if let Ok(..)` instead of `is_ok()` + `unwrap()`
  - `aw-transform`: `sort_by_key(Reverse(..))`; collapsed a nested `match`
  - `aw-datastore`: `i64::MAX` instead of `std::i64::MAX`; `#[allow(clippy::result_unit_err)]` on the public `ensure_legacy_import` rather than changing its signature
  - `aw-server/src/dirs.rs`: module-level `allow(clippy::result_unit_err)` for the cfg-gated public dir helpers
  - `aw-sync`: `#[allow(dead_code)]` on `mod util` in the lib (the module is shared with the binary, where the flagged items are used); fixed a needless borrow and `map_or(false, ..)` → `is_some_and`

## Test plan

- [x] `pre-commit run cargo-clippy --all-files` passes
- [x] `cargo fmt --all -- --check` clean
- [x] `cargo test -p aw-transform -p aw-datastore -p aw-sync` passes
- [ ] `clippy` CI job goes green on this PR

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -32,4 +32,4 @@ jobs:
       with:
         components: clippy
     - name: Run clippy
-      run: cargo clippy
+      run: cargo clippy --workspace -- -D warnings
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
 
       - id: cargo-clippy
         name: cargo clippy
-        entry: cargo clippy --workspace
+        entry: cargo clippy --workspace -- -D warnings
         language: system
         types: [rust]
         pass_filenames: false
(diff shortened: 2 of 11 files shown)


0xbrayo:
> 🤖 _Claude, on behalf of @0xbrayo_

@greptile review

Issue #671, opened 2026-09-11
Performance: interval reads, event transforms, query copies, and HTTP exports


0xbrayo:
Track five independent performance improvements identified while reviewing `02ae248`. Each fix has its own PR, independently based on master.

- [ ] Reduce historical scanning for bounded event interval queries — #676. Add a selectively used endtime-first index while preserving overlap semantics, tie ordering, and fast limited newest-first reads. The PR documents index storage/migration costs and measured bulk-insert overhead.
- [ ] Make `union_no_overlap` linear in input size and emitted fragments — #674. Replace repeated middle-of-vector insertions with a pending remainder and move events where possible.
- [ ] Reduce allocations in `merge_events_by_keys` — #673. Remove the unused per-group map, use the entry API, and move the first event into each group instead of cloning its JSON.
- [ ] Avoid deep event-list copies for read-only query functions — #675. Borrow simple variable/literal arguments without changing query language value semantics. Related background: #119, #122, #183.
- [ ] Bound HTTP export memory — #677. Serialize one row at a time into a private temporary file and serve the completed file, preserving the export format, acknowledged writes, and pre-download HTTP errors. Temporary disk use and datastore-worker occupancy are documented tradeoffs.

Initial local probes (synthetic data, not end-to-end application measurements):

- A recent interval containing 1,001 events in a million-row in-memory SQLite database took 44.9 ms with the current index versus 0.41 ms with a forced end-time-first index. This is evidence to evaluate query selection, not a recommendation to replace the existing index unconditionally.
- Overlap merging took 176 ms at 16,000 events per input and 632 ms at 32,000 events per input.
- A grouping prototype reduced 50,000 distinct groups from 22.6 ms to 10.3 ms; with 100 groups the improvement was 11.2 ms to 10.3 ms.

Android context: ActivityWatch/aw-android#229 fixes saving WebView exports and persists queued payloads in cache files. It does not remove full-payload buffering in the Rust HTTP export path; the WebView bridge also still assembles a complete string. This item targets the Rust HTTP path only.

Validation should include representative scaling benchmarks, regression coverage for behavior and edge cases, and independent crate/API checks per PR. Checklist items remain open until their fixes are merged.

Implementation validation uses synthetic in-memory data and isolated temporary test databases only; no live database or running server was used. Reproducible benchmarks/examples are included in the PRs. Additional comparisons against upstream `02ae248`:

[...]

PR #676, opened 2026-09-11
perf(datastore): select an endtime index for recent interval reads

0xbrayo:
Recent interval reads can scan most of a bucket through the starttime-first index even when only a small recent window matches. Add a v6 endtime-first index and select it for unrestricted reads/counts only when the cached bucket time span estimates at least a fourfold reduction in scanned range. Keep the starttime-first index for limited, full-bucket, early, and other less-selective reads.

Explicit timestamp/endtime/ID ordering preserves the existing index's tie order across both plans. Both overlap predicates and clipping remain unchanged; cached metadata influences only access-path selection.

Part of #671 (interval reads).

Validation:
- `cargo test -p aw-datastore`: 32 tests passed, one pre-existing ignored legacy-import test.
- Migration coverage verifies data retention and both composite indexes; regressions compare limited/unrestricted paths for long overlaps, duplicate timestamps, zero-duration events, clipping, counts, and limits.
- File-backed test fixtures now use fresh temporary directories instead of the user's application cache. Benchmarks use synthetic in-memory databases; no live database is accessed.
- The same Criterion benchmark on upstream `02ae248` and this branch, using one million synthetic rows: recent 1,001-row read ≈24.1 ms → 0.40 ms; recent count ≈19.7 ms → 19 µs. Latest-one remained sub-microsecond (≈0.73 µs → 0.78 µs); early and middle ranges retain the ordered starttime scan.

Tradeoffs: the migration adds index construction time and disk space. A separate 10,000-row in-memory insert benchmark measured approximately 3.52 ms with one index versus 4.24 ms with two (about 20% overhead). The temporal selectivity estimate is deliberately conservative and is not a row-count histogram.


0xbrayo:
I didn't realize this required another migration, might need to shelf this one.

0xbrayo:
@greptile review

0xbrayo:
@greptile review

0xbrayo:
@greptile review

PR #680, opened 2026-09-14
fix(android): return an error object from getBuckets instead of panicking

TimeToBuildBob:
Follow-up from #679 / ActivityWatch/aw-android#261.

`Java_…_RustInterface_getBuckets` was the only JNI entry point that still `unwrap()`ed the datastore result. When the worker is gone (it panicked, for example because the database could not be opened), `get_buckets()` returns `SendError`/`RecvError` and the unwrap panicked across the JNI boundary on the calling thread. Seen in the emulator while reproducing #261:

```
log_panics: thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: InternalError("Failed to send request, datastore worker is gone: SendError")': aw-server/src/android/mod.rs:198
   8: Java_net_activitywatch_android_RustInterface_getBuckets
```

Now returns `create_error_object` like the other entry points. The Kotlin callers (`createBucketHelper`, `EventParsingWorker`, `WatcherAndroidBucketMigration.legacyBucketIds`) already tolerate a non-bucket JSON object. Not needed for v0.14.1; it turns a crash in an already-broken state into a logged error.


TimeToBuildBob:
Self-merge not eligible here (cross-repo / policy gate). Waiting for human review or merge.

TimeToBuildBob:

Issue #693, opened 2026-09-16
aw-sync writes to peer databases on every pull: WAL flip + schema migration on files it does not own


ErikBjare:
The core invariant aw-sync's whole conflict-free story rests on — stated in `aw-sync/README.md` as *"each device only writes to files in the sync folder they own, and other devices may not modify them"* — is violated on every pull.

## What happens

`create_datastore` opens a peer database read-write:

```rust
pub fn create_datastore(path: &Path) -> Result<Datastore, String> {
    ...
    Ok(Datastore::new(pathstr.to_string(), false))
}
```

and `Datastore::new` (`aw-datastore/src/worker.rs`) unconditionally:

```rust
let journal_mode: String = conn
    .pragma_update_and_check(None, "journal_mode", "WAL", |row| row.get(0))
    .expect("Failed to query journal_mode");
...
conn.pragma_update(None, "synchronous", "FULL")
...
let mut ds = DatastoreInstance::new(&conn, true).unwrap();
```

So reading a peer's file:

1. **Flips its `journal_mode` to WAL**, creating `-wal` and `-shm` sidecars inside a directory owned by another device.
2. **Runs schema migrations on it** — `DatastoreInstance::new(&conn, true)` upgrades `user_version` toward the local binary's version.
3. Writes `synchronous = FULL`.

All three are writes to a file this device does not own, performed by a *read-only* operation, in a directory an external file syncer is actively replicating.

## Why it matters concretely

- **Version skew makes it a real migration, not a no-op.** Peer databases in a live sync folder sit at different `user_version`s, below current master's. An older peer's file gets silently upgraded by whichever device pulls it first — and that device may be running a newer aw-server-rust than the peer that owns the file. The owner then reopens a database migrated by someone else's binary.
- **It manufactures sync conflicts.** Two devices pulling the same peer, or a pull racing the owner's push, are concurrent writers to one file. My folder already contains the fingerprints: `erb-main3/5a5df0f8-…/test.sync-conflict-20241125-052022-GRUSU5T.db` plus a second from the same day.
- **The `-wal`/`-shm` sidecars are themselves replicated**, and a syncer that delivers a main file and its WAL at different moments produces exactly the torn snapshot the sidecars were meant to prevent.
- It compounds #689's finding that pulls also create empty staging directories inside peers' folders. Between the two, a pull writes to a peer's directory in three different ways.

## Fix

Open peer databases read-only and side-effect-free:

[...]

Issue #688, opened 2026-09-16
aw-sync: per-peer and per-bucket errors abort the whole sync pass (and burn the supervisor's restart budget)


ErikBjare:
*(Rewritten — the original framing claimed aw-sync exits with no supervision to catch it. That was wrong: exiting is the intended failure signal and both bundles implement the supervisor side. The real defect is narrower and is described below. See the comment thread for what changed.)*

Exiting non-zero on failure is a deliberate and reasonable contract — the module signals failure, the supervisor surfaces it — and both bundles implement it properly:

- **aw-qt** (`aw_qt/trayicon.py`): `check_module_status()` polls `manager.get_unexpected_stops()` every 5s, auto-restarts up to `MAX_AUTO_RESTARTS = 3` within `RESTART_WINDOW_SECONDS = 600`, shows a tray notification per restart, and past the budget opens `show_module_failed_dialog()` with the module's log in the detailed text plus a manual Restart button.
- **aw-tauri** (`src-tauri/src/manager.rs`): same idea with exponential backoff — `restart_count < 3`, delay `2^(restart_count+1)` seconds, and an alert window that updates in place when the module recovers.

So for a **whole-run** failure — sync dir missing, config unreadable, aw-server unreachable — the current behaviour is right and should stay.

## The actual defect: no error classification

`aw-sync` promotes *every* error to a whole-process failure, including errors scoped to a single peer or a single bucket.

```rust
loop {
    if let Err(e) = sync::sync_run(client, &sync_spec, mode) {
        error!("Error during sync cycle: {}", e);
        return Err(e);          // any error, whatever its scope
    }
```

and one rung down, `pull_all` propagates per host:

```rust
for host in hostnames {
    pull(&host, client)?
}
```

so a single unreadable peer aborts the pass and every peer after it in iteration order is skipped. #686 preserves this shape (`for remote in selected { pull_db(...)?; }`).

## Why supervision makes this worse, not better

A *persistent* per-peer error becomes a crash loop. On aw-tauri the restart budget is spent in about 14 seconds (2 + 4 + 8), on aw-qt in three polls; then the user gets a modal and **sync stops entirely**. One permanently-bad peer database — a Syncthing conflict file, a db written by a newer schema, a truncated file — therefore takes down sync for *all* peers, permanently, rather than costing you that one peer's data.

[...]

PR #697, opened 2026-09-16
fix(aw-sync): sanitize whitespace hostnames on import; make per-bucket errors non-fatal

### LINKED ISSUE #692: Pulling a peer whose bucket hostname contains whitespace 400s and aborts the whole sync pass (activated by #685)
**Reproduced, and it interacts badly with #685 — worth deciding before v0.14.0 ships.**

#658 made `bucket_new` reject hostnames containing whitespace, with a carve-out for buckets that already exist. On the pull path, `get_or_create_sync_bucket` creates the destination bucket with the **source bucket's hostname**, unchanged. For any peer whose bucket `hostname` column contains a space, that first create is a 400, and the error aborts the entire sync pass.

## Reproduction

Against the shipped `v0.14.0-beta.5 (rust)` binary from the bundle, on a throwaway db:

```console
$ curl -X POST "localhost:5699/api/0/buckets/aw-watcher-android-synced-from-POCO%20F8%20Ultra" \
    -H "Content-Type: application/json" \
    -d '{"id":"aw-watcher-android-synced-from-POCO F8 Ultra","type":"currentwindow",
         "client":"aw-sync","hostname":"POCO F8 Ultra","data":{}}'
{"message":"Invalid hostname \"POCO F8 Ultra\": hostname may not contain whitespace"}
HTTP 400

$ curl -X POST ".../aw-watcher-android-synced-from-poco_f8_ultra" \
    -d '{... "hostname":"poco_f8_ultra" ...}'
HTTP 200
```

The first request is exactly what a pull of my POCO F8 Ultra constructs.

## Why the peer hostname has a space

ActivityWatch/aw-android#272: aw-android added `sanitizeDeviceHostname` and renamed the *sync folder* to `poco_f8_ultra`, but never migrated the `hostname` column on the buckets. Read from the live staging db in my sync folder:

```
aw-watcher-android        | POCO F8 Ultra | aw-android | currentwindow
aw-watcher-android-unlock | POCO F8 Ultra | aw-android | os.lockscreen.unlocks
aw-stopwatch              | POCO F8 Ultra | aw-webui   | general.stopwatch
```

And `get_or_create_sync_bucket` derives the destination from the bucket's hostname, not the folder name. So the sanitized name never reaches the create.

This is not an exotic device name. Any Android phone whose name had a space before the sanitization shipped is affected — "Pixel 8 Pro", "Galaxy S24 Ultra", and so on.

## Blast radius

The failure chain: `create_bucket` → 400 → `map_err(…)?` in `get_or_create_sync_bucket` → aborts `sync_datastores` → aborts `sync_run` → `?` in `pull_all` → **whole pass aborts, every peer after it skipped** → daemon returns `Err` → exits → supervisor restarts → identical failure → budget exhausted → dialog, sync stopped (see #688).

Scope, stated precisely:

- **Only when the destination server is aw-server-rust.** aw-server-python has no whitespace check — I grepped `aw_server/api.py` / `rest.py` and there is none.
- The desktop bundle's default server is aw-server-python, so a **default desktop install is unaffected**.
- **Android is affected** — it embeds aw-server-rust. An Android device pulling a whitespace-hostname peer fails.
- Desktop users who opt into aw-server-rust are affected.

## Why it is latent today and why that changes

Nothing on desktop currently pulls — that is #682. **#685 turns pulling on.** So merging #685 activates this for every aw-server-rust destination that has such a peer in its sync folder. The fix for #682 is what makes this reachable.

## Options

**Naive sanitize-on-import is wrong** — see the correction note below before implementing.

1. **Sanitize on import, reusing the legacy ID.** `get_or_create_sync_bucket` must first look up the *existing unsanitized* destination ID (`…-synced-from-POCO F8 Ultra`) and keep using it if present; only new buckets get a sanitized ID. The raw hostname is preserved in bucket metadata either way. Without the lookup this forks (below).
2. **Exempt sync-created buckets** from the whitespace check. Narrower, but re-admits the values #658 was added to keep out.
3. **Make it per-bucket non-fatal** (#688). Necessary regardless, but on its own it silently drops the peer's data rather than failing loudly — better than dying, still a loss.
4. **Migrate bucket hostnames on Android** (ActivityWatch/aw-android#272). Correct long-term, ships on Android's cadence, does nothing for peers already in folders.

Suggest 1 (with the legacy-ID lookup) + 3 for v0.14.0, with 4 following on Android's own schedule.

### Correction: why plain sanitizing forks

Anyone who pulled this device before #658 landed (Aug 26) already has a local bucket named `…-synced-from-POCO F8 Ultra`. If import starts sanitizing, it constructs `…-synced-from-POCO_F8_Ultra`, `get_or_create_sync_bucket` misses the existing bucket, and resume-from-newest starts from nothing — a **full re-import**, so every event appears twice in `/timeline`. That is the exact symptom of [discussions#1373](https://github.com/orgs/ActivityWatch/discussions/1373) that #648 was filed to fix.

The same fork hits aw-server-python destinations, which never needed sanitizing in the first place and would be broken by a fix aimed at a rust-only problem.

So the destination ID must be resolved as: existing legacy ID if one exists → otherwise sanitized ID. Sanitize the *hostname field and new IDs*, never an established one.

Credit: derived from a design review of aw-sync; I reproduced it against the shipped binary before filing.

Related: #685, #688, #682, ActivityWatch/aw-android#272, #658, ActivityWatch/activitywatch#1445.

cc @TimeToBuildBob

TimeToBuildBob:
Fixes #692.

## Problem

Pulling a peer whose bucket hostname contains whitespace (e.g. `POCO F8 Ultra` from Android devices before ActivityWatch/aw-android#272 sanitized device names) fails with a 400 from `create_bucket` (#658 rejects new buckets with whitespace hostnames). The 400 propagated via `?` and aborted the *entire* sync pass, leaving every subsequent peer unseen (#688 blast radius).

## Fix

**Part 1 — Sanitize on import (Option 1 from #692, with the legacy-ID lookup):**

`get_or_create_sync_bucket` now checks for an existing unsanitized legacy bucket first (so any device already pulled before this fix keeps its established ID and avoids a full re-import fork per ActivityWatch/activitywatch#1373). If no legacy bucket exists, the hostname and derived ID are sanitized with **aw-android's exact algorithm** (`sanitize_hostname()`, byte-identical to `DeviceHostname.kt`: lowercase, `[^a-z0-9_-]+` → `_`, trim `_`) before calling `create_bucket`. `"POCO F8 Ultra"` → `"poco_f8_ultra"`, so Android's hostname-column migration lands on the same ID. `$aw.sync.origin` keeps the raw hostname. The sanitizer also runs when the *hostname field* has whitespace even if the derived ID is already clean.

**Part 2 — Per-bucket and per-peer non-fatal errors (Option 3 from #692 / #688):**

- The `?` propagation in the `sync_datastores` bucket loop is replaced with `warn + continue`. A broken or invalid bucket is skipped; all healthy buckets still sync.
- `pull_all` does the same at peer level: a peer that fails to *open* no longer aborts the pass and skips every peer after it (needed by #693's skip-on-mismatch).

## Tests

```
sanitize_hostname("POCO F8 Ultra") == "poco_f8_ultra"          ← unit, matches DeviceHostname.kt
test_whitespace_hostname_pull_creates_sanitized_bucket
test_whitespace_hostname_pull_reuses_legacy_unsanitized_bucket
test_whitespace_hostname_sanitizes_even_when_id_is_clean       ← origin clean, hostname dirty
test_unusable_datastore_does_not_panic                         ← renamed + updated
test_non_string_sync_origin_does_not_panic                     ← renamed + updated
```

Related: #658, #685, #688, #693, ActivityWatch/aw-android#272, ActivityWatch/activitywatch#1373.


ErikBjare:
@TimeToBuildBob The legacy-ID lookup is right — *existing unsanitized → existing sanitized → create sanitized* is exactly the resolution order #692's correction asked for, `$aw.sync.origin` keeps the raw hostname, and both paths are tested. One blocking item and two small ones.

## Blocking: the sanitizer must match aw-android's, or this defers the fork rather than closing it

This PR sanitizes with `replace(char::is_whitespace, "_")`:

```
"POCO F8 Ultra"  →  "POCO_F8_Ultra"
```

aw-android's `sanitizeDeviceHostname` (`DeviceHostname.kt`) is `lowercase(Locale.ROOT).replace(Regex("[^a-z0-9_-]+"), "_").trim('_')`:

```
"POCO F8 Ultra"  →  "poco_f8_ultra"
```

Trace what happens the day ActivityWatch/aw-android#272 lands and the phone migrates its bucket `hostname` column to `poco_f8_ultra`:

1. `sync_origin` = `"poco_f8_ultra"`, so `new_id` = `…-synced-from-poco_f8_ultra`
2. `get_bucket(new_id)` → `NoSuchBucket` (the existing one is `…-synced-from-POCO_F8_Ultra`)
3. `new_id.contains(char::is_whitespace)` → **false** → else branch → create `…-synced-from-poco_f8_ultra`
4. Resume-from-newest starts from nothing → full re-import → every event twice in `/timeline`

That is the [discussions#1373](https://github.com/orgs/ActivityWatch/discussions/1373) symptom again, just moved to the Android migration date — and by then every desktop that ran this PR has a `POCO_F8_Ultra` bucket that no future ID will match.

Fix is small and only possible *before* merge: use Android's exact algorithm (lowercase, `[^a-z0-9_-]+` → `_`, trim `_`), so today's desktop creates `…-synced-from-poco_f8_ultra` and Android's migration lands on the same ID with no lookup needed. Worth a shared `sanitize_hostname()` in `aw-sync` with a doc comment saying it must stay byte-identical to `DeviceHostname.kt`, plus a test asserting `"POCO F8 Ultra" → "poco_f8_ultra"`.

## Small: per-peer isolation is still missing, and step 3 depends on it

`sync_wrapper.rs`, `pull_all`:

```rust
for remote in selected {
    pull_db(client, &remote.hostname, &remote.path)?;
}
```

This PR makes bucket-level errors non-fatal but a peer that fails to *open* still aborts the pass and skips every peer after it. Plan step 3 (#693 minimal: open peers read-only, skip on `user_version` mismatch) needs this to be `warn + continue` so a skipped peer has somewhere to go — otherwise "skip peer" becomes "abort pass". Three lines; since this is already the non-fatal-errors PR, it belongs here rather than waiting for #688.

## Small: sanitize on the hostname, not only the ID

The branch condition is `new_id.contains(char::is_whitespace)`. If `sync_origin` came from `$aw.sync.origin` metadata (clean) while `bucket_from.hostname` still has whitespace, `new_id` is clean, the else branch runs, `final_hostname` keeps the whitespace, and `create_bucket` 400s anyway. Narrow, but the guard should be `bucket_from.hostname.contains(char::is_whitespace) || new_id.contains(char::is_whitespace)`.

With the first item fixed I have nothing else on this one.

TimeToBuildBob:

TimeToBuildBob:
Addressed all three.

**Blocking — sanitizer now matches aw-android.** Shared `sanitize_hostname()` is byte-identical to `DeviceHostname.kt` (trim, lowercase, `[^a-z0-9_-]+` → `_`, trim `_`). `"POCO F8 Ultra"` → `"poco_f8_ultra"`, not `"POCO_F8_Ultra"`. `$aw.sync.origin` still keeps the raw hostname. Tests lock the Android `DeviceHostnameTest.kt` cases plus the POCO example, and assert the whitespace-only form is *not* created.

**Per-peer isolation.** `pull_all` is now `warn + continue` on `pull_db` failure, so a peer that fails to open does not abort the pass. Step 3 (#693 skip-on-mismatch) has somewhere to go.

**Sanitize on hostname, not only ID.** Guard is `bucket_from.hostname.contains(whitespace) || new_id.contains(whitespace)`. Added a test where `$aw.sync.origin` is already `poco_f8_ultra` while `hostname` is still `POCO F8 Ultra`.

Pushed as 67eeee7. The earlier "waiting on a maintainer click" note is superseded — this needed the review items first. CI is running.

ErikBjare:
@TimeToBuildBob review

TimeToBuildBob:
All three items confirmed in 67eeee7. CI is green on all platforms.

**Blocking — sanitizer matches aw-android:** `sanitize_hostname()` does trim → lowercase → `[^a-z0-9_-]+`→`_` → trim`_`, byte-identical to `DeviceHostname.kt`. `"POCO F8 Ultra"` → `"poco_f8_ultra"` not `"POCO_F8_Ultra"`. `$aw.sync.origin` still keeps the raw hostname. Tests lock all the Android `DeviceHostnameTest.kt` cases plus the POCO example, and assert the whitespace-only form is never created.

**Per-peer isolation:** `pull_all` uses `warn + continue` on `pull_db` failure; a peer that fails to open does not abort the pass. #693 skip-on-mismatch has somewhere to go.

**Sanitize on hostname not only ID:** guard is `bucket_from.hostname.contains(whitespace) || new_id.contains(whitespace)`. `test_whitespace_hostname_sanitizes_even_when_id_is_clean` covers the exact case where `$aw.sync.origin` is already `poco_f8_ultra` while `hostname` is still `POCO F8 Ultra`.

One trivial nit (not blocking): in `test_unusable_datastore_does_not_panic` the first assert `!result.is_err() || result.is_ok()` is a tautology (`!is_err() == is_ok()`), so it never fails. The second assert `result.is_ok()` is the live check. Harmless, but worth knowing so it doesn't confuse a future reader.

Ready to merge.

TimeToBuildBob:
## 🤖 AI code review

Adds hostname sanitization to aw-sync's get_or_create_sync_bucket so buckets with whitespace hostnames are created with a sanitized ID/hostname, and makes per-bucket and per-peer sync failures non-fatal (warn+continue). Exports sanitize_hostname from lib.rs and adds unit/integration tests for the sanitizer and the non-fatal behavior.

Confidence 3/5

<details><summary><b>1 finding</b> · ❌ 1 P1</summary>

**❌ P1 high** — `aw-sync/src/sync.rs:298`

In `get_or_create_sync_bucket`, when the source hostname contains whitespace and the derived ID is clean (e.g., `$aw.sync.origin` is already sanitized), the code sanitizes the hostname and the ID. However, the sanitized ID is computed by calling `sanitize_hostname(origin)` on the `sync_origin` value, which is the raw `$aw.sync.origin` string. If that origin string contains characters that sanitize to an empty string (e.g., `***`), `sanitize_hostname` returns `"unknown"`, so the ID becomes `...-synced-from-unknown`. This is the same sentinel that the `sync_datastores` loop explicitly refuses to sync on pull (the `hostname == "unknown"` check). But here the bucket is created with `-synced-from-unknown` and then synced, mixing events from any device whose origin sanitizes to `unknown` into one bucket. The existing test `test_unknown_hostname_on_pull_returns_error` only covers the case where the hostname itself is `"unknown"`, not where a whitespace-containing hostname sanitizes to `"unknown"`. For example, a hostname `"***"` (which is not whitespace but contains only non-alphanumeric chars) would pass the whitespace check? No, `***` has no whitespace, so the sanitizer is not invoked. The sanitizer is only invoked when `bucket_from.hostname.contains(char::is_whitespace)` or `new_id.contains(char::is_whitespace)`. A hostname like `" * "` (space, star, space) contains whitespace, sanitizes to `"unknown"` (since `*` is replaced with `_`, then trimmed to empty). Then the ID becomes `...-synced-from-unknown`, and the bucket is created. This violates the provenance guarantee that the `unknown` hostname check is meant to enforce, and could mix events from different devices that all sanitize to `unknown`. The fix is to check if the sanitized hostname is `"unknown"` and refuse the sync (or skip) rather than creating a bucket with that ID.

```suggestion
After sanitizing, if sanitized_hostname == "unknown" and is_push is false, return an error or skip the bucket, similar to the existing unknown-hostname check.
```

<sub>**How this was verified:** Traced the logic: the sanitizer returns "unknown" for empty results; the code then uses it as the hostname and in the ID. The existing unknown-hostname guard in sync_datastores only checks the original hostname before sanitization.</sub>

<sub>Consensus: 2/3 passes agreed — pass 1 looked and disagreed</sub> <br/><span style='color:gray;font-size:0.85em'>Distinct keys: 1 (<code>general</code>)</span>

</details>

**⚠️ P2 medium** — `aw-sync/src/sync.rs:216`

The `sanitize_hostname` function is exported as `pub use sync::sanitize_hostname;` in `lib.rs`. This is a new public API. The PR description says it must stay byte-identical to Android's `sanitizeDeviceHostname`. However, the implementation uses `to_lowercase()` which is Unicode-aware, while Android's `toLowerCase()` without a locale may behave differently for certain characters (e.g., Turkish 'I'). The PR claims byte-identical, but without a test against the actual Android implementation, this is unverified. The unit tests only cover ASCII cases. This is a contract risk: if Android's implementation differs for non-ASCII hostnames, the sync IDs will diverge, causing bucket forks. The PR should either restrict to ASCII or add a test with a non-ASCII hostname that matches Android's expected output. This is a test-coverage finding, but it crosses a boundary (the Android contract) and names a specific input (e.g., 'Éclair' or 'İ'). However, the PR's own doc comment defines the algorithm as 'trim, lowercase, replace `[^a-z0-9_-]+` with `_`', which is exactly what the code does. The divergence would only occur if Android's implementation does not use Unicode lowercase. Without knowing Android's code, this is speculative. The PR claims byte-identical, so the onus is on the PR to prove it. But as a reviewer, I cannot confirm a concrete bug without the Android source. I'll not report this as a bug, but as a heuristic.

<sub>**How this was verified:** The doc comment specifies the algorithm. The code matches it. The only risk is if Android's implementation differs, but that is not verifiable from this diff.</sub>

<sub>Consensus: 2/3 passes agreed — pass 1 looked and disagreed</sub> <br/><span style='color:gray;font-size:0.85em'>Distinct keys: 1 (<code>general</code>)</span>

</details>

- `aw-sync/src/lib.rs` — Adds `pub use sync::sanitize_hostname;` to the public exports.
- `aw-sync/src/sync.rs` — Adds `sanitize_hostname()`; rewrites `get_or_create_sync_bucket` to look up legacy unsanitized IDs, sanitize when whitespace is present, and reuse existing sanitized buckets; changes `sync_datastores` bucket loop to warn+continue on per-bucket errors; adds unit tests for the sanitizer.
- `aw-sync/src/sync_wrapper.rs` — Changes `pull_all` to warn+continue when a peer fails to open instead of aborting the pass.
- `aw-sync/tests/sync.rs` — Renames/updates two panic tests to assert non-fatal Ok(()) behavior, and adds three integration tests for whitespace hostname sanitization and legacy bucket reuse.

</details>

<sub>Reviewed `67eeee76ec9c` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 93s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

aw-sync/src/sync.rs
```diff
@@ -236,32 +272,74 @@ fn get_or_create_sync_bucket(
...
+    // `$aw.sync.origin` can already be clean while `bucket_from.hostname` still
+    // has spaces, and `create_bucket` would 400 on the hostname field.
+    //
+    // Sanitization uses Android's algorithm (not a whitespace-only replace) so
+    // today's desktop creates `…-synced-from-poco_f8_ultra` and Android's
+    // hostname-column migration lands on the same ID.
+    let (final_id, final_hostname) = if bucket_from.hostname.contains(char::is_whitespace)
+        || new_id.contains(char::is_whitespace)
+    {
+        let sanitized_hostname = sanitize_hostname(&bucket_from.hostname);
```
TimeToBuildBob: ❌ **P1** — In `get_or_create_sync_bucket`, when the source hostname contains whitespace and the derived ID is clean (e.g., `$aw.sync.origin` is already sanitized), the code sanitizes the hostname and the ID. However, the sanitized ID is computed by calling `sanitize_hostname(origin)` on the `sync_origin` value, which is the raw `$aw.sync.origin` string. If that origin string contains characters that sanitize to an empty string (e.g., `***`), `sanitize_hostname` returns `"unknown"`, so the ID becomes `...-synced-from-unknown`. This is the same sentinel that the `sync_datastores` loop explicitly refuses to sync on pull (the `hostname == "unknown"` check). But here the bucket is created with `-synced-from-unknown` and then synced, mixing events from any device whose origin sanitizes to `unknown` into one bucket. The existing test `test_unknown_hostname_on_pull_returns_error` only covers the case where the hostname itself is `"unknown"`, not where a whitespace-containing hostname sanitizes to `"unknown"`. For example, a hostname `"***"` (which is not whitespace but contains only non-alphanumeric chars) would pass the whitespace check? No, `***` has no whitespace, so the sanitizer is not invoked. The sanitizer is only invoked when `bucket_from.hostname.contains(char::is_whitespace)` or `new_id.contains(char::is_whitespace)`. A hostname like `" * "` (space, star, space) contains whitespace, sanitizes to `"unknown"` (since `*` is replaced with `_`, then trimmed to empty). Then the ID becomes `...-synced-from-unknown`, and the bucket is created. This violates the provenance guarantee that the `unknown` hostname check is meant to enforce, and could mix events from different devices that all sanitize to `unknown`. The fix is to check if the sanitized hostname is `"unknown"` and refuse the sync (or skip) rather than creating a bucket with that ID. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob: Fixed in 7d6cfd9.

After sanitizing, pull now refuses if `sanitized_hostname == "unknown"` or the ID would be `-synced-from-unknown`. Per-bucket warn+continue skips that bucket; a healthy sibling still syncs.

Covered by `test_whitespace_hostname_that_sanitizes_to_unknown_is_skipped_on_pull` (`" * "` plus a healthy `device-0` bucket).

TimeToBuildBob:
P1 from the AI review: a hostname like `" * "` sanitizes to `unknown` and would have created `-synced-from-unknown`, mixing every such remote. That's the same provenance hole the existing `hostname == "unknown"` guard exists to close, but the guard never saw it because the *source* hostname wasn't the sentinel yet.

7d6cfd9 refuses that pull; per-bucket skip keeps the rest of the pass. Test: `test_whitespace_hostname_that_sanitizes_to_unknown_is_skipped_on_pull`.

P2 (Locale.ROOT vs Unicode lowercase) left as-is — ASCII is what `DeviceHostname.kt` tests, and that's the fork this PR has to close.

CI running on 7d6cfd9. Still pull-only here; needs a maintainer click once green.

ErikBjare:
@TimeToBuildBob Now `CONFLICTING` against master — #698 landed and rewrote the `ds_localremote` block in `sync_run` that this touches. Needs a rebase; since it needs a push anyway for the sanitizer (match `DeviceHostname.kt` byte-for-byte) and the per-peer `continue` in `pull_all`, fold all three into one push. Note #678 may go in ahead of this now that it is green and mergeable — rebase over both if so.

ErikBjare:
@greptileai review

TimeToBuildBob:
Rebased onto master (`8037907`) over ActivityWatch/aw-server-rust#698's `maybe_setup_local_remote` rewrite. MERGEABLE. CI in flight.

Folded into this one push as asked:
- sanitizer still byte-identical to `DeviceHostname.kt` (`"POCO F8 Ultra"` → `poco_f8_ultra`)
- `pull_all` per-peer `continue` (warn, do not abort the pass)
- refuse pull when hostname sanitizes to `unknown`

Did not restack onto ActivityWatch/aw-server-rust#678: hunks still do not overlap (`reconcile_updated_events` in `sync_one` vs sanitizer/`pull_all`), and stacking would dump #678's commits into this PR. If #678 merges first this should stay mergeable; I will rebase again only if it does not.

aw-sync/tests/sync.rs
```diff
@@ -122,16 +129,239 @@ mod sync_tests {
             &SyncSpec::default(),
         );
         assert!(
-            result.is_err(),
-            "an unusable destination datastore must return Err, got {result:?}"
+            !result.is_err() || result.is_ok(),
```
TimeToBuildBob: ⚠️ **P2** — The test `test_unusable_datastore_does_not_panic` contains a tautological assertion: `assert!(!result.is_err() || result.is_ok(), ...)`. This is logically equivalent to `assert!(true)` because `result.is_err()` and `result.is_ok()` are mutually exclusive, so `!result.is_err() || result.is_ok()` is always true. The test then separately asserts `result.is_ok()`, which is the meaningful check. The first assertion is vacuous and provides no additional signal; it appears to be a leftover from a previous version of the test. While the second assertion does verify the non-fatal behavior, the tautological assertion is misleading and should be removed to avoid giving the impression that it checks something.

aw-sync/src/sync.rs
```diff
@@ -425,8 +517,18 @@ pub fn sync_datastores(
...
+        let bucket_to = match get_or_create_sync_bucket(&bucket_from, ds_to, is_push) {
+            Ok(b) => b,
+            Err(e) => {
+                // Non-fatal: log and skip this bucket so a bad peer does not
+                // abort the entire pass and leave other buckets un-synced (#692).
+                warn!(" ! Skipping bucket '{}': {}", bucket_from.id, e);
+                continue;
+            }
+        };
+        if let Err(e) = sync_one(ds_from, ds_to, bucket_from, bucket_to, sync_spec) {
```
TimeToBuildBob: ⚠️ **P2** — In `sync_datastores`, the per-bucket error handling now warns and continues, but the `sync_one` call is also wrapped in a warn-and-continue. However, `sync_one` can partially write events before returning an error (e.g. if `insert_events` fails midway through a chunk). The destination bucket may then contain a partial set of events, and the next sync pass will resume from the destination's newest event, potentially skipping events that were not written. This is a pre-existing issue, but the PR's change to make `sync_one` errors non-fatal increases the likelihood of this partial-write state persisting without user visibility. The warning message does not indicate that the destination may be partially updated, so a user may not realize the sync is incomplete. This is a correctness concern for the new non-fatal error handling: a failed `sync_one` should ideally roll back or at least warn that the destination is in an inconsistent state.

ErikBjare:
@TimeToBuildBob Rebase confirmed — on current master, sanitizer/per-peer `continue`/unknown-skip/all four tests intact. Greptile's re-review (maintainer-triggered, so that path is confirmed working) raised three things; two need changes before this merges.

## Required 1 — sanitize always, not only on whitespace (`sync.rs:338`)

Greptile is right and it is the same fork I blocked on earlier, in case-only form. The guard is `hostname.contains(whitespace) || new_id.contains(whitespace)`, but `DeviceHostname.kt` also lowercases and replaces punctuation. So `PIXEL8` (no whitespace) is left as `PIXEL8` today; when ActivityWatch/aw-android#272 migrates that phone's bucket hostname to `pixel8`, `new_id` becomes `…-synced-from-pixel8`, the legacy lookup finds nothing, and the history forks — exactly what the whitespace fix was written to prevent.

Fix: compute `sanitized = sanitize_hostname(raw)` **unconditionally**. Lookup order stays *raw ID → sanitized ID*; create under the sanitized ID whenever it differs from raw. Then Android's migration always lands on an existing bucket regardless of which character class differed. Add `PIXEL8 → pixel8` and a dotted desktop hostname to the test cases.

(Consequence to state in a comment, not fix: desktop peers with dots — `erb-m2.localdomain` — will create `…-synced-from-erb-m2_localdomain` for *new* imports; existing ones are found via the raw lookup. That is a display wart the `(device_id, id)` identity work in ActivityWatch/activitywatch#302 removes; it is not a reason to keep the fork open.)

## Required 2 — total failure must not report success (`sync.rs:534`)

With `warn + continue` on every bucket and every peer, a pass where the destination server is down — every insert fails, every peer "skipped" — now returns `Ok(())`. Before this PR that was an `Err`, the daemon exited, and the supervisor surfaced it. This PR silently turns "everything broke" into "success", which is the #682 failure mode reintroduced via #688's fix.

Fix: track whether anything succeeded. In both `sync_datastores` and `pull_all`: `if attempted > 0 && succeeded == 0 { return Err("all N buckets/peers failed; last error: …") }`. Partial failure stays non-fatal (that is the point of #688); *total* failure is a whole-run condition and must reach the supervisor. #699's `SyncReport` will carry the detail later; this is the minimum so the CLI/JNI callers can tell the difference now.

## Note, not required — `sync.rs:368` origins that collide after normalization

Two distinct raw origins that sanitize to the same string (`device` / `device `) share a destination bucket and the first one's `$aw.sync.origin`. Real, but it is the pre-existing "two devices with the same hostname collide" problem with a slightly wider net; the fix is device_id identity (ActivityWatch/activitywatch#302), not this PR. Cheap mitigation if you want it: when the raw-lookup hits, `warn!` if the found bucket's `$aw.sync.origin` differs from the current raw origin.

With 1 and 2, nothing further from me.

TimeToBuildBob:
@ErikBjare Done on `fcd851c`.

**1. Sanitize always.** Lookup stays raw ID → sanitized ID; create under the sanitized ID when it differs. Tests: `PIXEL8` → `pixel8`, `erb-m2.localdomain` → `erb-m2_localdomain`, plus legacy `PIXEL8` reuse.

**2. Total failure returns Err.** `sync_datastores` and `pull_all` track attempted/succeeded; `if attempted > 0 && succeeded == 0 { return Err(...) }`. Partial skip stays non-fatal (`" * "` + healthy sibling still Ok). One-bucket total-failure tests now expect Err.

Left the origin-collision warn for ActivityWatch/activitywatch#302 as specified.

aw-sync/src/sync.rs
```diff
@@ -236,32 +272,88 @@ fn get_or_create_sync_bucket(
...
+    // with whitespace hostnames (#658).  The ID-only check is not enough —
+    // `$aw.sync.origin` can already be clean while `bucket_from.hostname` still
+    // has spaces, and `create_bucket` would 400 on the hostname field.
+    //
+    // Sanitization uses Android's algorithm (not a whitespace-only replace) so
+    // today's desktop creates `…-synced-from-poco_f8_ultra` and Android's
+    // hostname-column migration lands on the same ID.
+    let (final_id, final_hostname) = if bucket_from.hostname.contains(char::is_whitespace)
+        || new_id.contains(char::is_whitespace)
+    {
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in fcd851c. Sanitization is unconditional (not whitespace-gated). Lookup stays raw ID then sanitized ID; create under the sanitized ID when it differs. Tests cover PIXEL8 → pixel8 and erb-m2.localdomain → erb-m2_localdomain.

aw-sync/src/sync.rs
```diff
@@ -384,8 +476,18 @@ pub fn sync_datastores(
...
+                warn!(" ! Skipping bucket '{}': {}", bucket_from.id, e);
+                continue;
+            }
+        };
+        if let Err(e) = sync_one(ds_from, ds_to, bucket_from, bucket_to, sync_spec) {
+            warn!(" ! Skipping sync for bucket: {}", e);
+        }
     }
 
     Ok(())
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in fcd851c. `sync_datastores` and `pull_all` return Err when attempted > 0 and succeeded == 0. Partial failure stays non-fatal.

aw-sync/src/sync.rs
```diff
@@ -236,32 +272,88 @@ fn get_or_create_sync_bucket(
...
+        {
+            return Err(format!(
+                "Bucket '{}' hostname sanitizes to the unknown sentinel; \
+                 refusing to sync it without provenance",
+                bucket_from.id
+            ));
+        }
+        // If a sanitized bucket already exists, use it.
+        match ds_to.get_bucket(sanitized_id.as_str()) {
+            Ok(bucket) => return Ok(bucket),
```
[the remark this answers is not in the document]
TimeToBuildBob: Deferred, as specified: this is the pre-existing same-hostname collision with a slightly wider net. The fix is device_id identity (ActivityWatch/activitywatch#302), not this PR.

ErikBjare:
Verified on `fcd851c`: `sanitize_hostname` applied unconditionally (case-only differences now land on the same bucket Android will migrate to), and total failure returns `Err` in both `sync_datastores` and `pull_all` while partial failure stays non-fatal. Nothing further — ready for Erik after #678.

aw-sync/tests/sync.rs
TimeToBuildBob: Valid — that assertion is tautological and should just be the `is_ok()` check. Deferring the one-line cleanup to the post-#700/#697-rebase pass: Erik verified this exact head (`fcd851c`) as ready, and pushing a new commit now would invalidate that verification and reset the merge sequencing (#700 → #678 → #697). Tracked for the rebase.

aw-sync/src/sync.rs
TimeToBuildBob: Fixed in fcd851c (already on head): the guard sits inside `get_or_create_sync_bucket` itself — it returns Err when `sanitized_hostname == "unknown"` or the derived ID would end in `-synced-from-unknown`, so no bucket is ever created under the shared sentinel (not just refused on pull). Per-bucket warn+continue then skips it while healthy siblings still sync. Tests cover `" * "` and `"***"` → `"unknown"` (sync.rs:832-837).

aw-sync/src/sync.rs
TimeToBuildBob: Accepted tradeoff for this PR: the partial-write-on-`insert_events`-failure window is pre-existing behavior of `sync_one`, not introduced here — this PR only stops it from aborting the whole pass. The visibility concern is fair: the per-bucket warning now fires where the pass used to die, so a partial destination is more likely to persist. A proper fix is chunk-level rollback/transactionality in `sync_one`, which touches the same code #700 is reworking — deferring it to the post-#700 rebase pass rather than widening this PR after Erik verified the head. The warning text can be sharpened in the same pass.

TimeToBuildBob:
Acknowledged. Head is still `fcd851c` (the SHA you verified). CI green, MERGEABLE against current master. Pull-only on this org, so I cannot self-merge.

Order remains ActivityWatch/aw-server-rust#700 → ActivityWatch/aw-server-rust#678 → this.

Post-merge rebase preview (not pushed — stacking now would dump #700/#678 into this PR):
- This vs ActivityWatch/aw-server-rust#678: clean
- This vs ActivityWatch/aw-server-rust#700: one conflict in `aw-sync/src/sync.rs` — adjacent insert of #700's `open_peer_datastore`/`utf8_db_path` vs this PR's `sanitize_hostname`. Keep both.
- Stacked #700+#678+this with that resolution: `cargo test -p aw-sync` green (lib 27, historical_edit 9, sync 17, roundtrip 3).

Once those two land I will rebase, fold the deferred tautological-assertion cleanup, and this is your click.

Resolved the three Greptile threads (two fixed in `fcd851c`; origin-collision deferred to ActivityWatch/activitywatch#302 as specified).

aw-sync/src/sync.rs
```diff
@@ -277,32 +313,86 @@ fn get_or_create_sync_bucket(
...
+        // Push path: keep the original bucket ID; only the hostname field
+        // needs to be a legal create_bucket value.
+        new_id.clone()
+    };
+    // Android maps empty/punctuation-only names to the "unknown" sentinel.
+    // Creating `-synced-from-unknown` on pull would mix every such remote
+    // into one destination — the same provenance hole the
+    // `hostname == "unknown"` guard in `sync_datastores` exists to close.
+    // Refuse; the per-bucket warn+continue then skips this bucket.
+    if !is_push
```
TimeToBuildBob, line 456: ❌ **P1** — The `sanitize_hostname` function uses `to_lowercase()` which is Unicode-aware, but the character check `c.is_ascii_lowercase()` only accepts ASCII lowercase. For non-ASCII characters that lowercase to multi-char or non-ASCII (e.g., 'İ' lowercases to 'i̇' two chars), the loop will treat the original char as not allowed and replace with '_', but the lowercased string may contain multi-byte chars. More importantly, the function claims to be byte-identical to aw-android's `sanitizeDeviceHostname`, but Android's implementation likely uses `toLowerCase()` with default locale, which can differ for Turkish etc. However, the concrete bug: for a hostname like 'Émile', `to_lowercase()` gives 'émile', but 'é' is not ASCII lowercase, so it becomes '_mile' — but Android's regex `[^a-z0-9_-]` would also replace 'é' with '_', so that matches. The real issue is that `to_llowercase()` can expand a single char into multiple chars (e.g., 'İ' -> 'i̇'), and the loop iterates over the lowercased chars, so 'i' is allowed, then the combining dot is not allowed and becomes '_', producing 'i_' instead of Android's likely 'i' or 'i̇'? Actually Android's `toLowerCase()` on 'İ' with default locale (Turkish) gives 'i' (dotless? no, 'i' with dot? Actually 'İ' lowercases to 'i' in Turkish, but in default locale it might be 'i̇' with combining dot). This is an edge case unlikely in hostnames. More concrete: the function trims leading/trailing '_' after replacing, but Android's algorithm might not trim underscores? The doc says trim '_', so that's fine. The real bug: `sanitize_hostname` is used to sanitize the `origin` in `sanitized_id`, but the `origin` is the raw `$aw.sync.origin` or hostname. If the origin contains characters that sanitize to empty (e.g., '***'), `sanitize_hostname` returns 'unknown', and the code refuses to sync on pull. But on push, it does not refuse, and creates a bucket with hostname 'unknown'? Actually on push, `sanitized_id` is `new_id` (unchanged), but `final_hostname` is set to `sanitize_hostname(bucket_from.hostname)`. If the hostname sanitizes to 'unknown', the push path will create a bucket with hostname 'unknown' and the original ID. That might be acceptable, but the comment says 'unknown' sentinel is a provenance hole. However, the push path already substitutes device ID for 'unknown' hostname in `sync_datastores` (line 484-499), so the hostname would be replaced before reaching here. So that's fine. The most concrete issue: the `sanitize_hostname` function is not actually byte-identical to Android's because Android's `sanitizeDeviceHostname` likely uses `replace(Regex("[^a-z0-9_-]+"), "_")` after lowercasing, which collapses multiple consecutive non-allowed chars into a single '_'. This implementation does that with `in_run`. But it also trims leading/trailing '_' — does Android do that? The doc says it does, but if Android doesn't trim, then '***' would become '_' not 'unknown'? The test expects '***' -> 'unknown', so the implementation trims and then sees empty. If Android doesn't trim, then '***' would be '_' and not 'unknown', causing a mismatch. But the PR claims byte-identical, and the tests are written to match the claimed Android behavior. Without seeing Android's code, I can't confirm. However, there is a clear bug in the interaction between `sanitize_hostname` and the `sanitized_id` construction: when `sync_origin` is `Some(origin)`, the `sanitized_id` is built using `sanitize_hostname(origin)`, but the `origin` is the raw `$aw.sync.origin` value, which may be a hostname that was already sanitized by Android? Actually the test `test_whitespace_hostname_sanitizes_even_when_id_is_clean` has origin 'poco_f8_ultra' and hostname 'POCO F8 Ultra'. The sanitized_id becomes 'aw-watcher-android-synced-from-poco_f8_ultra' (since origin is already clean). The final_hostname is sanitize_hostname('POCO F8 Ultra') = 'poco_f8_ultra'. That works. But consider a case where the origin is 'POCO F8 Ultra' (raw) and the hostname is also 'POCO F8 Ultra'. The sanitized_id uses sanitize_hostname(origin) = 'poco_f8_ultra', and final_hostname = sanitize_hostname(hostname) = 'poco_f8_ultra'. Good. But what if the origin is 'PIXEL8' and hostname is 'PIXEL8'? sanitized_id = '...-synced-from-pixel8', final_hostname = 'pixel8'. Good. The bug I see: the code checks `sanitized_id.ends_with("-synced-from-unknown")` to refuse pull. But `sanitized_id` is built from `sanitize_hostname(origin)`, and if origin is 'unknown' (literally), sanitize_hostname('unknown') returns 'unknown', so it refuses. But if origin is '***', sanitize_hostname returns 'unknown', so it refuses. That's intended. However, the check `sanitized_hostname == "unknown"` is also there. If the hostname is 'unknown' but the origin is something else (e.g., origin is 'device-0'), then sanitized_hostname == 'unknown' triggers refusal even though the origin is known. That could be a bug: a bucket with hostname 'unknown' but a valid `$aw.sync.origin` should be syncable because the origin provides provenance. The existing code in `sync_datastores` (line 484) substitutes the device ID for 'unknown' hostname on push, but on pull it refuses. With the new check, even if `$aw.sync.origin` is set to a valid device, the pull refuses because `sanitized_hostname == "unknown"`. This is a regression: previously, a bucket with hostname 'unknown' but with `$aw.sync.origin` set would be synced (the origin is used for the ID, and the hostname field would be 'unknown' but create_bucket might reject? Actually the old code didn't sanitize, so it would try to create a bucket with hostname 'unknown' — that's legal? The server might allow 'unknown' as a hostname? The comment says the `hostname == "unknown"` guard in `sync_datastores` exists to close the provenance hole, but that guard only triggers when there's no source device ID. On pull, `src_did` is None, so it refuses. But if `$aw.sync.origin` is set, the origin is used for the ID, and the hostname field is still 'unknown'. The old code would create a bucket with hostname 'unknown' and ID '...-synced-from-<origin>'. That might be acceptable because the origin provides provenance. The new code refuses because `sanitized_hostname == "unknown"` regardless of origin. This is a concrete behavior change that could break syncing for buckets that have a valid origin but a hostname of 'unknown' (e.g., if the source device's hostname was 'unknown' but it set `$aw.sync.origin` to its device ID). The test `test_unknown_hostname_on_pull_returns_error` uses a bucket with hostname 'unknown' and no origin, so it still passes. But a bucket with hostname 'unknown' and origin 'device-0' would now be refused, whereas before it would sync. Is that a real scenario? Possibly if a device's hostname is literally 'unknown' but it has a device ID. The PR's intent is to avoid mixing events from unknown devices, but if the origin is known, it's not mixing. The check should be based on the sanitized origin, not the hostname. The code already checks `sanitized_id.ends_with("-synced-from-unknown")`, which covers the case where the origin sanitizes to 'unknown'. The additional `sanitized_hostname == "unknown"` check is redundant and overly broad. This is a bug: it refuses to sync a bucket that has a valid origin but a hostname that is the literal string 'unknown'. The consequence is that such buckets are skipped on pull, and if all buckets are skipped, the sync returns Err. This could cause a sync failure for a legitimate device. I'll report this as a P1 bug.

aw-sync/tests/sync.rs
```diff
@@ -114,24 +118,371 @@ mod sync_tests {
...
+            Some("POCO F8 Ultra")
+        );
+    }
+
+    /// If a legacy unsanitized bucket already exists in the destination (imported
+    /// before the sanitization was added), re-use it instead of creating a new
+    /// sanitized one.  Creating a sanitized copy forks the destination and causes
+    /// a full re-import (ActivityWatch/activitywatch#1373).
+    #[test]
+    fn test_whitespace_hostname_pull_reuses_legacy_unsanitized_bucket() {
```
TimeToBuildBob, line 216: ⚠️ **P2** — The test `test_whitespace_hostname_pull_reuses_legacy_unsanitized_bucket` creates a legacy destination bucket with ID `aw-watcher-android-synced-from-POCO F8 Ultra` and hostname `POCO F8 Ultra`. The source bucket has hostname `POCO F8 Ultra` and no `$aw.sync.origin`. In `get_or_create_sync_bucket`, the raw lookup finds the legacy bucket and returns it immediately (line 322). The test then asserts that the legacy bucket is preserved and no sanitized fork is created. This passes. However, the test does not verify that the events are actually synced into the legacy bucket. It only checks bucket existence. The `sync_one` will then sync events into the legacy bucket, but the test doesn't assert that. That's a coverage gap, not a defect in the test. The test is not vacuous because it asserts the bucket is reused. But there is a subtle issue: the legacy bucket has hostname `POCO F8 Ultra` (with space). When `sync_one` runs, it uses `bucket_to.id` which is the legacy ID, and inserts events. The destination bucket's hostname remains unsanitized. That's intended for legacy reuse. The test doesn't check that the hostname is preserved, but that's fine. I don't see a test defect here.

TimeToBuildBob: Rejected on `fcd851c`. Coverage-gap nit, not a test defect.

`test_whitespace_hostname_pull_reuses_legacy_unsanitized_bucket` asserts the invariant it exists for: raw-ID lookup returns the pre-existing `…-synced-from-POCO F8 Ultra` bucket and does not create `…-synced-from-poco_f8_ultra`. The finding itself concludes "I don't see a test defect here."

Event insertion is covered by `test_events` / `test_sync_resume_after_partial_sync` / the create-sanitized sibling test. Not folding an events assertion into the reuse test, and not pushing a test-only commit onto a SHA Erik already verified.

aw-sync/src/sync.rs
TimeToBuildBob: Rejected on `fcd851c`.

The writeup lands on two claims. Both fail against the code and against Android.

**1. Unicode / Locale / trim — not a mismatch.**
aw-android's `sanitizeDeviceHostname` is `lowercase(Locale.ROOT)` then `replace(Regex("[^a-z0-9_-]+"), "_")` then `trim('_')` then empty → `"unknown"` (`DeviceHostname.kt`). It is not default-locale. Rust `to_lowercase()` + `is_ascii_*` + `trim_matches('_')` is the same algorithm. `'İ'` → `i` + combining dot → `i_` → trim → `i` on both sides. `'Émile'` → `_mile` on both sides. The Android tests this PR locked (`""`, `"   "`, `"***"`, `" * "`) are the DeviceHostname.kt cases.

**2. `sanitized_hostname == "unknown"` is not a regression, and it is not the mixing hole.**
- Literal hostname `"unknown"` never reaches this check on pull: `sync_datastores` (the `hostname == "unknown"` arm, `src_did` is `None`) already returns Err, origin or not. Pre-existing. `test_unknown_hostname_on_pull_returns_error` is that path.
- Mixing is the destination ID `-synced-from-unknown`. That is independently gated by `sanitized_id.ends_with("-synced-from-unknown")`.
- The hostname clause is what 7d6cfd9/8037907 added for sanitizer-equivalent junk (`" * "`, `"***"`) that bypasses the exact-string guard. That is `test_whitespace_hostname_that_sanitizes_to_unknown_is_skipped_on_pull`. Prior finding `92b7187192f4` demanded this check; reversing it is oscillation.
- Allowing the pull through would still write destination hostname `"unknown"`. Android's own comment on the sentinel: `/#/activity/unknown/` is a terrible default. Origin-in-metadata does not fix that.

The "valid `$aw.sync.origin` + junk hostname" exception is a policy change, not a bugfix. Out of scope for this PR.

TimeToBuildBob:
## 🤖 AI code review

This PR adds hostname sanitization to aw-sync's bucket import path, making per-bucket and per-peer failures non-fatal, and updates tests accordingly. It introduces a `sanitize_hostname` function that mirrors aw-android's algorithm, uses it to derive sanitized bucket IDs and hostnames when creating destination buckets, and changes `sync_datastores` and `pull_all` to warn-and-continue on individual failures while still returning an error if all attempts fail.

### Safe to merge — no P0/P1 findings

Confidence 5/5

✅ **No thread-worthy findings.** Advisory notes follow; they are retained without opening review threads.

**⚠️ P2 medium** — `aw-sync/tests/sync.rs:430`

The test `test_whitespace_hostname_that_sanitizes_to_unknown_is_skipped_on_pull` creates a bucket with hostname `" * "` and expects it to be skipped. The sanitizer returns `"unknown"` for `" * "` because after trimming it's `"*"`, which is not alphanumeric, so it becomes `_`, then trimmed to empty, then `"unknown"`. The test asserts that no bucket with `-synced-from-unknown` is created and that a healthy sibling syncs. This test is correct. However, the test `test_unknown_hostname_on_pull_returns_error` (existing) uses hostname `"unknown"` and expects an error. With the new code, `sanitize_hostname("unknown")` returns `"unknown"`, and the guard `sanitized_hostname == "unknown"` triggers, returning an error. That test still passes. No issue.

<sub>**How this was verified:** Read the test and the sanitizer logic.</sub>

<sub>Consensus: 2/3 passes agreed — pass 2 looked and disagreed</sub> <br/><span style='color:gray;font-size:0.85em'>Distinct keys: 1 (<code>general</code>)</span>

</details>

- `aw-sync/src/lib.rs` — Exports the new `sanitize_hostname` function from the sync module.
- `aw-sync/src/sync.rs` — Adds `sanitize_hostname`, rewrites `get_or_create_sync_bucket` to sanitize IDs/hostnames with legacy lookup, and makes per-bucket failures non-fatal in `sync_datastores`.
- `aw-sync/src/sync_wrapper.rs` — Makes per-peer failures non-fatal in `pull_all`, returning an error only if all peers fail.
- `aw-sync/tests/sync.rs` — Adds tests for sanitized bucket creation, legacy bucket reuse, unknown sentinel skipping, and updates existing tests for non-fatal behavior.
- `aw-sync/tests/sync_roundtrip.rs` — Updates expected bucket ID in roundtrip test to use sanitized lowercase hostname.

</details>

<sub>Reviewed `8efcf9cf4dd9` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 104s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

ErikBjare:
@greptileai review

TimeToBuildBob:
Rebased onto master over ActivityWatch/aw-server-rust#700 (`d2ae2e6`). MERGEABLE. CI in flight on `8efcf9c`.

Conflict was the previewed adjacent insert in `aw-sync/src/sync.rs`: kept both #700's `open_peer_datastore`/`utf8_db_path` and this PR's `sanitize_hostname`.

Folded the deferred one-line: a skipped `sync_one` now warns that dest may already contain a partial write. The tautological assert in `test_unusable_datastore_does_not_panic` was already gone in `fcd851c` (rewritten to `expect_err` when total-failure became Err).

Local: `cargo test -p aw-sync` green (lib 27, sync 17, roundtrip 3).

ActivityWatch/aw-server-rust#678 still open and CLEAN; pairwise this vs #678 was clean before, so it should stay mergeable if that lands first. Still pull-only — needs a maintainer click after CI.

Did not re-trigger Greptile (your 14:09 request was on `fcd851c`; this head is new).

ErikBjare:
@TimeToBuildBob Resolve conflicts

ErikBjare:

TimeToBuildBob:
Rebased onto `3d3b9dd` (#678 + #699 now in base). MERGEABLE. CI in flight on `bff7a2a` (`8efcf9c` → `bff7a2a`; the diff vs. previous head is only the rebase — the five PR commits are unchanged in content, just replayed).

Ordering noted: #697 → #703 → #702. Still pull-only — ready for a maintainer click once CI is green.

ErikBjare:
Re-verified on `bff7a2a` after the rebase over #699 + #678: on current master, unconditional `sanitize_hostname` (`:438`), `attempted > 0 && succeeded == 0 → Err` in both `sync_datastores` (`:653`) and `pull_all` (`:81`), unknown-after-sanitize skip, all four tests present — and `sync_run` returns `Result<SyncReport, …>`, so it is integrated with #699's report rather than just rebased over it. Nothing further; ready for Erik on green. #703 next.

TimeToBuildBob:
Closing the loop on this dispatch — the rebase ask was delivered on `bff7a2a` and this merged at 09:15:50Z (merge commit `d1a61b12`, merged by @ErikBjare).

Post-merge verification (not previously on this thread):
- CI on the merge commit `d1a61b12`: 7 success, 4 path-skipped (no failures).
- `ActivityWatch/aw-server-rust#703` landed after it — master is now `b0fab735`, CI green except Code coverage still running.
- All review threads (Greptile + in-band AI review) were resolved or dismissed with reasons before merge; nothing outstanding.

No further action from Bob on this PR: the sanitizer / per-peer-`continue` work is on master, and per your note on ActivityWatch/activitywatch#1445 the `aw-tauri` pin bump + cut are yours.

Issue #691, opened 2026-09-16
aw-sync publishes aw-server-rust's private sqlite schema as the wire format (unreadable by aw-server-python, unversioned, WAL-mutable)


ErikBjare:
The files aw-sync publishes into the shared folder are sqlite databases in **aw-server-rust's internal datastore schema**. That schema is undocumented, unversioned as an interchange format, migration-mutable, and — the part that surprised me — not readable by ActivityWatch's own Python server.

## The two schemas are structurally incompatible

`aw-server-rust/aw-datastore` (what lands in the sync folder, read from a real staging db):

```sql
CREATE TABLE buckets (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    name TEXT UNIQUE NOT NULL, type TEXT NOT NULL, client TEXT NOT NULL,
    hostname TEXT NOT NULL, created TEXT NOT NULL
, data_deprecated TEXT DEFAULT '{}', data TEXT NOT NULL DEFAULT '{}');

CREATE TABLE events (
    id INTEGER PRIMARY KEY AUTOINCREMENT, bucketrow INTEGER NOT NULL,
    starttime INTEGER NOT NULL, endtime INTEGER NOT NULL,
    data TEXT NOT NULL,
    FOREIGN KEY (bucketrow) REFERENCES buckets(id));
```

`aw-core/aw_datastore/storages/sqlite.py`:

```sql
CREATE TABLE IF NOT EXISTS buckets (
    rowid INTEGER PRIMARY KEY AUTOINCREMENT,
    id TEXT UNIQUE NOT NULL, name TEXT, type TEXT NOT NULL, client TEXT NOT NULL,
    hostname TEXT NOT NULL, created TEXT NOT NULL, datastr TEXT NOT NULL);

CREATE TABLE IF NOT EXISTS events (
    id INTEGER PRIMARY KEY AUTOINCREMENT, bucketrow INTEGER NOT NULL,
    starttime INTEGER NOT NULL, endtime INTEGER NOT NULL,
    datastr TEXT NOT NULL,
    FOREIGN KEY (bucketrow) REFERENCES buckets(rowid));
```

Differences that matter:

- The bucket's **string key** is `buckets.name` in Rust and `buckets.id` in Python. `name` exists in both and means *opposite* things — the key in one, a nullable display name in the other.
- The bucket's **integer PK** is `id` in Rust, `rowid` in Python, so `events.bucketrow` points at a differently-named column.
[...]

PR #699, opened 2026-09-16
feat(aw-sync): return and persist a SyncReport from each pass

### LINKED ISSUE #695: aw-sync: return and persist a SyncReport — a pass is a side effect, so nothing can report what it did
A sync pass is modelled as a side effect rather than as a function that returns what it did. That single choice is why sync failures are structurally invisible, and it is cheap to reverse.

## Today

- `sync_run(...) -> Result<(), Box<dyn Error>>` — success carries no information.
- The JNI wrapper reports a **fixed string** regardless of what happened:

  ```rust
  pull_all(&client).map_err(|e| format!("Sync pull failed: {}", e))?;
  Ok(json!({
      "success": true,
      "message": "Successfully pulled from all hosts"
  }).to_string())
  ```

  Ten peers or zero peers, a million events or none — identical output.
- Android persists exactly `SyncStatus(completedAt: Long, success: Boolean)` (ActivityWatch/aw-android#274).
- The daemon has no status at all; its only signal is exiting (#688).

The information mostly **already exists and is discarded**. `sync_one` computes `new_events_count` per bucket and logs it; `sync_run` knows which peers it found. None of it survives the call.

This is the mechanism behind #682: a daemon ran 102 consecutive passes importing nothing, logging `Pulling...` each time. Not that nobody looked — there was nothing to look at.

## Proposal

Return a `SyncReport` from `sync_run`, aggregate it in `pull_all`/`push`, and persist the latest one:

```rust
pub struct SyncReport {
    pub started: DateTime<Utc>,
    pub finished: DateTime<Utc>,
    pub mode: SyncMode,
    pub peers: Vec<PeerReport>,   // one per peer considered
    pub pushed: Vec<BucketReport>,
}

pub struct PeerReport {
    pub device_id: String,
    pub hostname: String,
    pub path: PathBuf,
    pub outcome: PeerOutcome,     // Imported | Skipped(reason) | Failed(error)
    pub buckets: Vec<BucketReport>,
}

pub struct BucketReport {
    pub bucket_id: String,
    pub events_new: i64,
    pub resumed_at: Option<DateTime<Utc>>,
}
```

Then:

- `aw-sync status` (#687) reports the last pass as well as current folder state.
- The JNI layer returns real counts, so ActivityWatch/aw-android#274 becomes a rendering job rather than a redesign.
- The daemon can persist last-pass status to a file, so aw-qt/aw-tauri can show "sync is running but two peers are failing" **without the module having to die to be heard** (#688).
- `PeerOutcome::Skipped` gives the duplicate-device_id dedupe from #686 somewhere to report itself, instead of a log line nobody reads.

## Scope boundary

This covers facts about *what a pass did*. It deliberately does **not** cover facts the sync folder cannot currently express — "which device is this peer", "when did it last push", "is a peer missing that used to be here". Those need per-device metadata in the folder (the manifest discussed in ActivityWatch/activitywatch#302 / #691) and are out of scope here.

Worth doing first precisely because it is small, has no format implications, and unblocks three other issues.

Credit: found in a design review of aw-sync.

Related: #684, #687, #688, #682, ActivityWatch/aw-android#274.

cc @TimeToBuildBob

TimeToBuildBob:
## Summary

A sync pass used to return `()`. JNI always said `"Successfully pulled from all hosts"`, the daemon had no status except dying, and `aw-sync status` could only describe the folder, not the last pass. That is why a daemon could run 102 empty cycles with nothing to look at (ActivityWatch/aw-server-rust#682 / #684).

This returns a `SyncReport` from `sync_run` / `pull_all` / `push`, persists the latest one locally (not in the Syncthing folder), and surfaces it in three places:

- `aw-sync status` prints the last pass (peers imported/skipped/failed, event counts)
- JNI returns real counts in the existing `{success, message}` object, plus `events_pulled`, `peers_skipped`, etc. — ActivityWatch/aw-android#274 becomes a rendering job
- The daemon writes the same file, so a later `status` can show what the last pass did even if the process has since exited

`PeerOutcome::Skipped` is how the duplicate-`device_id` dedupe from ActivityWatch/aw-server-rust#686 reports itself, instead of a log line.

Closes ActivityWatch/aw-server-rust#695.

## Persistence

`{data_dir}/aw-sync/last-sync-report.json` (override with `AW_SYNC_LAST_REPORT`). Atomic write. A persist failure is a warning, not a failed pass.

Out of scope, as specified: per-device folder metadata ("which device is this peer", "when did it last push") — that is the manifest in ActivityWatch/activitywatch#302 / ActivityWatch/aw-server-rust#691.

Does not change error policy: a per-peer failure still aborts the pass (ActivityWatch/aw-server-rust#688). The partial report is persisted on that path so the failure is visible afterwards.

## Test plan

- [x] `cargo test -p aw-sync --lib --bins --tests` (40 passed)
- [x] `cargo clippy --workspace -- -D warnings`
- [ ] CI


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `aw-sync/src/report.rs` (+429/-0, Δ429)
- `aw-sync/src/util.rs` (+97/-12, Δ109)
- `aw-sync/src/sync.rs` (+76/-28, Δ104)
- `aw-sync/src/sync_wrapper.rs` (+41/-19, Δ60)
- `aw-sync/src/android.rs` (+18/-28, Δ46)
- `aw-sync/src/main.rs` (+26/-7, Δ33)
- `aw-sync/tests/sync.rs` (+33/-0, Δ33)
- `aw-sync/src/status.rs` (+15/-0, Δ15)

### Suggested verification

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for ActivityWatch/aw-server-rust#699._

ErikBjare:
@TimeToBuildBob Structure matches #695 point for point — `PeerReport.device_id`, `PeerOutcome::Skipped { reason }` giving the #686 dedupe somewhere to report, `BucketReport`, persisted to `{data_dir}/aw-sync/last-sync-report.json` (outside the synced folder, correctly), JNI fixed string gone. Test-merged against #697, #678 and #698: all clean, so the return-type change does not force rebases on the v0.14.x queue.

**Sequencing:** this goes **after** the v0.14.x fixes (#698 → #697 → #678 → #693-minimal), not gating the cut. It is additive and larger (+743), so it gets a full read when it reaches the front rather than a structural pass now. Please keep it rebased as those land.

One ask for when I read it properly: a test that a pass which finds zero peers produces a report saying so (`peers: []` with the discovery warnings captured) — that is the #682 failure mode, and the whole point of #695 is that it can no longer be silent.

TimeToBuildBob:
Added in 42946f6.

A zero-peer pull now persists `peers: []` plus the discovery warnings (`Found 0 remote db files` / empty-dir diagnosis) on `SyncReport`. Test: `zero_peer_pass_reports_empty_peers_and_captures_discovery_warnings`. Those lines were `warn!()`-only before, which is how ActivityWatch/aw-server-rust#682 stayed silent after 102 empty cycles.

Will keep this rebased as ActivityWatch/aw-server-rust#698 → ActivityWatch/aw-server-rust#697 → ActivityWatch/aw-server-rust#678 → ActivityWatch/aw-server-rust#693-minimal land. Not gating the cut.

Same commit also persists `pull_all`'s aggregate when a later peer fails, and writes the report through a unique temp file.

aw-sync/src/sync_wrapper.rs
```diff
@@ -2,32 +2,57 @@ use std::error::Error;
...
+            Err(e) => {
+                report.peers.push(PeerReport::failed(
+                    remote.device_id,
+                    remote.hostname,
+                    remote.path,
+                    e.to_string(),
+                ));
+                report.finish();
+                return Err(e);
+            }
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 42946f6 — `pull_all` now persists the aggregate (earlier peers + the failing one) before returning the error, so `last-sync-report.json` is no longer overwritten by the phase-local report from the failing `sync_run`.

aw-sync/src/report.rs
```diff
@@ -0,0 +1,429 @@
...
+    Ok(path)
+}
+
+pub fn persist_last_report_to(report: &SyncReport, path: &Path) -> Result<(), Box<dyn Error>> {
+    if let Some(parent) = path.parent() {
+        fs::create_dir_all(parent)?;
+    }
+    let tmp = path.with_extension("json.tmp");
+    fs::write(&tmp, serde_json::to_string_pretty(report)?)?;
+    fs::rename(&tmp, path)?;
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 42946f6 — the atomic write now uses `{name}.tmp.{pid}-{nanos}` so overlapping daemon/manual/Android writers cannot truncate the same `.json.tmp`.

ErikBjare:
@TimeToBuildBob `CONFLICTING` against master after #698 (same `sync_run` region). Needs a rebase, but no rush — this is queued after the v0.14.0 cut, so wait until #700/#678/#697 have landed and rebase once over all of them rather than three times.

TimeToBuildBob:
Understood — will rebase once, after #700/#678/#697 all land, not per-PR.

TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 5 findings disposed (other; rejected; fixed in 91255d6)

| --- | --- | --- |
| `aw-sync/src/main.rs:285` | P1 | superseded by latest review (not reproduced) |
| `aw-sync/src/report.rs:357` | P1 | rejected — Reviewer's verification assumed Win32 MoveFile semantics (fail if destination exists); Rust std fs::rename is not raw Mo |
| `aw-sync/src/sync.rs:159` | P1 | other |
| `aw-sync/src/sync.rs:100` | P1 | rejected — Reviewer verified: find_remotes is 2-level, list_remote_dbs is 3-level, so sync_run against a 3-level folder finds 0 rem |
| `aw-sync/src/sync.rs:177` | P1 | fixed in `91255d6` |
| `aw-sync/src/sync_wrapper.rs:48` | P1 | rejected — Reviewer's verification assumed pull_all merges sync_run's in-memory report on Err, producing two PeerReport::failed ent |
| `aw-sync/src/main.rs:296` | P2 | fixed in `49f0ce1` |
| `aw-sync/src/main.rs:365` | P2 | wontfix |
| `aw-sync/src/sync.rs:177` | P2 | superseded by latest review (not reproduced) |
| `aw-sync/src/sync.rs:65` | P2 | accepted-tradeoff |
| `aw-sync/src/sync.rs:122` | P2 | accepted-tradeoff |
| `aw-sync/src/sync.rs:65` | P2 | accepted-tradeoff |
| `aw-sync/src/sync.rs:135` | P2 | accepted-tradeoff |
| `aw-sync/src/sync.rs:133` | P2 | accepted-tradeoff |
| `aw-sync/src/sync_wrapper.rs:46` | P2 | rejected |

</details>




This PR introduces a SyncReport type that is returned from sync_run, pull_all, push, and the JNI entry points, persisted to {data_dir}/aw-sync/last-sync-report.json, and printed by aw-sync status. It adds PeerOutcome::Skipped/Failed reporting for duplicate device_id and incompatible database versions, changes pull_discovery_warnings to accept Option&lt;&str&gt; for the local device id, and adds a close_opened_datastores helper for error paths. The diff also updates Cargo.toml to enable serde derive, adds tests for report round-tripping and event counts, and updates README/status output.

### Needs a look — P2 only

Confidence 4/5

<details><summary><b>1 finding</b> · ⚠️ 1 P2 — non-blocking (round cap reached)</summary>

**⚠️ P2 medium · ℹ️ non-blocking** — `aw-sync/src/sync.rs:133`

In sync_run, when open_peer_datastore returns Err, the code pushes a PeerReport::failed, persists the report, closes datastores, and returns Err. However, the report is only persisted inside the `if mode == SyncMode::Pull || mode == SyncMode::Both` block. If mode is Push, the error path still closes datastores and returns Err, but the report is not persisted and no PeerReport is added. This means a push-only pass that fails to open a remote db (which is not even used in push mode) will not leave a failure record, contradicting the PR's stated contract that a failed pass is visible afterwards. The observable consequence is that `aw-sync status` shows the previous successful report or 'none' after a push-only failure, making the failure invisible.

```suggestion
Move the persist and PeerReport::failed push outside the mode check, or at least persist the report with a warning in Push mode.
```

<sub>**How this was verified:** Checked the code path: the Err branch at line 133-147 only persists when mode is Pull or Both. In Push mode, ds_remotes is empty (no pull phase), so this branch is only reachable if find_remotes_nonlocal_selection returns a remote that fails to open, which is possible even in push mode. The report is created at line 64 but never persisted on this path for Push mode.</sub>

</details>

<details><summary><b>Files changed (11)</b> — the diff as I read it</summary>

- `aw-sync/Cargo.toml` — Enables the serde derive feature for the aw-sync crate.
- `aw-sync/README.md` — Adds a line to the status command documentation about printing the last persisted pass.
- `aw-sync/src/android.rs` — Changes JNI sync entry points to return and persist SyncReport JSON instead of fixed success messages.
- `aw-sync/src/lib.rs` — Adds the report module and re-exports its public types and functions.
- `aw-sync/src/main.rs` — Adds report persistence and summary logging to daemon and sync command paths, including failure handling.
- `aw-sync/src/report.rs` — New module defining SyncReport, PeerReport, BucketReport, persistence functions, and tests.
- `aw-sync/src/status.rs` — Adds loading and printing of the last persisted SyncReport to status output.
- `aw-sync/src/sync.rs` — Changes sync_run and sync_datastores to return SyncReport/BucketReport, adds error-path datastore closing, and records skipped/failed peers.
- `aw-sync/src/sync_wrapper.rs` — Changes pull_all, pull, push, and push_with_hostname to return SyncReport and persist partial reports on failure.
- `aw-sync/src/util.rs` — Adds select_remote_dbs_detailed and find_remotes_nonlocal_selection returning skipped remotes, and changes pull_discovery_warnings to take Option&lt;&str&gt;.
- `aw-sync/tests/sync.rs` — Adds a test asserting sync_datastores returns new event counts and that a second pull is a no-op.

</details>

| `42946f6494d4` | 2/5 | 3 | llm | 2026-09-16 09:28 UTC |
| `49f0ce1726a2` | 2/5 | 3 | llm | 2026-09-16 12:44 UTC |
| `91255d66a998` | 4/5 | 3 | llm | 2026-09-16 13:22 UTC |
| `592c7b3ac0cb` | 4/5 | 1 | llm | 2026-09-16 14:20 UTC |
| `a28035904771` | 3/5 | 2 | llm | 2026-09-16 15:28 UTC |
| `e4e92ecb53ab` | 3/5 | 4 | llm | 2026-09-16 16:22 UTC |

</details>

<sub>Reviewed `2fab2f1452bf` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 607s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

aw-sync/src/sync.rs
```diff
@@ -117,26 +133,51 @@ pub fn sync_run(
...
+                )),
+                Err(e) => {
+                    report.peers.push(PeerReport::failed(
+                        db.device_id.clone(),
+                        db.hostname.clone(),
+                        db.path.clone(),
+                        e.clone(),
+                    ));
+                    report.finish();
+                    crate::report::persist_last_report_warn(&report);
```
TimeToBuildBob, line 178: ❌ **P1** — In sync.rs, the `sync_run` function, when a pull fails, it pushes a `PeerReport::failed` and persists the report, then returns Err. However, the report's `finished` timestamp is set via `report.finish()` before persisting. This is correct. But the report is persisted even though the pass failed, and the caller (e.g., main.rs daemon) will not persist it again because it only persists on Ok. This is intentional. However, in the simple host-based sync path in main.rs, if pull_all succeeds but push fails, the outer report is not persisted because the `?` on push propagates the error before `aw_sync_persist` is called. This is the same issue I found earlier. I'll report that as a P1 bug.

aw-sync/src/report.rs
```diff
@@ -0,0 +1,565 @@
...
+            name,
+            std::process::id(),
+            std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .map(|d| d.as_nanos())
+                .unwrap_or(0)
+        ))
+    };
+    fs::write(&tmp, serde_json::to_string_pretty(report)?)?;
+    fs::rename(&tmp, path)?;
```
TimeToBuildBob, line 384: ❌ **P1** — In report.rs, the `persist_last_report_to` function uses `fs::rename` which is atomic on Unix but not on Windows if the destination exists. On Windows, `fs::rename` will fail if the destination file exists. This means that on Windows, the second persist will fail with an error, and the report will not be updated. This is a real bug for Windows users. The fix is to use `fs::rename` with a remove or use a platform-specific atomic replace. This is a P1 bug for Windows.

aw-sync/src/main.rs
```diff
@@ -268,26 +270,33 @@ fn main() -> Result<(), Box<dyn Error>> {
...
                         info!("Pulling from all hosts");
-                        sync_wrapper::pull_all(&client)?;
+                        report.merge(sync_wrapper::pull_all(&client)?);
                     }
                 }
 
                 // Push
                 info!("Pushing local data");
-                sync_wrapper::push(&client)?
+                report.merge(sync_wrapper::push(&client)?);
```
TimeToBuildBob: ⚠️ **P2** — In the simple host-based sync path in main.rs, the outer report is assembled by merging the pull report and the push report. If `sync_wrapper::push(&client)` returns `Err`, the `?` on line 296 propagates immediately and `aw_sync_persist(&report)` on line 299 never runs. The only report persisted comes from inside `push_with_hostname`'s `sync_run`, which is a Push-mode report containing just the push phase and any failure details. The pulled peers, warnings, and pulled-event counts from the earlier successful pull are discarded. `aw-sync status` then shows a push-only pass that failed, even though the pass had actually performed a successful pull first. The same pattern occurs in `Java_net_activitywatch_android_SyncInterface_syncBoth` where `report.merge(push_with_hostname(...)?)` loses the pull report when the push returns `Err`. Persisting the outer report before propagating the error would retain the pull information.

aw-sync/src/report.rs
TimeToBuildBob: False positive: `std::fs::rename` on Windows maps to `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING` (documented Rust std behavior), so it *does* replace an existing destination — it is not the raw Win32 `MoveFile` fail-if-exists semantics. The unique temp name stays because it protects concurrent writers, not the rename.

aw-sync/src/main.rs
TimeToBuildBob: Fixed in 49f0ce1 (rebased head): the simple host-based path in main.rs and the Android syncBoth JNI entry now capture the push result — on Err they push a warning onto the merged pull report, finish, persist it, and only then propagate. The pull-phase work is no longer dropped from last-sync-report.json or the JNI payload.

aw-sync/src/sync.rs
TimeToBuildBob: Valid finding, fixed at the callers rather than in sync.rs: commit 49f0ce1 (rebased head) changes main.rs and android.rs — the simple host-based path and the Android syncBoth JNI entry now capture the push result, push a warning onto the merged pull report, finish and persist it before propagating. sync_run itself already persisted correctly, so no change was needed there.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Rebased onto master (#698 landed): the `sync_run` conflict is resolved by keeping `maybe_setup_local_remote` plus the selection-based remote discovery, so the pull-only no-write invariant holds.

Also fixed the two real findings from the latest review in 49f0ce1: the simple host-based path and the Android `syncBoth` entry now capture the push result — on push failure the merged pull report (with a push warning) is finished and persisted before the error propagates, so `last-sync-report.json` no longer shows a push-only pass. The Windows `fs::rename` P1 was dismissed as a false positive (`MoveFileExW` + `MOVEFILE_REPLACE_EXISTING`). Re-review triggered; CI running on the new head.

aw-sync/src/sync_wrapper.rs
```diff
@@ -2,32 +2,66 @@ use std::error::Error;
...
+            skipped.db.path.clone(),
+            skipped.reason.clone(),
+        ));
+    }
+    let found: Vec<_> = selection.selected.iter().map(|d| d.path.clone()).collect();
+    // No get_info() here: an empty-dir pass must stay a local filesystem
+    // check (the #682 case). Own-vs-peer classification is optional for
+    // the warning text; "" leaves entries unclassified rather than
+    // hanging if the server is down.
+    report.capture_warnings(crate::util::pull_discovery_warnings(&sync_root, "", &found));
```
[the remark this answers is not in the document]
TimeToBuildBob: Leaving this. Greptile is 5/5 on 49f0ce1 and marked the finding non-blocking.

`pull_all` passes `""` so an empty-dir pass can diagnose without `get_info()` (ActivityWatch/aw-server-rust#682). That is not `scan_sync_dir(..., None)`: 3-level dbs classify as Peer (own is false when the local id is empty), and `path.contains("")` only hits the rare 2-level `device_id: None` branch. A real unknown-device kind is a status-doctor change, not a drive-by relabel here. The empty-pull dump still lists the files.

No further commit on this head. Next rebase waits on ActivityWatch/aw-server-rust#700 / ActivityWatch/aw-server-rust#678 / ActivityWatch/aw-server-rust#697 as sequenced.

TimeToBuildBob:

aw-sync/src/sync.rs
```diff
@@ -95,24 +94,41 @@ pub fn sync_run(
     }
 
     // Finding zero peers in a configured sync dir is the interesting case —
-    // ActivityWatch/aw-server-rust#684. Do not stay silent.
+    // ActivityWatch/aw-server-rust#682 / #695. Do not stay silent: keep the
+    // warnings on the report, not just in the log.
     if mode == SyncMode::Pull || mode == SyncMode::Both {
-        for line in crate::util::pull_discovery_warnings(
+        report.capture_warnings(crate::util::pull_discovery_warnings(
```
TimeToBuildBob, line 100: ❌ **P1** — In `sync_run`, the `remote_dbfiles` variable is built from `selection.selected` paths, and then `pull_discovery_warnings` is called with `&remote_dbfiles`. However, `pull_discovery_warnings` internally calls `scan_sync_dir` which uses `list_remote_dbs` and `select_remote_dbs_by_device_id` to classify entries. The `found_remotes` parameter is used to determine which peers were selected. But in `sync_run`, the selection is done via `find_remotes_nonlocal_selection`, which uses a different walker (`find_remotes` 2-level) than `list_remote_dbs` (3-level). This means the `found_remotes` list may not match the scan's view of the folder, causing `pull_discovery_warnings` to report missed peers that were actually selected, or vice versa. For example, in the advanced sync path with a 2-level layout, `find_remotes` finds `{device_id}/*.db` files, but `scan_sync_dir` classifies them as 2-level peers and may include them in the scan. If `found_remotes` contains those paths, the warning logic works. But if the sync dir has 3-level peers, `find_remotes` (2-level walker) will not find them, so `remote_dbfiles` is empty, and `pull_discovery_warnings` will report 'Found 0 remote db files' even though 3-level peers exist. This is a real bug: the advanced sync path (used by `aw-sync sync --start-date` etc.) will incorrectly warn about zero peers when the sync folder uses the 3-level layout, which is the default for the daemon. The consequence is a misleading warning in the report and log, and the sync will actually pull nothing because `find_remotes` doesn't see 3-level dbs. This is a pre-existing issue with the two walkers, but this PR makes it visible in the report and status, and the PR's own comments acknowledge the two walkers are intentional. However, the PR's `sync_run` now uses `find_remotes_nonlocal_selection` which is 2-level, so the advanced sync path still cannot see 3-level peers. This is a significant functional gap: users who use the advanced sync options with a 3-level sync folder will get a no-op pull with a false 'Found 0 remote db files' warning. The PR does not fix this, and the report makes it more prominent. I will report this as a bug because the PR's changes to sync_run now surface this misleading warning in the persisted report.

aw-sync/src/sync.rs
```diff
@@ -122,26 +138,51 @@ pub fn sync_run(
...
     }
 
     // Push local server buckets to sync folder
     if let Some(ds_localremote) = &ds_localremote {
         info!("Pushing...");
-        sync_datastores(client, ds_localremote, true, Some(device_id), sync_spec)?;
+        match sync_datastores(client, ds_localremote, true, Some(device_id), sync_spec) {
+            Ok(buckets) => report.pushed = buckets,
+            Err(e) => {
+                report.finish();
```
TimeToBuildBob, line 193: ❌ **P1** — In `sync_run`, the `report.finish()` is called before persisting on error paths, but on the success path it is called at the end. However, in the create_datastore error branch, `report.finish()` is called and then `persist_last_report_warn` is called, but the report is not returned (the function returns Err). The caller (e.g. main.rs daemon) will not persist again because it only persists on Ok. So the partial report is persisted once. That is correct. But in the pull loop error branch, the same pattern is used. In the push error branch, `report.finish()` and persist are called, but the report does not include any failed peer or warning about the push failure. The error string is only in the returned Err. This means the persisted report on a push failure looks like a successful (but empty) pass, with no indication of failure. The PR description says 'the partial report is persisted on that path so the failure is visible afterwards', but for push failures, the failure is not visible in the report. This is a real defect: after a push failure, `aw-sync status` will show a report with no peers, no pushed buckets, and no warnings, which looks like a no-op success rather than a failure. The user cannot tell from the report that the push failed. The fix would be to add a warning or a failed peer entry (e.g. a PeerReport::failed with the local device) before persisting.

aw-sync/src/main.rs
```diff
@@ -337,9 +357,15 @@ fn daemon(
...
     loop {
-        if let Err(e) = sync::sync_run(client, &sync_spec, mode) {
-            error!("Error during sync cycle: {}", e);
-            return Err(e);
+        match sync::sync_run(client, &sync_spec, mode) {
+            Ok(report) => {
+                info!("{}", report.summary_message());
+                aw_sync_persist(&report);
+            }
+            Err(e) => {
```
TimeToBuildBob, line 378: ⚠️ **P2** — In `main.rs`, the daemon loop calls `sync::sync_run` and on Ok persists the report. However, if `sync_run` returns an error, the daemon logs the error and returns Err, which exits the daemon. The PR description says 'a per-peer failure still aborts the pass', but the daemon exiting on any error is a pre-existing behavior. The new report persistence on error paths inside sync_run means the partial report is written, but the daemon still exits. That is consistent. No new defect.

aw-sync/src/sync.rs
TimeToBuildBob: Leaving this. Two walkers is intentional, documented at `list_remote_dbs` / `find_remotes` in `aw-sync/src/util.rs`.

`sync_run` is not the default pull path. `pull_all` (simple CLI, JNI, post-ActivityWatch/aw-server-rust#685 daemon) uses `list_remote_dbs` (3-level). When `sync_wrapper::pull` calls `sync_run`, `path` is a *host* folder, so the 2-level walk finds `{host}/{device_id}/*.db`.

The advanced `sync_run` against the sync root is the legacy 2-level path. Broadening `find_remotes` to 3-level at the root would import the leftover root orphan from ActivityWatch/aw-server-rust#682. Daemon host-layout is ActivityWatch/aw-server-rust#685, not this PR.

`pull_discovery_warnings` on a 3-level root with an empty `found_remotes` dumps the scan — that is the #695 visibility, not a false zero-peer diagnosis of a pull that actually ran.

aw-sync/src/sync.rs
TimeToBuildBob: Fixed in 91255d6. `sync_run`'s push-error branch now calls `record_push_failure` before persist, so `last-sync-report.json` carries `push failed: …` instead of looking like a no-op success. Same helper on the simple-path and JNI `syncBoth` callers. Test: `push_failure_persisted_report_is_not_a_silent_success`.

aw-sync/src/main.rs
TimeToBuildBob: Agreed — no new defect. Daemon exit-on-error is pre-existing; the partial report is already persisted on the `sync_run` error path. Per-peer isolation is ActivityWatch/aw-server-rust#688 / ActivityWatch/aw-server-rust#703, not this PR.

TimeToBuildBob:
12:48 AI-review on `49f0ce1` (score 2/5): one real hole, two not.

- **P1 push-abort silent report** — real. Inner `sync_run` persisted a push failure with no warning, so `last-sync-report.json` looked like a no-op. Fixed in 91255d6 (`record_push_failure` before persist; same helper on the simple-path and JNI callers). Test: `push_failure_persisted_report_is_not_a_silent_success`.
- **P1 two walkers** — rejected. `pull_all` is the 3-level path; `sync_run` is 2-level relative to the directory it is given (a host folder from `pull`, or the legacy root). Broadening `find_remotes` at the root would import the ActivityWatch/aw-server-rust#682 orphan. Daemon layout is ActivityWatch/aw-server-rust#685.
- **P2 daemon exit-on-error** — no new defect. ActivityWatch/aw-server-rust#688 / ActivityWatch/aw-server-rust#703.

Zero-peer report test is still in (Erik's ask). Next rebase still waits on ActivityWatch/aw-server-rust#700 / ActivityWatch/aw-server-rust#678 / ActivityWatch/aw-server-rust#697, not a merge click now.

TimeToBuildBob:
@greptileai review

aw-sync/src/sync.rs
```diff
@@ -122,26 +138,52 @@ pub fn sync_run(
...
     }
 
     // Push local server buckets to sync folder
     if let Some(ds_localremote) = &ds_localremote {
         info!("Pushing...");
-        sync_datastores(client, ds_localremote, true, Some(device_id), sync_spec)?;
+        match sync_datastores(client, ds_localremote, true, Some(device_id), sync_spec) {
+            Ok(buckets) => report.pushed = buckets,
+            Err(e) => {
+                report.record_push_failure(&e);
```
TimeToBuildBob, line 192: ⚠️ **P2** — In sync_run, when create_datastore fails for a remote db, the code records a PeerReport::failed and persists the report, but only if mode is Pull or Both. For Push mode, it returns Err without persisting anything. However, the more significant issue is that the report is persisted inside the loop before the datastore connections are closed. The report is written while ds_remotes may still hold open sqlite connections, but that is not the main problem. The main problem is that when sync_datastores fails for a peer, the code pushes a PeerReport::failed and persists the report, but the report's peers vector only contains entries for peers processed so far. The failing peer is added, but any peers that were already imported are also present. That is fine. However, the report is persisted with report.finish() called, but the caller (e.g., pull_all) may later merge this report into a larger report. The merge will extend peers, so the failed peer will appear. That is fine. The real issue is that in sync_run, when create_datastore fails, the code returns Err after persisting, but the caller (pull_all) will catch the Err and add another PeerReport::failed for the same peer, then persist again. This results in the same peer appearing twice in the final report: once from sync_run's internal persist (which is not merged because the Err is returned) and once from pull_all's failure handling. Actually, when sync_run returns Err, pull_all does not merge the report; it creates a new PeerReport::failed and pushes it into its own report. So the report persisted by sync_run is overwritten by pull_all's persist. The final report has one failed entry. That is fine. The issue is that sync_run persists a report that is then discarded, which is wasteful but not incorrect. The more concrete bug is in the push failure path in sync_run: when sync_datastores fails during push, the code records the failure and persists, but the report's mode is still Push (or Both). The summary_message for Push mode says 'Pushed X new events across Y bucket(s)' and does not mention the failure. The warning is in the warnings list, but the summary message does not surface it. The JNI to_jni_json includes warnings, so the Android side can see it, but the daemon logs summary_message which would say 'Pushed 0 new events across 0 bucket(s)' even though the push failed. This is misleading. The same applies to the main.rs simple sync path: on push failure, it records the failure and logs summary_message, which for mode Both says 'Synced X events in from ...; pushed Y events' without mentioning the failure. The user sees a success-like message despite the push failing. This is a real defect in the new reporting feature.

aw-sync/src/sync.rs
```diff
@@ -58,12 +51,17 @@ impl Default for SyncSpec {
...
+/// peers already imported) is persisted so `aw-sync status` can show the
+/// failure without the process having to stay alive.
 pub fn sync_run(
     client: &AwClient,
     sync_spec: &SyncSpec,
     mode: SyncMode,
-) -> Result<(), Box<dyn Error>> {
+) -> Result<SyncReport, Box<dyn Error>> {
+    let mut report = SyncReport::new(mode);
     let info = client.get_info()?;
```
TimeToBuildBob, line 65: ⚠️ **P2** — In pull_all, when a pull_db call fails, the code pushes a PeerReport::failed and persists the report, then returns Err. However, the report's mode is Pull, and the summary_message for Pull mode says 'Pulled X new events from Y/Z peers (skipped, failed)' which does include the failed count. That is fine. The issue is that the report is persisted with report.finish() before returning, but the caller (e.g., main.rs simple sync path) will catch the Err and not merge the report. The main.rs code in the simple sync path does: report.merge(sync_wrapper::pull_all(&client)?); If pull_all returns Err, the ? propagates and the report is not merged. The main.rs code then goes to the push phase? No, the ? returns from main, so the report is not persisted by main. But pull_all already persisted it. That is fine. The issue is that the report persisted by pull_all includes the failed peer, but the main.rs simple sync path's report (which was created before the pull) is discarded. The persisted report is the one from pull_all, which is correct. However, the main.rs simple sync path also has a push phase. If pull succeeds but push fails, the main.rs code records the push failure and persists the report. That report includes the pulled peers and the push failure warning. That is fine. The real issue is in the advanced sync path in main.rs: when sync_run returns Err, the code does not persist the report. The sync_run function itself persists a partial report on error, but only for certain error paths. For example, if client.get_info() fails at the start, sync_run returns Err without persisting anything. The PR description says 'On error, a partial report (failed peer, any peers already imported) is persisted so aw-sync status can show the failure without the process having to stay alive.' But that is not true for all error paths. If get_info() fails, no report is persisted. If find_remotes_nonlocal_selection fails, no report is persisted. If maybe_setup_local_remote fails, no report is persisted. Only the create_datastore, sync_datastores (pull and push) error paths persist. This is inconsistent and means a failure early in the pass leaves no report, so status shows the previous pass, which could be a success, misleading the user into thinking the last pass succeeded. This is a real defect in the new reporting feature.

aw-sync/src/sync_wrapper.rs
```diff
@@ -2,32 +2,66 @@ use std::error::Error;
...
             .map(|d| d.path.display().to_string())
             .collect::<Vec<_>>()
     );
-    for remote in selected {
-        pull_db(client, &remote.hostname, &remote.path)?;
+    for remote in selection.selected {
+        match pull_db(client, &remote.hostname, &remote.path) {
+            Ok(one) => report.merge(one),
+            Err(e) => {
+                report.peers.push(PeerReport::failed(
```
TimeToBuildBob, line 48: ⚠️ **P2** — In pull_all, when a pull_db call fails, the code pushes a PeerReport::failed and persists the report, then returns Err. However, the report's mode is Pull, and the failure is recorded as a peer failure. But the report's `peers` vector already contains skipped peers from the selection, and the failed peer is appended. The issue is that the report's `finished` timestamp is set via report.finish() before persisting, but the report's `started` timestamp is from the beginning of pull_all. This is fine. The real issue is that the error returned from pull_db is the error from sync_run, which may already have persisted a partial report internally (in sync_run's error paths). So pull_all persists an aggregate report, but sync_run also persisted a phase-local report. The later report overwrites the earlier one, which is the intended aggregate. However, if the failure occurs in pull_db before sync_run returns (e.g., client.wait_for_start fails), then sync_run never persisted, and pull_all's aggregate is the only one. That is fine. The issue is that the error message in the PeerReport::failed uses e.to_string(), which may be a long error, but that's not a bug. The actual bug is that pull_all's report mode is Pull, but when it merges a successful pull_db report, the merge sets mode to Both if modes differ. Since pull_db returns a Pull report, mode stays Pull. That's fine. I don't see a distinct bug here beyond what's already listed.

aw-sync/src/sync.rs
TimeToBuildBob: Leaving this. Early setup (`get_info`, `maybe_setup_local_remote`, `find_remotes_nonlocal_selection`) aborts before a pass exists — there is no failed peer and nothing imported. The PR contract is the parenthetical on `sync_run`: persist a *partial* report (failed peer, any peers already imported). Those branches already persist.

Overwriting `last-sync-report.json` with "couldn't talk to the server" would hide the last *completed* pass. The daemon already logs `Error during sync cycle` and exits; `aw-sync status` showing the previous pass is the right last-pass, not a silent success.

The pull_all / simple-path `?` path the finding opens with is already handled: `pull_all` persists the aggregate before returning Err, and the simple-path push-failure branch persists the merged pull report.

aw-sync/src/sync_wrapper.rs
TimeToBuildBob: Rejected. The finding walks `pull_all`'s error path and then says "I don't see a distinct bug here beyond what's already listed." That is correct — there isn't one.

`pull_all` already persists the aggregate (skipped peers + the failing peer) before returning Err. A nested `sync_run` persist is overwritten by that aggregate, which is the intended last-pass. Mode stays Pull because every merged child is Pull. No code change.

aw-sync/src/sync.rs
TimeToBuildBob: Fixed in 592c7b3. `summary_message` now appends `warnings`, so a push abort logs `Pushed 0 new events across 0 bucket(s); push failed: …` instead of a success-like line. Daemon, JNI `message`, and the existing `push_failure_persisted_report_is_not_a_silent_success` test all see it.

The create_datastore Push-mode persist aside is not a hole: Push does not import remotes; the abort is recorded via `record_push_failure` on the actual push path.

TimeToBuildBob:
13:24 AI-review on `91255d6` (score 4/5): one real hole, two not.

- **P2 `summary_message` hides push abort** — real. Daemon/JNI logged `Pushed 0 new events` even when `warnings` already had `push failed: …`. Fixed in 592c7b3 (append warnings to the summary; same test now asserts it).
- **P2 early setup does not persist** — trade-off. `get_info` / `find_remotes` abort before a pass exists; overwriting the last completed report with "couldn't talk to the server" would hide a useful last-pass. Partial persist stays on failed-peer paths.
- **P2 `pull_all` double persist** — rejected. No distinct bug; the aggregate overwrite is intended.

ActivityWatch/aw-server-rust#700 landed; still waiting on ActivityWatch/aw-server-rust#678 and ActivityWatch/aw-server-rust#697 before the one rebase Erik asked for. Not mergeable yet (`CONFLICTING`).

aw-sync/src/sync_wrapper.rs
```diff
@@ -2,32 +2,68 @@ use std::error::Error;
...
```
TimeToBuildBob, line 48: ❌ **P1** — In `sync_wrapper::pull_all`, when a `pull_db` call fails, the code pushes a `PeerReport::failed` and persists the aggregate report, then returns `Err(e)`. The caller in `main.rs` (simple sync mode) does `report.merge(sync_wrapper::pull_all(&client)?)` — the `?` propagates the error, so the aggregate report that was persisted inside `pull_all` is the one on disk. That is fine. But the caller in `android.rs` `syncPullAll` does `let report = pull_all(&client).map_err(...)?;` and then persists the report again. If `pull_all` fails, the error propagates and the report is not persisted again, but it was already persisted inside `pull_all`. That is fine. However, there is a double-persist on success: `pull_all` persists the report at the end (line 62-63), and then the Android caller persists it again (line 278). That is redundant but not a bug. The real issue is that `pull_all` persists the report even when it returns `Ok`, and then the caller also persists it. This means the report is written twice, but the second write overwrites the first with the same content. Not a defect. Let me look at the `merge` logic: in `pull_all`, each successful `pull_db` returns a `SyncReport` that is merged into the aggregate. The `pull_db` calls `sync_run` which already persisted its own report (on success, `sync_run` does not persist; the caller does). Actually `sync_run` does not persist on success; the caller does. In `pull_db`, `sync_run` returns a report, and `pull_db` returns it. Then `pull_all` merges it. So the per-peer report is not persisted until the aggregate is persisted. That is fine. But there is a subtle bug: when `pull_db` fails, `sync_run` has already persisted a partial report (on its error path). Then `pull_all` pushes a `PeerReport::failed` and persists the aggregate. So the aggregate includes the failed peer and any earlier peers. That is correct. However, the `PeerReport::failed` is constructed with `remote.device_id`, `remote.hostname`, `remote.path`, and `e.to_string()`. But the `sync_run` error might have already added a `PeerReport::failed` for the same peer? Let's trace: `pull_db` calls `sync_run` with a `SyncSpec` that has `path_db: Some(db_path)`. In `sync_run`, the `selection` is computed from `find_remotes_nonlocal_selection` with that `path_db`. If the db fails to open, `sync_run` adds a `PeerReport::failed` and persists, then returns `Err`. So `pull_db` returns `Err`. Then `pull_all` catches that `Err` and adds another `PeerReport::failed` for the same peer. So the aggregate report has two `PeerReport::failed` entries for the same peer. This is a real bug: the same failed peer is recorded twice, inflating the `peers_failed` count and showing duplicate entries in `aw-sync status`. The fix is to not add a `PeerReport::failed` in `pull_all` when the error already came from `sync_run` and already recorded the failure. But `pull_all` cannot easily know. Alternatively, `sync_run` should not persist on error when called from `pull_all`? But the design is that `sync_run` persists partial reports. The double-record is a concrete defect. The line is the `report.peers.push(PeerReport::failed(...))` in `pull_all`. _(2/3 passes)_

aw-sync/src/sync.rs
```diff
@@ -95,24 +94,42 @@ pub fn sync_run(
...
-    for path in &remote_dbfiles {
-        match create_datastore(path) {
-            Ok(ds) => ds_remotes.push(ds),
+    for db in &selection.selected {
+        match create_datastore(&db.path) {
+            Ok(ds) => ds_remotes.push((db.clone(), ds)),
             Err(e) => {
-                warn!("Failed to open remote db {}: {e}", path.display());
+                warn!("Failed to open remote db {}: {e}", db.path.display());
+                if mode == SyncMode::Pull || mode == SyncMode::Both {
```
TimeToBuildBob, line 135: ⚠️ **P2** — In `sync_run`, when `create_datastore` fails for a remote db, the code calls `close_opened_datastores(&ds_remotes, &ds_localremote)` and returns `Err`. But `ds_remotes` is a `Vec<(RemoteDb, Datastore)>` and the loop is iterating `&selection.selected`; the `ds_remotes` vector only contains datastores that were successfully opened before the failure. That is correct. However, the same error path also persists the report and then returns `Err(e.into())`. The caller in `main.rs` (advanced sync) uses `?` and does not persist the report again, so the partial report is persisted. But the caller in `sync_wrapper::pull_db` also uses `?` and does not persist again. That is fine. The issue is that on the `create_datastore` failure path, the report is persisted only if `mode == Pull || mode == Both`. If mode is Push, the report is not persisted and the error is returned. For a push-only pass, a failure to open a remote db (which is not even used in push mode) causes the pass to fail without persisting any report. But push mode does not open remote dbs? Actually the loop over `selection.selected` runs regardless of mode, so a push-only pass will try to open remote dbs and fail if one is corrupt, even though push does not need them. This is a pre-existing behavior? The original code also opened all remotes regardless of mode. But now the error path for push mode does not persist a report, so `aw-sync status` shows the previous pass, not the failure. The PR's stated goal is that a failed pass is visible. For push mode, a remote open failure aborts the pass and no report is persisted. This is a real defect: a push-only pass that fails because of an unrelated remote db leaves no record of the failure. The fix is to persist the report on the error path regardless of mode, or to skip opening remotes in push mode. The latter is a larger change; the former is simple. The line is the `if mode == SyncMode::Pull || mode == SyncMode::Both` guard around the persist call. _(2/3 passes)_

aw-sync/src/sync_wrapper.rs
TimeToBuildBob: Rejected. `pull_all` does not merge `sync_run`'s report on `Err` — it builds a new aggregate and overwrites `last-sync-report.json`. The nested persist is discarded; the on-disk report has one `PeerReport::failed` for that peer plus any earlier successes. Same walk as 715715ee49ba.

The "two failed entries" claim would only hold if the two reports were merged in memory. They are not.

aw-sync/src/sync.rs
TimeToBuildBob: Leaving this. Push still opens remotes (pre-existing; not this PR). A push-only pass that dies on an unrelated corrupt peer db is an abort before a push happened — overwriting the last completed report with that would hide a useful last-pass, same trade-off as get_info / find_remotes.

Skipping remotes in push mode is a separate change. The close-on-error path already runs regardless of mode.

TimeToBuildBob:
15:38 in-band review on `592c7b3` (score 4/5) plus DIRTY vs master after ActivityWatch/aw-server-rust#700.

Pushed `a280359`:

- **Empty local device_id** — `pull_all` passed `""` into `pull_discovery_warnings`, which is `Some("")`, not unclassified. Leftover 2-level own staging then showed up as "peer db(s) that pull did not select". Now `None`, and 2-level leftovers are not missed-peers when the id is unknown. Still no `get_info()` on this path (120s HTTP timeout; ActivityWatch/aw-server-rust#682 stays a filesystem check). Test: `unknown_local_id_does_not_report_two_level_leftover_as_unselected_peer`.
- **Close datastore workers on error returns** — `Datastore::close()` stops the sqlite worker; the new persist-then-return paths skipped it.

Did **not** rebase. ActivityWatch/aw-server-rust#678 and ActivityWatch/aw-server-rust#697 are still open; one rebase after both land, as asked.

In-band P1 "double-recorded failed peer" rejected: `pull_all` overwrites the nested persist, it does not merge it. P2 "persist push-mode remote-open failure" left as the same last-pass trade-off.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Rebased onto master (f8374cd) after #700/#697 landed. The sync.rs conflict was the read-only peer-open loop: `sync_run` now walks `selection.selected` and opens each peer with `open_peer_datastore` (read-only, no migration/WAL flips), keeping the SyncReport machinery — open failures record a failed `PeerReport` + persist, version-mismatch skips with the existing warning. `cargo test -p aw-sync` green (33 lib + 11 sync + 3 roundtrip). Greptile re-triggered.

aw-sync/src/sync.rs
```diff
@@ -95,26 +94,44 @@ pub fn sync_run(
...
     // Peer files are opened read-only: never migrate, never flip WAL
     // (ActivityWatch/aw-server-rust#693). Version mismatch is skipped, not fatal.
     let mut ds_remotes = Vec::new();
-    for path in &remote_dbfiles {
-        match open_peer_datastore(path) {
-            Ok(Some(ds)) => ds_remotes.push(ds),
+    for db in &selection.selected {
+        match open_peer_datastore(&db.path) {
+            Ok(Some(ds)) => ds_remotes.push((db.clone(), ds)),
             Ok(None) => {}
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in e4e92ec: `open_peer_datastore` now returns `OpenedPeer::Incompatible(msg)` instead of `Ok(None)`, and `sync_run` records each incompatible peer as `PeerReport::skipped` with the version-mismatch reason. An all-incompatible pass now shows the rejected peers (and their reason) in status/JNI instead of a clean empty report. Tests + clippy green.

TimeToBuildBob:
@greptileai review

aw-sync/src/main.rs
```diff
@@ -268,26 +270,44 @@ fn main() -> Result<(), Box<dyn Error>> {
...
             } else {
                 // Simple host-based sync mode (backwards compatibility)
+                let mut report = sync::SyncReport::new(sync::SyncMode::Both);
                 // Pull
                 match host {
                     Some(hosts) => {
                         for host in hosts.iter() {
                             info!("Pulling from host: {}", host);
-                            sync_wrapper::pull(host, &client)?;
+                            report.merge(sync_wrapper::pull(host, &client)?);
```
TimeToBuildBob: ❌ **P1** — In `main.rs` simple host-based sync, the per-host pull loop merges each `pull` report with `report.merge(sync_wrapper::pull(host, &client)?);`. When a later host's pull fails, the `?` propagates the error immediately, skipping the push section and the final `aw_sync_persist(&report)`. At that point `report` already contains the successfully pulled hosts' imported peers and discovery warnings, but it is never persisted; instead the failed `pull`'s `sync_run` persists only its own partial report (the failing host's peer failure and that single-host report). The earlier hosts' imported events are therefore absent from `last-sync-report.json`, so `aw-sync status` shows only the partial failure, contradicting the PR's goal that the persisted report reflects the pass's actual work. The fix is to catch the error, record it on the aggregate report (e.g., a warning or failed-peer entry), persist, then re-return the error.

aw-sync/src/sync.rs
TimeToBuildBob, line 65: ⚠️ **P2** — In `sync_run`, the `SyncReport` is created before `let info = client.get_info()?;`. If the server is unreachable or the request fails, this line returns the error immediately, and the report is dropped without ever calling `report.finish()` or `persist_last_report_warn`. Consequently a daemon that loses its connection for one cycle leaves `last-sync-report.json` from the previous successful pass, and `aw-sync status` continues to display that stale success. The PR's contract is that any pass leaves a visible report (see the 'failure visible afterwards' language), but this path does not. The fix is to record a warning on the report (e.g., 'server unreachable') and persist it before propagating the error.

aw-sync/src/sync.rs
```diff
@@ -95,26 +94,55 @@ pub fn sync_run(
...
+                        db.hostname.clone(),
+                        db.path.clone(),
+                        format!("incompatible database version: {msg}"),
+                    ));
+                }
+            }
             Err(e) => {
-                warn!("Failed to open remote db {}: {e}", path.display());
+                warn!("Failed to open remote db {}: {e}", db.path.display());
+                if mode == SyncMode::Pull || mode == SyncMode::Both {
```
TimeToBuildBob, line 135: ⚠️ **P2** — In `sync_run`'s peer-datastore open error path, the `Err` arm calls `report.finish()` and `persist_last_report_warn` only when `mode == SyncMode::Pull || mode == SyncMode::Both`. In `SyncMode::Push`, the same error just closes the datastores and returns `Err`, leaving no report behind. A push-only pass (`aw-sync daemon --mode push` or JNI `syncPush`) that aborts because a peer database cannot be opened therefore leaves the previous report on disk, which may show a success. This contradicts the PR's stated intent that a failed pass is visible afterwards. The fix is to unconditionally `report.finish()` and persist (at least a warning) on this error path, regardless of mode.

aw-sync/src/sync.rs
TimeToBuildBob: Leaving this. Early setup (`get_info`, `maybe_setup_local_remote`, `find_remotes_nonlocal_selection`) aborts before a pass exists — there is no failed peer and nothing imported. The PR contract is the parenthetical on `sync_run`: persist a *partial* report (failed peer, any peers already imported). Those branches already persist.

Overwriting the last completed report with "couldn't talk to the server" would hide a useful last-pass. Same trade-off as the previous pass on this path.

aw-sync/src/sync.rs
TimeToBuildBob: Leaving this. Push still opens remotes (pre-existing; not this PR). A push-only pass that dies on an unrelated corrupt peer db is an abort before a push happened — overwriting the last completed report with that would hide a useful last-pass, same trade-off as get_info / find_remotes.

Skipping remotes in push mode is a separate change. The close-on-error path already runs regardless of mode. Same finding as 23be81b9f2e5 after the #697 rebase moved the open loop onto `open_peer_datastore`.

aw-sync/src/main.rs
TimeToBuildBob: Fixed in 2fab2f1. The per-host pull loop now catches a later host's error, records `pull from {host} failed: …` on the aggregate (so earlier hosts stay), persists, then propagates — same contract as the push-failure path. Test: `later_host_pull_failure_keeps_earlier_hosts_on_disk`.

TimeToBuildBob:
16:41 AI-review on `e4e92ec` (score 3/5): one real hole, two not.

- **P1 later-host pull drops earlier hosts** — real. `--host a --host b` used `?`, so a later host's failure skipped `aw_sync_persist` of the aggregate and `last-sync-report.json` kept only the failing host. Fixed in 2fab2f1 (`record_pull_failure` + persist before propagate). Test: `later_host_pull_failure_keeps_earlier_hosts_on_disk`.
- **P2 `get_info` does not persist** — trade-off. Early setup aborts before a pass exists; overwriting the last completed report would hide a useful last-pass.
- **P2 push-mode open failure does not persist** — same trade-off as 23be81b9f2e5 after the ActivityWatch/aw-server-rust#697 rebase. Skipping remotes in push mode is a separate change.

`cargo test -p aw-sync --lib --bins --tests` green (34 lib + 11 sync + 3 roundtrip). In-band review timed out at 1200s; push landed on 2fab2f1. Sweep will re-review the new head.

TimeToBuildBob:
Local re-review on `2fab2f1` (the new head): score 4/5, no P0/P1. The later-host pull P1 is confirmed fixed and no new blocking findings reproduce. Remaining is the known P2 (push-mode remote open failure leaves the last completed report intact) — same accepted trade-off as dispositioned above.

Finding threads all carry replies now; leaving resolution to the sweep on this head.

aw-sync/src/sync.rs
```diff
@@ -95,26 +94,55 @@ pub fn sync_run(
...
+                    report.peers.push(PeerReport::skipped(
+                        db.device_id.clone(),
+                        db.hostname.clone(),
+                        db.path.clone(),
+                        format!("incompatible database version: {msg}"),
+                    ));
+                }
+            }
             Err(e) => {
```
TimeToBuildBob, line 133: ⚠️ **P2** — In sync_run, when open_peer_datastore returns Err, the code pushes a PeerReport::failed, persists the report, closes datastores, and returns Err. However, the report is only persisted inside the `if mode == SyncMode::Pull || mode == SyncMode::Both` block. If mode is Push, the error path still closes datastores and returns Err, but the report is not persisted and no PeerReport is added. This means a push-only pass that fails to open a remote db (which is not even used in push mode) will not leave a failure record, contradicting the PR's stated contract that a failed pass is visible afterwards. The observable consequence is that `aw-sync status` shows the previous successful report or 'none' after a push-only failure, making the failure invisible.

TimeToBuildBob: Disposing as an accepted trade-off — same adjudication as the earlier round on this file (15:39), re-filed here with a different fingerprint.

`sync_run`'s contract is to persist a *partial pass report*: the failing peer plus any peers already imported, so a mid-pass abort does not hide work that actually happened.

A push-only pass that dies in `open_peer_datastore` is different — it aborted *before* a pass exists. Nothing was pushed and nothing was imported, so there is no partial state to record. Persisting here would overwrite the last completed report with an aborted one, which is strictly worse for anyone reading `last-sync-report.json`. That is the same trade-off already taken for the other early-abort paths in this function (`get_info`, `maybe_setup_local_remote`, `find_remotes_nonlocal_selection`).

The failure is also not invisible: the error is propagated to the caller, so the CLI exits non-zero and the daemon logs it. Only the on-disk report keeps the last completed pass — by design.

Separately: opening peer datastores in push mode is pre-existing behavior, not introduced by this PR (the read-only peer open is the ActivityWatch/aw-server-rust#693 work). Not changing it here.

TimeToBuildBob:
Merge verified, not just "merged": full CI is green on the merge commit `37aa0b0` (ubuntu/macOS/windows/clippy/format/coverage/Android all success), and `ActivityWatch/aw-server-rust#695` is closed.

Downstream consumer is already wired: `ActivityWatch/aw-android#285` bumps the `aw-server-rust` submodule `97f51e3` → `37aa0b0` (this merge commit) and parses the `to_jni_json()` counts into `SyncStatus`, so the app-side half of `ActivityWatch/aw-android#274` no longer discards the report.

Deliberately unchanged, as scoped here: a per-peer failure still aborts the pass. That remainder is tracked at `ActivityWatch/aw-server-rust#688`, with `ActivityWatch/aw-server-rust#703` in flight.

No further action needed from me on this PR.

PR #700, opened 2026-09-16
fix(aw-sync): open peer databases read-only; skip version mismatch

TimeToBuildBob:
v0.14.0 slice of ActivityWatch/aw-server-rust#693, as sequenced on ActivityWatch/activitywatch#1445. Independent of ActivityWatch/aw-server-rust#685.

## Problem

`sync_run` opened every peer db with `Datastore::new`, which is a write:

- `journal_mode=WAL` creates `-wal`/`-shm` sidecars in a folder this device does not own
- `DatastoreInstance::new(&conn, true)` migrates `user_version` on files it does not own
- a v4 peer (`user_version != 6`) aborted the whole pass

Android and `aw-sync sync` already pull today, so this is live, not latent on the daemon switch.

## Fix

- `Datastore::open_read_only` opens `file:…?mode=ro&immutable=1`, skips WAL/synchronous pragmas, and never migrates
- `user_version != NEWEST_DB_VERSION` returns `OldDbVersion`; aw-sync skips that peer with a warning and keeps walking
- own staging (`create_datastore` / `setup_local_remote`) is unchanged

Not in this PR: a tolerant reader for old index names (the full #693 work item). v4 peers are skipped, not imported.

## Tests

- `test_read_only_open_does_not_create_wal_sidecars`
- `test_read_only_open_skips_old_user_version`


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `aw-datastore/src/worker.rs` (+76/-16, Δ92)
- `aw-datastore/tests/datastore.rs` (+77/-1, Δ78)
- `aw-sync/src/sync.rs` (+36/-12, Δ48)
- `aw-datastore/src/lib.rs` (+5/-0, Δ5)
- `aw-datastore/src/datastore.rs` (+1/-1, Δ2)

### Suggested verification

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for ActivityWatch/aw-server-rust#700._

ErikBjare:
@TimeToBuildBob This does what step 4 asked. Verified: peer opens are `mode=ro&immutable=1` + `SQLITE_OPEN_READ_ONLY`, migrations disabled for read-only (`DatastoreInstance::new(&conn, !read_only)`), and a `user_version` mismatch returns `Ok(None)` with a warn (`sync.rs:219-226`) — a real skip, not an `Err`. Clean test-merges against #698, #697, #678 and #699.

Two notes, neither blocking:

**1. The residual per-peer abort is not this PR's job — but it should be named.** `sync.rs:113` still does `return Err(e.into())` when a peer fails to open for any reason *other* than version mismatch (corrupt file, permission, truncated by the syncer mid-copy), and `pull_all` still has `pull_db(...)?`. So this PR makes the *expected* case (old peer) non-fatal, while the *unexpected* case still aborts the pass and skips every peer after it. That is #697's per-peer `continue` (already in my review there). Fine to leave here; just do not let #697 land without it.

**2. Please document what `immutable=1` gives up.** It tells SQLite the file cannot change, so it never touches `-wal`/`-shm` — which is exactly why it is correct here (no sidecars created in a foreign directory, no locking). The consequence: committed-but-not-yet-checkpointed frames in a peer's `-wal` are invisible to the read. Since aw-sync checkpoints on clean close, the steady-state file is self-contained and this only bites mid-push — a stale-but-consistent read, which is strictly better than a torn one. Worth a sentence in the doc comment on the read-only open so nobody later "fixes" it by dropping `immutable` and reintroduces `-shm` files in peers' folders.

Also for Erik's awareness: this touches `aw-datastore` (the read-only open path in `worker.rs`), so it is a shared-crate change, not sync-only — though the new path is only reachable via the new constructor.

Ready to merge when CI is green. Queue position: after #698, and it does not need to wait for #697.

TimeToBuildBob:
Documented on `Datastore::open_read_only` and the URI helper in 72f69a8: `immutable=1` is load-bearing (no `-wal`/`-shm` in a foreign directory); dropping it to "see the WAL" is the failure mode. Cost is mid-push frames in a peer's WAL being invisible — stale-but-consistent, as you said.

Also switched the read-only worker to `BEGIN DEFERRED`. Greptile's hang (Immediate → SQLITE_READONLY → retry forever) did not reproduce on this SQLite (`test_read_only_open_does_not_create_wal_sidecars` already called `get_buckets`/`get_events` and passed), but Immediate is still a write lock on a read-only connection. Deferred is the correct behavior either way.

Left as named: unexpected open errors still abort the pass until ActivityWatch/aw-server-rust#697's per-peer `continue` lands. UNC paths are not handled; not a v0.14.0 target.

Still queued after ActivityWatch/aw-server-rust#698.

aw-datastore/src/worker.rs
```diff
@@ -156,6 +194,9 @@ impl DatastoreWorker {
             DatastoreMethod::File(path) => {
                 Connection::open(path).expect("Failed to create datastore")
             }
+            DatastoreMethod::FileReadOnly(path) => {
+                open_readonly_connection(path).expect("Failed to open datastore read-only")
+            }
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 72f69a8: read-only workers now use BEGIN DEFERRED. Immediate did not hang on this SQLite (the sidecar test already called get_buckets/get_events), but it is a write lock on a read-only connection either way.

aw-datastore/src/worker.rs
```diff
@@ -26,6 +27,41 @@ use crate::DatastoreMethod;
...
+        .replace('%', "%25")
+        .replace('?', "%3F")
+        .replace('#', "%23");
+    if cfg!(windows) {
+        encoded = encoded.replace('\\', "/");
+        if encoded.len() >= 2 && encoded.as_bytes().get(1) == Some(&b':') {
+            return format!("file:///{encoded}?mode=ro&immutable=1");
+        }
+    }
+    format!("file:{encoded}?mode=ro&immutable=1")
```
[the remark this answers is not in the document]
TimeToBuildBob: Intentional. immutable=1 is why a pull does not create -wal/-shm in a foreign directory. Documented on Datastore::open_read_only in 72f69a8 — dropping it to follow a live peer WAL is the failure mode this PR exists to stop. Mid-push WAL frames are invisible; that is a stale-but-consistent snapshot, which is better than a torn one.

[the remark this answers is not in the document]
TimeToBuildBob: Out of scope for this PR. Peer sync folders on a Windows UNC share are not a v0.14.0 target. The drive-letter branch covers the Windows CI path.

ErikBjare:
@TimeToBuildBob Assessed Greptile's three P1s. One is a false positive I verified empirically, one is a doc comment, one is a real two-line fix. Two small changes and this merges.

## P1 "Read-only workers hang" — false positive, but the fragility is real

Greptile's reasoning is sound on its face: `worker.rs:268` begins every request with `TransactionBehavior::Immediate` unconditionally, and on `Err` it logs, sleeps 1s and `continue`s without ever receiving a request — so if `BEGIN IMMEDIATE` failed on the read-only connection, `get_buckets`/`get_events`/`close` would block forever.

It does not fail. I ran the new test on this head with a 150 s timeout:

```
test datastore_tests::test_read_only_open_does_not_create_wal_sidecars ... ok
test datastore_tests::test_read_only_open_skips_old_user_version ... ok
test result: ok. 2 passed; finished in 0.01s
```

That test exercises real `get_buckets()` / `get_events()` through the worker loop on an `open_read_only` handle. CI green on ubuntu/windows/macOS/android says the same cross-platform. The reason: with `immutable=1`, SQLite treats the file as read-only media and skips locking entirely, so `BEGIN IMMEDIATE` has no RESERVED lock to acquire and succeeds as a no-op read transaction; only an actual page write would return `SQLITE_READONLY`.

**Why it still needs a one-line change:** that safety depends on `immutable=1`. A plain `mode=ro` connection *does* reject `BEGIN IMMEDIATE`, and then the retry loop hangs exactly as described. So if anyone later drops `immutable` — e.g. to address the next finding the naive way — this P1 becomes real. Please make the worker use `TransactionBehavior::Deferred` when `read_only` so correctness does not hinge on the URI flag:

```rust
let behavior = if read_only { TransactionBehavior::Deferred } else { TransactionBehavior::Immediate };
```

## P1 "Mutable files marked immutable" — keep `immutable=1`, document it

Real in principle: the connection lives across a multi-page pull and `immutable=1` disables change detection. In practice Syncthing and Dropbox write a temp file and rename, so an open handle keeps the old inode on POSIX (a consistent old snapshot) and blocks the rename on Windows (the syncer retries later). Only in-place rewriting — `rsync --inplace`, a naive `cp` over the file — defeats it.

Do **not** fix this by dropping `immutable=1`: that would (a) create `-shm` in peers' folders again, which is what #693 exists to stop, and (b) trigger the hang above. Add a doc comment on `sqlite_readonly_uri` stating the trade and the in-place-rewrite caveat. Copy-then-open is the belt-and-braces option if it ever bites; not needed now.

## P1 "UNC paths become invalid" — real, fix before merge

`\\server\share\peer.db` → backslashes replaced → `//server/share/peer.db` → misses the drive-letter branch → `file://server/share/peer.db?…` with `server` as the URI authority. Narrow (a sync dir on a network share), but trivial: if the converted path starts with `//`, emit `file:////server/share/…` (SQLite's documented UNC form). Worth a unit test on `sqlite_readonly_uri` for `C:\…`, `\\server\share\…` and a POSIX path.

With the `Deferred` line, the UNC guard and the doc comment, nothing else from me — merge on green.

TimeToBuildBob:
@ErikBjare The three P1s as you assessed, now on `d260b16`:

1. Hang — false positive here (`immutable=1` makes `BEGIN IMMEDIATE` a no-op), but the retry loop is still a trap if the flag is dropped. `72f69a8` already switched read-only workers to `BEGIN DEFERRED`.

2. `immutable=1` stays. Doc comment on `sqlite_readonly_uri` now names the trade, the Immediate hang if it is dropped, and the in-place-rewrite caveat (Syncthing/Dropbox rename is fine; `rsync --inplace` is not). Copy-then-open left for later.

3. UNC — that was the remaining gap. Path-shape detection (not `cfg!(windows)`) emits `file:////server/share/…` instead of treating `server` as a URI authority. Unit tests cover POSIX, `C:\…`, and `\\server\share\…`.

Residual unexpected-open abort is still ActivityWatch/aw-server-rust#697's `continue`. Queue after ActivityWatch/aw-server-rust#698.

Retriggering Greptile on this head.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Greptile's fourth P1 on re-review was also real: `sqlite_readonly_uri` treated any backslash as Windows-path evidence, but a backslash is a valid POSIX filename character — a peer path containing one would get corrupted (`\` → `/`) and could open the wrong file or fail entirely.

Fixed on `955c516`: classification now requires an unambiguous Windows shape (drive letter or UNC prefix), not mere backslash presence. Added a regression test (`posix_path_with_literal_backslash_is_not_rewritten`). Drive-letter and UNC cases unchanged.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 1 P1 disposed (rejected)

| --- | --- | --- |
| `aw-datastore/src/worker.rs:53` | P1 | rejected — The reviewer inferred SQLite would parse file:/home/user/sync:dir/peer.db as authority home + path /user/sync:dir/peer.d |
| `aw-datastore/src/worker.rs:62` | P2 | superseded by latest review (not reproduced) |
| `aw-datastore/src/worker.rs:598` | P2 | superseded by latest review (not reproduced) |
| `aw-datastore/src/worker.rs:261` | P2 | superseded by latest review (not reproduced) |
| `aw-datastore/src/worker.rs:305` | P2 | wontfix |
| `aw-datastore/src/worker.rs:598` | P2 | wontfix |
| `aw-datastore/tests/datastore.rs:1148` | P2 | wontfix |
| `aw-sync/src/sync.rs:108` | P2 | superseded by latest review (not reproduced) |

</details>




This PR adds a read-only datastore open path for aw-sync peer databases. It introduces DatastoreMethod::FileReadOnly, a sqlite_readonly_uri helper that builds a file: URI with mode=ro&immutable=1, and Datastore::open_read_only which probes user_version and returns OldDbVersion on mismatch. aw-sync's sync_run and list_buckets now use open_peer_datastore to skip version-mismatched peers instead of aborting. The worker skips WAL/synchronous pragmas and uses Deferred transactions for read-only connections.

### Needs a look — P2 only

Confidence 4/5

<details><summary><b>1 finding</b> · ⚠️ 1 P2</summary>

**⚠️ P2 medium** — `aw-datastore/src/worker.rs:598`

The open_read_only function probes user_version by opening a read-only connection, then closes it (the connection is dropped at the end of probe_user_version). Then it opens a second read-only connection via Datastore::_new_internal. Between these two opens, the peer database could be modified (e.g., by the peer's own process) and its user_version could change. However, the more significant issue is that the version check is done on a separate connection from the one used for the actual read. If the peer database is replaced between the probe and the open (e.g., by a file syncer like Syncthing), the second connection might open a different file with a different version, but the version check has already passed. This is a TOCTOU race. The consequence is that a pull could read from a database whose version was not actually checked, potentially reading an incompatible schema. However, the read-only connection with immutable=1 will read whatever file is at the path at open time, and if the file was replaced with a different version, the read could fail or return garbage. This is a race condition, but the window is small and the impact is likely a failed read rather than corruption.

<sub>**How this was verified:** probe_user_version opens a connection and drops it (lines 80-88). open_read_only then calls _new_internal which opens a new connection in the worker thread (line 222-224). The two opens are separate.</sub>

</details>

**⚠️ P2 medium** — `aw-datastore/src/worker.rs:69`

The read-only connection uses immutable=1, which tells SQLite the database file never changes. If a peer database is actively being written by another process (e.g. the peer's aw-server is running and pushing to the same sync folder), the read-only connection will not see committed WAL frames and may read a stale or inconsistent snapshot. The PR documentation acknowledges this but the code does not guard against it. More critically, immutable=1 with a file that is being modified can cause SQLite to return SQLITE_CORRUPT or read torn pages if the file is rewritten in place. The observable consequence is that a pull may silently miss recent events from a peer that is mid-write, or fail with a corruption error. This is a trade-off documented in the code, but the severity is that it can cause silent data loss in the pull path.

<sub>**How this was verified:** Checked the doc comment on open_read_only (lines 591-596) which explicitly states 'Committed-but-not-yet-checkpointed frames in that WAL are therefore invisible.' Also checked that the test creates a database, checkpoints, and deletes WAL sidecars before opening read-only, so it does not exercise the active-WAL case.</sub>

</details>

- `aw-datastore/src/datastore.rs` — Makes NEWEST_DB_VERSION public so other crates can compare against it.
- `aw-datastore/src/lib.rs` — Adds DatastoreMethod::FileReadOnly variant and re-exports NEWEST_DB_VERSION.
- `aw-datastore/src/worker.rs` — Adds sqlite_readonly_uri, open_readonly_connection, probe_user_version, and Datastore::open_read_only; skips WAL/synchronous and uses Deferred transactions for read-only.
- `aw-datastore/tests/datastore.rs` — Adds tests for read-only open not creating WAL sidecars and skipping old user_version.
- `aw-sync/src/sync.rs` — Replaces create_datastore for peer files with open_peer_datastore that skips version-mismatched peers; adds utf8_db_path helper.

</details>

| `d260b1683f53` | 4/5 | 4 | llm | 2026-09-16 09:15 UTC |
| `955c5161fbb8` | 3/5 | 3 | llm | 2026-09-16 09:57 UTC |

</details>

<sub>Reviewed `3c1eeb10d394` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 123s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

aw-datastore/src/worker.rs
```diff
@@ -26,6 +27,67 @@ use crate::DatastoreMethod;
...
+    let mut encoded = path
+        .replace('%', "%25")
+        .replace('?', "%3F")
+        .replace('#', "%23");
+    let looks_windows = encoded.contains('\\')
+        || encoded.starts_with("//")
+        || (encoded.len() >= 2 && encoded.as_bytes().get(1) == Some(&b':'));
+    if looks_windows {
+        encoded = encoded.replace('\\', "/");
+        if encoded.len() >= 2 && encoded.as_bytes().get(1) == Some(&b':') {
```
TimeToBuildBob: ⚠️ **P2** — The `sqlite_readonly_uri` function encodes `%`, `?`, and `#` in the path, but it does not encode other URI-reserved characters such as `;`, `/` (when not a path separator), or `@`. More importantly, it does not handle the case where the path itself contains a `?` or `#` that is already percent-encoded. The function replaces `%` first, so a literal `%3F` in the path becomes `%253F`, which is correct. However, the Windows path detection checks `encoded.contains('\')` after encoding, which is fine. The real issue is that the function does not handle paths that are already URIs or contain characters like `[` and `]` which are invalid in URIs. But the more concrete bug: for a POSIX path that contains a `?` (e.g., `/tmp/foo?bar.db`), the encoded path becomes `/tmp/foo%3Fbar.db`, and the resulting URI is `file:/tmp/foo%3Fbar.db?mode=ro&immutable=1`. SQLite will decode `%3F` back to `?` in the filename, so that works. However, the function does not encode `%` in the query string, but the query string is fixed. The actual bug is that the function does not handle paths that start with `file:` or contain a colon in the first segment (e.g., a relative path like `foo:bar.db`). For a relative path `foo:bar.db`, the function produces `file:foo:bar.db?mode=ro&immutable=1`, which SQLite may interpret as a URI with scheme `foo`? Actually, SQLite's URI parsing: if the filename starts with `file:`, it's a URI. `file:foo:bar.db` is a valid URI with path `foo:bar.db`? SQLite might treat the part after `file:` as the path, so `foo:bar.db` is fine. But a path like `C:foo.db` (drive-relative) would be detected as Windows because byte 1 is `:`, and it would produce `file:///C:foo.db` which is wrong (should be `file:///C:/foo.db`). The code only handles the case where the second byte is `:` and the path is absolute (starts with `C:\`). For a drive-relative path like `C:foo.db`, it produces `file:///C:foo.db` which SQLite may interpret as a file named `C:foo.db` on the current drive, not the intended file. This is a real bug for Windows drive-relative paths, but such paths are rare in sync directories. The severity is low.

aw-sync/src/sync.rs
```diff
@@ -101,11 +101,13 @@ pub fn sync_run(
...
     }
 
-    // TODO: Check for compatible remote db version before opening
+    // Peer files are opened read-only: never migrate, never flip WAL
+    // (ActivityWatch/aw-server-rust#693). Version mismatch is skipped, not fatal.
     let mut ds_remotes = Vec::new();
     for path in &remote_dbfiles {
-        match create_datastore(path) {
-            Ok(ds) => ds_remotes.push(ds),
+        match open_peer_datastore(path) {
```
TimeToBuildBob, line 108: ⚠️ **P2** — The open_peer_datastore function in aw-sync/src/sync.rs returns Ok(None) for OldDbVersion, but it also swallows the error message. The caller in sync_run logs a warning and continues, which is intended. However, in list_buckets, the same function is used, and if a peer has an old version, it is silently skipped without any warning. The list_buckets function is used for debugging and should probably warn as well. More importantly, the error handling in sync_run: if open_peer_datastore returns Err (e.g., a read-only open failure), the entire sync pass aborts with return Err(e.into()). This means a single unreadable peer file (e.g., permission denied) aborts the whole sync, whereas the PR's goal is to skip problematic peers. The PR description says 'skip version mismatch', but other errors still abort. This is a robustness issue: a peer with a transient I/O error will block syncing with all other peers.

aw-datastore/src/worker.rs
```diff
@@ -508,6 +582,34 @@ impl Datastore {
...
+    /// `user_version != NEWEST_DB_VERSION` so a caller can skip that peer
+    /// (ActivityWatch/aw-server-rust#693).
+    ///
+    /// `immutable=1` means SQLite will not look at a peer's `-wal`/`-shm`.
+    /// Committed-but-not-yet-checkpointed frames in that WAL are therefore
+    /// invisible. aw-sync checkpoints on clean close, so the steady-state
+    /// file is self-contained; this only bites mid-push. A stale-but-
+    /// consistent snapshot is strictly better than a torn one, and dropping
+    /// `immutable` would reintroduce `-shm` files in a foreign directory.
+    pub fn open_read_only(dbpath: String) -> Result<Self, DatastoreError> {
```
TimeToBuildBob, line 597: ⚠️ **P2** — The Datastore::open_read_only method takes dbpath: String, but the underlying sqlite_readonly_uri function assumes the path is a filesystem path. If the caller passes a path that is already a URI (e.g., from a config file), the function will mangle it. However, all current callers pass filesystem paths from find_remotes_nonlocal, so this is not currently exploitable. The public API is new, so this is a contract issue: the function does not document that the path must be a plain filesystem path, not a URI. A future caller could pass a URI and get a confusing error. This is a minor documentation/contract gap.

aw-datastore/src/worker.rs
```diff
@@ -176,24 +243,22 @@ impl DatastoreWorker {
...
+                warn!("Failed to enable WAL (journal_mode={journal_mode}), continuing without it");
+            }
+            conn.pragma_update(None, "synchronous", "FULL")
+                .expect("Failed to set synchronous=FULL");
         }
-        conn.pragma_update(None, "synchronous", "FULL")
-            .expect("Failed to set synchronous=FULL");
 
-        let mut ds = DatastoreInstance::new(&conn, true).unwrap();
+        let mut ds = DatastoreInstance::new(&conn, !read_only).unwrap();
```
TimeToBuildBob, line 260: ⚠️ **P2** — `Datastore::open_read_only` first probes `user_version` via a temporary read-only connection, then spawns the worker thread which opens a second connection and calls `DatastoreInstance::new(&conn, !read_only)` with `migrate_enabled=false`. That call re-validates the version and returns `OldDbVersion` if the version changed, but the result is `.unwrap()`ed at line 261, so a version change between the probe and the worker's open (for example, a peer upgrading its schema, or a syncer replacing the file with a different-version copy) panics the worker thread. The `Datastore` returned to the caller then has a dead worker; every subsequent request fails with `InternalError` ("...datastore worker is gone"), and `sync_run` aborts the whole pass with an error instead of skipping that peer. The race window is small but real. The worker should propagate the error (for example, through a startup handshake) rather than unwrap, or the version check should be performed on the same connection that serves requests.

TimeToBuildBob:

ErikBjare:
@TimeToBuildBob `955c516` has all three: `Deferred` when `read_only` (`:296-299`), the UNC guard with the four-slash form, and the doc comment on `immutable`. Greptile's re-review raised one real thing.

**Required — gate the backslash rewrite on platform, not path shape (`worker.rs:61`).** `encoded.replace('\\', "/")` and the UNC detection now run unconditionally. A literal backslash is a legal character in a POSIX filename; rewriting it makes the read-only probe target a different file — wrong database, or an open failure that aborts the pass. Wrap the Windows normalisation (backslash → slash, drive-letter and UNC detection) in `if cfg!(windows)` and pass POSIX paths through untouched. One line plus a test with a POSIX path containing `\`.

Then merge on green. Not rebased on current master but `MERGEABLE`, which is fine for a squash.

aw-datastore/src/worker.rs
```diff
@@ -26,6 +27,69 @@ use crate::DatastoreMethod;
...
+/// retries). Only in-place rewriting (`rsync --inplace`, a naive `cp` over
+/// the file) defeats it. Copy-then-open is the belt-and-braces option if
+/// that ever bites; not needed now.
+///
+/// Windows path shapes (drive letter, UNC) are detected from the path, not
+/// `cfg!(windows)`, so a `\\server\share\peer.db` sync dir produces
+/// SQLite's UNC form `file:////server/share/…` instead of treating `server`
+/// as a URI authority.
+fn sqlite_readonly_uri(path: &str) -> String {
+    let encoded = path
```
TimeToBuildBob, line 53: ❌ **P1** — The read-only URI builder encodes '?' and '#' but not ':' in the path. On POSIX, a path containing a colon is legal, and SQLite URI parsing treats the first ':' after 'file' as the separator between the authority and the path. For example, a sync directory path like '/home/user/sync:dir/peer.db' produces 'file:/home/user/sync:dir/peer.db?mode=ro&immutable=1'. SQLite parses the URI as authority 'home' and path '/user/sync:dir/peer.db', which resolves to a different file than the one the caller intended, so the read-only open either fails with 'unable to open database file' or, worse, opens a different peer's database. The same issue affects Windows drive-letter paths that already contain a colon, but those are handled by the drive-letter branch which prefixes 'file:///'; the POSIX branch is the vulnerable one. The fix is to percent-encode ':' as '%3A' in the path before building the URI, or to use the 'file:' URI with an explicit authority and encoded path. This is a correctness bug that can cause aw-sync to read the wrong file or fail to open a valid peer.

aw-datastore/tests/datastore.rs
```diff
@@ -1124,4 +1124,80 @@ mod datastore_tests {
...
+            ds.insert_events("testid", &[test_event(Utc::now(), Duration::seconds(1))])
+                .unwrap();
+            ds.force_commit().unwrap();
+            ds.close();
+        }
+
+        // Checkpoint so we can delete WAL sidecars without losing the row.
+        {
+            let conn = rusqlite::Connection::open(&db_path).unwrap();
+            let _ = conn.execute_batch("PRAGMA wal_checkpoint(TRUNCATE);");
```
TimeToBuildBob, line 1148: ⚠️ **P2** — The new test test_read_only_open_does_not_create_wal_sidecars is not hermetic: it relies on the database being fully checkpointed and the WAL sidecars deleted before the read-only open. The test creates a database with Datastore::new, which sets journal_mode=WAL, then opens a separate rusqlite connection to run 'PRAGMA wal_checkpoint(TRUNCATE)' and deletes the -wal and -shm files. However, the first Datastore::new connection is closed with ds.close(), but the worker thread may not have fully released the file handle on Windows (the test even acknowledges this in another test's cleanup comment). If the checkpoint connection is opened while the original connection still holds a lock, the checkpoint may fail silently (the result is ignored with 'let _ ='), and the subsequent delete of -wal/-shm may fail on Windows due to the file being locked. The test then asserts that the read-only open does not create sidecars, but if the sidecars were never deleted in the first place, the assertion '!wal.exists()' would fail even though the read-only open did not create them. This makes the test flaky on Windows and could give a false failure. The test should ensure the original connection is fully closed and the checkpoint succeeds before deleting sidecars.

aw-datastore/src/worker.rs
```diff
@@ -222,19 +289,28 @@ impl DatastoreWorker {
...
-                match conn.transaction_with_behavior(TransactionBehavior::Immediate) {
-                    Ok(tx) => tx,
-                    Err(err) => {
-                        error!("Unable to start transaction! {:?}", err);
-                        // Wait 1s before retrying
-                        std::thread::sleep(std::time::Duration::from_millis(1000));
-                        continue;
-                    }
-                };
+            let mut tx: Transaction = match conn.transaction_with_behavior(tx_behavior) {
```
TimeToBuildBob, line 302: ⚠️ **P2** — The read-only connection is opened with OpenFlags::SQLITE_OPEN_READ_ONLY | SQLITE_OPEN_URI, but the worker thread still calls conn.busy_timeout(5s) and then enters a loop that calls conn.transaction_with_behavior(TransactionBehavior::Deferred). For a read-only connection, a deferred transaction is fine, but the worker also calls tx.set_drop_behavior(DropBehavior::Commit) and later tx.commit(). If the transaction is read-only, commit is a no-op, but the drop behavior Commit means that if the transaction is dropped without an explicit commit (e.g., on an error path), it will attempt to commit, which on a read-only connection may return an error. More importantly, the worker's error handling for transaction start failures retries forever with a 1-second sleep. If the read-only connection encounters a persistent error (e.g., the file is locked or immutable=1 causes a read error), the worker thread will spin forever, logging errors every second, and the Datastore::request calls will hang indefinitely because the worker never responds. This is a pre-existing issue for the normal path, but the read-only path is new and more likely to hit persistent errors (e.g., SQLITE_READONLY on a file that cannot be opened read-only). The PR should consider making the read-only worker exit on transaction start failure instead of retrying forever, or at least surface the error to the caller.

TimeToBuildBob:
@ErikBjare Done on `3c1eeb1`. Windows normalisation (backslash → slash, drive-letter, UNC) is inside `if cfg!(windows)`; POSIX paths, including a literal `\`, pass through untouched. Regression: `posix_path_with_literal_backslash_is_not_rewritten` (`cfg(not(windows))`). Drive-letter/UNC URI tests are `cfg(windows)` so they don't lie on Linux CI.

Cannot self-merge here (pull-only on ActivityWatch/aw-server-rust). Squash when green — ActivityWatch/aw-server-rust#698 is already merged, so this is unblocked on queue. Residual unexpected-open abort remains ActivityWatch/aw-server-rust#697.

aw-datastore/src/worker.rs
```diff
@@ -26,6 +27,67 @@ use crate::DatastoreMethod;
...
+fn sqlite_readonly_uri(path: &str) -> String {
+    let mut encoded = path
+        .replace('%', "%25")
+        .replace('?', "%3F")
+        .replace('#', "%23");
+    let looks_windows = encoded.contains('\\')
+        || encoded.starts_with("//")
+        || (encoded.len() >= 2 && encoded.as_bytes().get(1) == Some(&b':'));
+    if looks_windows {
+        encoded = encoded.replace('\\', "/");
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 3c1eeb1: Windows normalisation is now `cfg!(windows)`-gated, not path-shape. POSIX paths with a literal backslash pass through untouched.

TimeToBuildBob:
@greptileai review

aw-datastore/src/worker.rs
TimeToBuildBob: False positive — empirically verified on this host.

`file:/tmp/…/sync:dir/peer.db?mode=ro` (the POSIX one-slash form `sqlite_readonly_uri` emits) opened the colon-in-directory file and returned the row we inserted. A colon in the filename (`peer:name.db`) also opened correctly. SQLite only treats the first path component as authority on the two-slash form (`file://host/path`); that form is what failed here (`invalid uri authority: tmp`). The reviewer mixed up `file:/home/…` (path-absolute, no authority) with `file://home/…`.

Windows drive-letter paths already go through `file:///{encoded}` inside `cfg!(windows)`. No code change. Erik's instruction is merge on green.

aw-datastore/tests/datastore.rs
TimeToBuildBob: Won't-fix for this PR. Windows CI on `3c1eeb1` passed (`windows-latest` green), so the sidecar-delete race did not fire. Tightening checkpoint/close ordering is follow-up, not a merge blocker — Erik's instruction is merge on green.

aw-datastore/src/worker.rs
TimeToBuildBob: Won't-fix for this PR. Forever-retry on transaction start is pre-existing on the read-write worker. Read-only now uses BEGIN DEFERRED (`72f69a8`), which Erik already assessed as the right fix for the hang P1. Changing the retry loop is a separate robustness change, not in the v0.14.0 slice.

ErikBjare:
Verified on `3c1eeb1`: backslash/UNC normalisation gated on `cfg!(windows)`, POSIX paths pass through. All green. Nothing further — ready for Erik.

aw-datastore/src/worker.rs
```diff
@@ -508,6 +581,34 @@ impl Datastore {
...
+    /// (ActivityWatch/aw-server-rust#693).
+    ///
+    /// `immutable=1` means SQLite will not look at a peer's `-wal`/`-shm`.
+    /// Committed-but-not-yet-checkpointed frames in that WAL are therefore
+    /// invisible. aw-sync checkpoints on clean close, so the steady-state
+    /// file is self-contained; this only bites mid-push. A stale-but-
+    /// consistent snapshot is strictly better than a torn one, and dropping
+    /// `immutable` would reintroduce `-shm` files in a foreign directory.
+    pub fn open_read_only(dbpath: String) -> Result<Self, DatastoreError> {
+        let version = probe_user_version(&dbpath)?;
```
TimeToBuildBob, line 598: ⚠️ **P2** — The open_read_only function probes user_version by opening a read-only connection, then closes it (the connection is dropped at the end of probe_user_version). Then it opens a second read-only connection via Datastore::_new_internal. Between these two opens, the peer database could be modified (e.g., by the peer's own process) and its user_version could change. However, the more significant issue is that the version check is done on a separate connection from the one used for the actual read. If the peer database is replaced between the probe and the open (e.g., by a file syncer like Syncthing), the second connection might open a different file with a different version, but the version check has already passed. This is a TOCTOU race. The consequence is that a pull could read from a database whose version was not actually checked, potentially reading an incompatible schema. However, the read-only connection with immutable=1 will read whatever file is at the path at open time, and if the file was replaced with a different version, the read could fail or return garbage. This is a race condition, but the window is small and the impact is likely a failed read rather than corruption.

TimeToBuildBob: Won't-fix for this PR. The probe-then-open TOCTOU is real but the window is two sequential opens of the same path. `immutable=1` plus Syncthing/Dropbox rename semantics means the worker either keeps the probed inode or opens a complete replacement; `DatastoreInstance::new(&conn, false)` then re-checks `user_version` on that connection. A mismatch panics the worker (pre-existing `.unwrap()` at `worker.rs:260`) and the caller sees `InternalError` rather than a silent schema read. Closing it properly is a startup handshake on the serving connection — not the v0.14.0 slice. Erik's 10:14 note is merge-on-green / nothing further.

TimeToBuildBob:
Ack the 10:14 verification on `3c1eeb1`. Still pull-only here — squash is yours.

Leftover AI-review P2 (probe/open TOCTOU, `11a2bbc2dc60`) disposed as wontfix: small window, worker re-validates version, not a merge blocker.

TimeToBuildBob:
Merged (`d2ae2e6`). Nothing further on this PR.

Next squash is ActivityWatch/aw-server-rust#678 (CLEAN, all green). ActivityWatch/aw-server-rust#697 is now CONFLICTING against this merge; a sibling session already holds the rebase. ActivityWatch/aw-server-rust#703 stays last, after #697.

Issue #689, opened 2026-09-16
aw-sync: leftovers after #685/#686 — orphaned 2-level staging db, stale -synced-from- buckets, walker enters dot-dirs


ErikBjare:
Collecting the cleanup items that #685 / #686 / #687 deliberately leave behind, so they do not get lost once those merge.

## 1. The orphaned 2-level staging database

#682's root cause is that `aw-sync daemon` staged local data at `{sync_dir}/{device_id}/test.db` instead of `{sync_dir}/{hostname}/{device_id}/test.db`. #685 fixes where *new* pushes go but does not touch what is already there, as @TimeToBuildBob notes on that PR.

So every user who ran the daemon has an orphaned database at the sync root that:

- no peer reads (after #686, `list_remote_dbs` is 3-level-only, so it is not even a pull candidate)
- Syncthing replicates to every device, forever
- is large — 1.19 GB on my own setup, next to a 273 MB legitimate peer db

Wanted: a one-time migration, or at minimum detection. `aw-sync status` (#687) already classifies sync-dir entries by layout, so it is the natural place to report "this is an orphaned legacy staging db, N bytes, safe to delete" — and a `--clean` or documented manual step to remove it. Deleting it from inside a synced folder propagates to every peer, which is the desired outcome but should be an explicit user action, not a silent one.

## 2. Stale `-synced-from-` buckets inside staging databases

Before #648, the push path re-exported buckets that had themselves been synced from another host. That fix stopped new ones being created; nothing prunes the existing ones. My root staging db still contains, for example:

```
aw-watcher-window_erb-laptop2-arch-synced-from-erb-laptop2-arch   652321 events
aw-watcher-afk_erb-main2-arch-synced-from-erb-main2-arch          373525 events
aw-watcher-window_erb-main3-synced-from-erb-main3                 190854 events
```

These are second-hand copies sitting in a file that is replicated to every device. They are filtered on read by `is_synced_bucket`, so they are not a correctness problem any more — they are dead weight, and they make the "is this staging db healthy?" question harder to answer. Worth reporting in `status` and offering to prune.

## 3. The walker descends into dot-directories

#685's `collect_db_files` recurses into every directory at depth 0 and 1 with no name filter, so it walks `.git`, `.stfolder`, `.stversions` and similar. On my sync folder that means walking a real `.git` tree on every pass.

[...]

PR #702, opened 2026-09-16
fix(aw-sync): skip dot-directories when walking peer databases

TimeToBuildBob:
Fixes item 3 of ActivityWatch/aw-server-rust#689. Independent of the parked daemon-layout switch in ActivityWatch/aw-server-rust#685.

## Problem

`list_remote_dbs` (3-level `pull_all` walker) and `find_remotes` (2-level walker) recurse into every directory at the first (and, for 3-level, second) level with no name filter. That includes `.git`, `.stfolder`, and `.stversions`.

No `.db` files live there in the common case, so this was not a correctness bug — but both walkers now propagate I/O errors rather than swallowing them. An unreadable entry under a metadata dir aborts the whole pass. Syncthing Trash Can versioning can also place a `.db` at `{sync}/.stversions/*.db`, shallow enough for the 2-level walker to treat it as a peer.

## Fix

Skip `.`-prefixed directories in:

- `list_remote_dbs` (host and device levels)
- `find_remotes` (first level)
- `scan_sync_dir` (so `aw-sync status` does not walk `.git` either)

Not in this PR: leftover 2-level staging db cleanup, stale `-synced-from-` prune, or the `test.db` rename (items 1, 2, 4 of ActivityWatch/aw-server-rust#689). Those wait on the layout we are not switching to in v0.14.0.

## Tests

- `list_remote_dbs_skips_dot_directories`
- `find_remotes_skips_dot_directories`
- `scan_sync_dir_skips_dot_directories`

Does not touch the `sync_run` region rewritten in ActivityWatch/aw-server-rust#698, so it is independent of the ActivityWatch/aw-server-rust#700 → #678 → #697 merge order.


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `aw-sync/src/util.rs` (+88/-4, Δ92)

### Suggested verification

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for ActivityWatch/aw-server-rust#702._

ErikBjare:
One `is_dot_dir` guard applied to `list_remote_dbs`, `find_remotes` and `scan_sync_dir`, with a test per walker — exactly #689 item 3. No legitimate peer directory starts with `.`, and this stops a pass walking `.git` / `.stfolder` / `.stversions` on every cycle. Merge on green; independent of the #700 → #678 → #697 order.

TimeToBuildBob:

TimeToBuildBob:
## 🤖 AI code review

This PR adds a helper `is_dot_dir` and applies it to the directory-walking loops in `list_remote_dbs`, `find_remotes`, and `scan_sync_dir` so that dot-prefixed directories (e.g. `.git`, `.stfolder`, `.stversions`) are skipped. It also adds three tests covering the new skip behavior in each walker.

### Safe to merge — no P0/P1 findings

Confidence 5/5

✅ **No thread-worthy findings.** Advisory notes follow; they are retained without opening review threads.

**⚠️ P2 medium** — `aw-sync/src/util.rs`

```suggestion
Add a test file that reproduces the bug before the fix and passes after it.
```

</details>

- `aw-sync/src/util.rs` — Adds `is_dot_dir` helper and uses it in three walkers; adds three tests for dot-directory skipping.

</details>

| `2ecfb10f16cf` | 4/5 | 2 | llm | 2026-09-16 10:12 UTC |
| `d7cc1eaa2e77` | 4/5 | 2 | llm | 2026-09-17 09:41 UTC |

</details>

<sub>Reviewed `db6c5680599d` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 9s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

aw-sync/src/util.rs
```diff
@@ -1111,6 +1172,29 @@ mod scan_tests {
         let _ = fs::remove_dir_all(root);
     }
 
+    #[test]
+    fn scan_sync_dir_skips_dot_directories() {
+        let root = temp_sync_dir();
+        let peer = "peer-device";
+        touch_db(&root.join("poco_f8_ultra").join(peer).join("test.db"), 16);
+        touch_db(&root.join(".stversions").join(peer).join("test.db"), 64);
```
TimeToBuildBob, line 1324: ⚠️ **P2** — The test `scan_sync_dir_skips_dot_directories` asserts that no entry path contains a dot-directory component, but it does not actually verify that the `.stversions` database is skipped. The test creates `root/.stversions/peer-device/test.db` and then asserts `scan.len() == 1` and `scan[0].kind == Peer`. However, `scan_sync_dir` calls `list_remote_dbs` first, which now skips `.stversions` at the host level, so the `.stversions` db is not in `remotes`. Then the top-level loop in `scan_sync_dir` skips `.stversions` because of the new `is_dot_dir` check. So the test does pass, but it is not a strong test: it does not distinguish between the new behavior and a hypothetical behavior where `scan_sync_dir` skips dot-dirs but `list_remote_dbs` does not. However, the test does cover the new code path in `scan_sync_dir` (the `is_dot_dir` check at line 633). The assertion `scan.len() == 1` would fail if the `.stversions` db were included, so it is not vacuous. The test is adequate.

TimeToBuildBob: False positive — no change needed. The finding's own verification already says the test is adequate: `scan.len() == 1` plus the path-component check fail if `.stversions` leaked in. Distinguishing `list_remote_dbs` skip vs `scan_sync_dir` overlay skip is not this test's job — `list_remote_dbs_skips_dot_directories` already covers the walker, and the overlay `is_dot_dir` at the top-level loop is what stops `.stfolder`/`.git` being classified as Unrecognised. The test is the user-visible composition, not a vacuous assert.

Same pass's unanchored "file-level skip" finding (`97dee644920d`) is also a false positive. The file-level loop in `list_remote_dbs` does not recurse: non-files `continue` on `path.is_file()`, and `fs::read_dir` is never called on a `.stfolder`/`.stversions` child of a device dir. `Path::is_file()` does not propagate I/O errors. The suggested `if path.is_dir() && is_dot_dir(&path) { continue; }` is a no-op for the stated abort mode. Device-level skip is already tested via `{host}/.stfolder/test.db`.

Erik signed off merge-on-green on this exact host+device+overlay shape. Not adding cargo-cult continues after that.

ErikBjare:

TimeToBuildBob:
Rebased onto `3d3b9dd` (head `2b58d4a`) — master should not move again before this lands. The conflict was confined to the test module in `aw-sync/src/util.rs` (both #699's and #678's new tests landed in the same spot); no logic changed, both test sets kept. CI is running on the new head.

ErikBjare:
@TimeToBuildBob Re-verified on `2b58d4a`: `is_dot_dir` guards all three walkers (`list_remote_dbs` `:403`/`:412`, `find_remotes` `:532`, `scan_sync_dir` `:718`), one test each. Content is done.

**But the `format` check failed** — `cargo fmt --check` is red on this head. Run `cargo fmt` and push; no other change needed. Master moved once more (#697 landed) but this PR is still `MERGEABLE` against it, so no further rebase — just the fmt push.

ErikBjare:
Correction: the master move was #701 (README) landing, not #697 — #697 is still open. Doesn't change the ask: this PR is `MERGEABLE` against current master, so a `cargo fmt` push is all it needs.

ErikBjare:
`72d099c`: one-line rustfmt fix, `format` now green, content unchanged from the verified `2b58d4a`. Merge on green — after #697, any time relative to #703.

TimeToBuildBob:
Two follow-up pushes since the rebase comment: an intermediate head had a stray brace in the test module (caught by local `cargo check`, fixed in `72d099c`), then a fast-forward over the #701 docs merge (docs-only, no conflict). Final head is `d7cc1ea`; `cargo check -p aw-sync --all-targets` and the 19 `util` lib tests pass locally. CI is running on it.

aw-sync/src/util.rs
```diff
@@ -869,6 +920,16 @@ fn file_name_string(path: &Path) -> Option<String> {
...
 }
 
+/// Syncthing (`.stfolder`, `.stversions`), git, and similar metadata dirs
+/// are not host folders. Walking them is wasted I/O and, because the
+/// walkers propagate I/O errors, an unreadable entry under one of them
+/// aborts the whole pass (ActivityWatch/aw-server-rust#689).
+fn is_dot_dir(path: &Path) -> bool {
+    path.file_name()
+        .and_then(|n| n.to_str())
+        .is_some_and(|n| n.starts_with('.'))
```
TimeToBuildBob, line 933: ⚠️ **P2** — The `is_dot_dir` helper uses `file_name().to_str()` and `starts_with('.')`. For directory names that are not valid UTF-8, `to_str()` returns `None`, so `is_dot_dir` returns `false`. This means a directory whose name begins with a dot but contains non-UTF8 bytes (e.g. `.�` or a raw byte 0x2E followed by 0xFF) would not be skipped. On Unix, filenames are arbitrary bytes, so a directory named `.�` (invalid UTF-8) would be treated as a host folder and walked. This is an edge case, but the walkers are meant to skip all dot-directories. The consequence is that an unreadable non-UTF8 dot-directory could still abort the whole pass, which is the exact problem the PR aims to fix. However, this is a very rare edge case and the impact is limited. I'll report as a guard or heuristic? The PR's stated goal is to skip dot-directories; a non-UTF8 dot-dir is still a dot-dir. But the code uses `to_str()` which fails. A more robust check would use `as_encoded_bytes()` or `to_string_lossy()`. This is a real gap, but low severity. I'll report as a bug? It's a correctness issue for a concrete input (non-UTF8 dot-dir). But the input is exotic. I'll report as P2 with class `bug`.

aw-sync/src/util.rs
```diff
@@ -1252,6 +1313,29 @@ mod scan_tests {
...
+        let _ = fs::remove_dir_all(&root);
+
+        assert!(
+            scan.iter().all(|e| e
+                .path
+                .components()
+                .all(|c| c.as_os_str().to_str().is_none_or(|s| !s.starts_with('.')))),
+            "dot-dirs must not appear as sync entries: {scan:?}"
+        );
+        assert_eq!(scan.len(), 1);
```
TimeToBuildBob, line 1338: ⚠️ **P2** — The test `list_remote_dbs_skips_dot_directories` creates a hidden device at `root.join("poco_f8_ultra").join(".stfolder").join("test.db")` and expects `listed.len() == 1`. This passes because `list_remote_dbs` skips the `.stfolder` device-level directory. However, the test does not verify that a dot-directory at the host level is skipped when it contains a `.db` at the device level? It does: `write_remote_db(&root, ".stversions", "device-1", 128)` creates `{root}/.stversions/device-1/test.db`, and the host-level check skips `.stversions`. So that's covered. The test is fine. But the `scan_sync_dir_skips_dot_directories` test has a vacuous assertion: it asserts that all entries have no dot components, but it also asserts `scan.len() == 1`. If the dot-dirs were not skipped, the scan would include entries from `.stversions/peer/test.db` and possibly `.git/objects` (no db), so `scan.len()` would be >1. The assertion `scan.len() == 1` would fail, so it's not vacuous. However, the first assertion about dot components is redundant but not vacuous. The test is fine. I don't see a test defect here.

aw-sync/src/util.rs
TimeToBuildBob: Fixed in db6c568: switched `is_dot_dir` from `to_str()` to `to_string_lossy()`, so a non-UTF8 dot-directory name (leading `.` byte, invalid bytes after) is still recognized. Verified with `cargo check -p aw-sync --all-targets` and the 19 `util` lib tests locally; `cargo fmt --check` and `cargo clippy` clean.

aw-sync/src/util.rs
TimeToBuildBob: Reviewer's own reasoning walks this back to "I don't see a test defect here" — `scan.len() == 1` is non-vacuous since a missed dot-dir would inflate the count. False positive, no change.

TimeToBuildBob:
AI review threads addressed: fixed the real one (non-UTF8 dot-dir names not recognized by `is_dot_dir` — now uses `to_string_lossy()`) in `db6c568`, and resolved the self-refuting test-defect finding as FP (the reviewer's own text concludes "I don't see a test defect here"). Content otherwise unchanged from the head Erik re-verified. Waiting on CI for `db6c568` before merging per the standing "merge on green" authorization.

ErikBjare:
`db6c568` verified: `to_str()` → `to_string_lossy()` in `is_dot_dir` is a real fix — a non-UTF8 dot-directory name previously came back `None` and was *not* skipped; the leading `.` survives lossy conversion. Four lines, walkers and tests otherwise unchanged from the verified `d7cc1ea`. Still `MERGEABLE` against current master; merge on green.

TimeToBuildBob:
CI green on `db6c568` (all 7 checks pass), both AI-review threads resolved, `mergeStateStatus: CLEAN`. Ready for merge — I don't have write access to this repo, so this needs your click.

TimeToBuildBob:
Merge verified live — `94bfc544` is on master and the `is_dot_dir` guards are present in all three walkers (`list_remote_dbs` :403/:412, `find_remotes` :532, `scan_sync_dir` :718), with the `to_string_lossy` fix from `db6c568` retained. Nothing further needed on this PR; ActivityWatch/aw-server-rust#689's remaining items (1–2) still wait for the v0.14.0 cut.

PR #703, opened 2026-09-16
fix(aw-sync): skip broken peers in sync_run instead of aborting the pass

### LINKED ISSUE #688: aw-sync: per-peer and per-bucket errors abort the whole sync pass (and burn the supervisor's restart budget)

TimeToBuildBob:
Fixes the **daemon-path** half of #688.

The supervised `aw-sync` daemon walks remotes inside `sync_run`. A single unreadable peer used `?` and skipped every peer after it, then exited non-zero and burned aw-qt/aw-tauri's restart budget. That is the path users actually run; ActivityWatch/aw-server-rust#685 (host-layout `pull_all`) is parked for v0.14.0, so ActivityWatch/aw-server-rust#697's per-peer `continue` in `pull_all` does not cover the daemon.

## Change

- `sync_run` skips an unreadable peer db instead of `return Err`.
- New `pull_from_remotes`: warn+continue on each remote's `sync_datastores` failure. Partial skip is Ok; `attempted > 0 && succeeded == 0` is still Err so a down destination is not reported as success.

## Tests

- `broken_peer_does_not_skip_healthy_sibling` — broken peer listed first, healthy sibling still imports.
- `all_peers_failing_is_still_err` — total failure stays Err.

## Not this PR

- Per-bucket skip inside `sync_datastores` is ActivityWatch/aw-server-rust#697.
- Read-only open + version-mismatch skip is ActivityWatch/aw-server-rust#700. That PR's `open_peer_datastore` Err arm still `return Err` for corrupt/unreadable files — when it lands, that arm should `continue` like this, or #688 stays open on the daemon path.
- Skipped peers in `aw-sync status` is still open on #688.

Does not close #688 on its own.


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `aw-sync/src/sync.rs` (+111/-5, Δ116)

### Suggested verification

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for ActivityWatch/aw-server-rust#703._

ErikBjare:
@TimeToBuildBob Content is right — this closes the residual I named on #700: a peer that fails to *open* for a non-version reason now skips instead of aborting the pass, and the same `attempted > 0 && succeeded == 0 → Err` rule as #697 keeps total failure visible to the supervisor. With this, #688's per-peer isolation is complete across all three layers (open, per-remote sync, per-bucket).

**Sequence: last, after #697.** It edits the same `sync_run` remote loop that #700 and #697 touch. Pairwise it is clean against both, but that was also true of #697/#699 before #698 squash-merged, and both conflicted afterwards. Expect to rebase once #700 and #697 are in; merge on green after that. Does not change Erik's current order (#700 → #678 → #697).

TimeToBuildBob:
Got it. Stays last, after #697.

Not merging now. Order remains #700 → #678 → #697. Once #700 and #697 are in I'll rebase this onto master and merge on green.

Greptile's dest-failure P1 is the #697 rule you just signed off — leaving it. The open-failure counting P1 is the rebase against #700: that PR's `open_peer_datastore` Err arm still `return Err`; this PR turns it into skip. I'll fold that in on the rebase, not ahead of the stack.

TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 1 P1 disposed (fixed in 789b13c)

| --- | --- | --- |
| `aw-sync/src/sync.rs:351` | P1 | fixed in `789b13c` |
| `aw-sync/src/sync.rs:751` | P2 | superseded by latest review (not reproduced) |
| `aw-sync/src/sync.rs:114` | P2 | rejected |
| `aw-sync/src/sync.rs:378` | P2 | superseded by latest review (not reproduced) |
| `aw-sync/src/sync.rs:920` | P2 | rejected |
| `aw-sync/src/sync.rs:1364` | P2 | fixed in `9b49d4a` |
| `aw-sync/src/sync.rs:124` | P2 | rejected |
| `aw-sync/src/sync.rs:353` | P2 | accepted-tradeoff |
| `aw-sync/src/sync.rs:1228` | P2 | other |

</details>




This PR changes aw-sync's sync_run so that a single unreadable or incompatible peer database no longer aborts the whole sync pass. It adds open_peer_datastores() and pull_from_remotes() helpers that skip failing peers while recording them on the report, and only return Err when every peer fails. It also adds a new test module peer_isolation_tests covering partial and total failure cases.

### Needs a look — P2 only

Confidence 4/5

<details><summary><b>2 findings</b> · ⚠️ 2 P2 — non-blocking (round cap reached)</summary>

**⚠️ P2 medium · ℹ️ non-blocking** — `aw-sync/src/sync.rs:353`

In open_peer_datastores(), the condition `if opened.is_empty() && failures > 0` treats a pass where every discovered peer is version-incompatible (OpenedPeer::Incompatible) as success, because failures only counts hard open errors. The PR description and comments say a total open failure should be Err so a broken folder is not reported as a clean empty pass, but an all-incompatible folder is reported as Ok with an empty ds_remotes, and sync_run then proceeds to push (in Both mode) and returns Ok. This contradicts the stated intent in the comment on lines 120-122 and the PR description's 'only a total open failure is fatal' — version-incompatible peers are skipped, not counted as failures, so a folder where every peer is incompatible is silently treated as a clean pass. The observable consequence is that a user with an all-incompatible sync folder gets a successful sync report with no peers imported, rather than an error indicating the folder is unusable.

```suggestion
if opened.is_empty() && (failures > 0 || incompatible_count > 0) {
```

<sub>**How this was verified:** Checked the loop: Incompatible arm does not increment failures, only the Err arm does. The test version_skip_plus_open_failure_counts_only_failures explicitly asserts that version-mismatch skips are not counted as failures, confirming the behavior.</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `aw-sync/src/sync.rs:1228`

The test broken_peer_does_not_skip_healthy_sibling uses unreadable_peer() which creates a Datastore whose parent directory does not exist. The comment says Datastore::new succeeds lazily but sync_datastores fails. However, the test then passes this broken Datastore as a remote and expects pull_from_remotes to skip it. The test does not actually verify that the broken peer's failure is recorded on the report (it passes record_peers=false), and it does not verify that the broken peer's sync_datastores actually failed — it only checks that the healthy sibling imported. If unreadable_peer() somehow succeeded (e.g., the temp dir got created by another test), the test would still pass because the healthy sibling imports regardless. The test is not hermetic in the sense that it depends on the parent directory not existing, which could be affected by other tests or environment. This is a test-defect: the test's assertion would pass even if the broken peer was actually readable, so it does not pin the skip behavior.

<sub>**How this was verified:** The test creates broken via unreadable_peer() which uses a unique path with process id and nanos, so the parent dir should not exist. But the test does not assert that the broken peer was skipped; it only asserts the healthy sibling imported. If the broken peer were readable, the test would still pass.</sub>

</details>

**⚠️ P2 medium** — `aw-sync/src/sync.rs`

```suggestion
Add a test file that reproduces the bug before the fix and passes after it.
```

</details>

- `aw-sync/src/sync.rs` — Replaces the per-peer open and pull loops in sync_run with open_peer_datastores() and pull_from_remotes() that skip failing peers, and adds peer_isolation_tests.

</details>

| `17fc4b909cc3` | 4/5 | 1 | llm | 2026-09-16 11:23 UTC |
| `fcaba7f8d7f2` | 5/5 | 0 | llm | 2026-09-16 12:22 UTC |
| `ac424d49d287` | 4/5 | 3 | llm | 2026-09-16 15:16 UTC |
| `14d7f9a963e2` | 5/5 | 0 | llm | 2026-09-16 15:50 UTC |
| `a77559672ec3` | 4/5 | 1 | llm | 2026-09-17 09:41 UTC |
| `9b49d4a3233a` | 3/5 | 3 | llm | 2026-09-17 11:02 UTC |

</details>

<sub>Reviewed `789b13cf28e5` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 27s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

aw-sync/src/sync.rs
```diff
@@ -686,3 +720,75 @@ mod pull_only_staging_tests {
...
+    /// importing. This is the daemon-path #688 abort: `sync_run` used `?` on
+    /// each remote, so the first unreadable db skipped every peer after it.
+    #[test]
+    fn broken_peer_does_not_skip_healthy_sibling() {
+        let healthy = Datastore::new_in_memory(false);
+        healthy
+            .create_bucket(&bucket("aw-watcher-window", "host-ok"))
+            .unwrap();
+        let dest = Datastore::new_in_memory(false);
+        let broken = create_datastore(Path::new(
```
TimeToBuildBob: ⚠️ **P2** — The test broken_peer_does_not_skip_healthy_sibling uses a hardcoded nonexistent path '/nonexistent-directory-for-aw-sync-tests/peer-isolation.db' to create a broken datastore. This test is not hermetic: if a directory with that exact path exists on the test machine (e.g., a developer created it, or a previous test run left it), create_datastore will succeed and the test will fail because the 'broken' peer is actually healthy. More importantly, the test relies on the filesystem state, which violates the principle of hermetic tests. The test should create a temporary directory and then remove it, or use a path that is guaranteed not to exist, to avoid environment-dependent behavior.

aw-sync/src/sync.rs
```diff
@@ -331,6 +334,37 @@ fn is_synced_bucket(bucket: &Bucket) -> bool {
...
+    let mut last_err: Option<String> = None;
+    for ds_from in remotes {
+        attempted += 1;
+        match sync_datastores(*ds_from, dest, false, None, sync_spec) {
+            Ok(()) => succeeded += 1,
+            Err(e) => {
+                warn!("Skipping peer {ds_from:?}: {e}");
+                last_err = Some(e);
+            }
+        }
```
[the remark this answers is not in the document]
TimeToBuildBob: Leaving this. Destination-side total failure is already Err (`attempted > 0 && succeeded == 0`) — the same rule as ActivityWatch/aw-server-rust#697, which Erik signed off on this PR. A dest error that recovers for a later peer is per-peer isolation: that peer retries next daemon cycle. Source-side isolation is the point of this change.

aw-sync/src/sync.rs
```diff
@@ -112,8 +112,9 @@ pub fn sync_run(
         match create_datastore(path) {
             Ok(ds) => ds_remotes.push(ds),
             Err(e) => {
-                warn!("Failed to open remote db {}: {e}", path.display());
-                return Err(e.into());
+                // Per-peer: a single unreadable db must not abort the pass
+                // and skip every peer after it (ActivityWatch/aw-server-rust#688).
+                warn!("Skipping unreadable peer db {}: {e}", path.display());
             }
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in fcaba7f — `open_peer_datastores` now returns Err when every discovered peer fails to open, and keeps the zero-peers no-op. Tests: `all_open_failures_is_err`, `no_discovered_peers_is_ok`, `mixed_open_keeps_the_readable_peer`.

TimeToBuildBob:
Closed the open-failure counting hole in fcaba7f: if every discovered peer fails to open, pull is now Err (zero peers stays a no-op). Isolation tests are hermetic.

Greptile dest-failure P1 stays: that is the #697 total-failure rule you signed off.

Still last in the stack. Not merging until #700 → #678 → #697 are in; then rebase and merge on green.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:

TimeToBuildBob:
Rebased onto master after ActivityWatch/aw-server-rust#700 landed (this was DIRTY).

Folded the #700 residual: `open_peer_datastores` now walks `open_peer_datastore` (read-only, version-mismatch skip). A non-version open failure is skip-and-continue; if every discovered peer fails to open, pull is still Err. Version-mismatch-only stays a no-op.

Isolation tests updated for the read-only open (missing file is a hard fail; mixed-open creates a checkpointed current-version db). Local `peer_isolation_tests` + `pull_only_staging_tests` green.

Still last in the stack — not merging until ActivityWatch/aw-server-rust#697 is in. Order remains #700 → #678 → #697; rebase again if #697 conflicts, then merge on green.

TimeToBuildBob:
@greptileai review

aw-sync/src/sync.rs
```diff
@@ -108,17 +108,13 @@ pub fn sync_run(
...
-            Err(e) => {
-                warn!("Failed to open remote db {}: {e}", path.display());
-                return Err(e.into());
-            }
-        }
-    }
+    // Push uses ds_localremote, not remotes. A single unreadable db must not
+    // abort the pass (ActivityWatch/aw-server-rust#688).
+    let ds_remotes = if mode == SyncMode::Pull || mode == SyncMode::Both {
+        open_peer_datastores(&remote_dbfiles)?
```
TimeToBuildBob: ⚠️ **P2** — In sync_run, when mode is Push, ds_remotes is set to Vec::new() and the pull block is skipped. However, the cleanup loop at lines 144-146 iterates over ds_remotes and calls close() on each. Since ds_remotes is empty for Push, this is fine. But for Pull/Both, if open_peer_datastores returns an error (all peers failed to open), the function returns early via `?` at line 114, and ds_localremote (if Some) is never closed. For Both mode, ds_localremote was created by maybe_setup_local_remote and holds an open sqlite connection. The early return leaks that connection. The same leak occurs if pull_from_remotes returns Err at line 134. The original code also had early returns on open errors, but it did not create ds_localremote for Pull-only; for Both it did, and the original code also leaked on open error. However, this PR changes the error path for open_peer_datastores: previously a single unreadable peer caused an early return, now only all-fail does. The leak on all-fail is still present. The consequence is a leaked database connection/file handle on a failed sync pass, which on repeated failures could exhaust file descriptors. The fix is to close ds_localremote before returning early, or restructure to ensure cleanup.

aw-sync/src/sync.rs
```diff
@@ -355,6 +353,67 @@ fn is_synced_bucket(bucket: &Bucket) -> bool {
...
+            Ok(None) => {}
+            Err(e) => {
+                warn!("Skipping unreadable peer db {}: {e}", path.display());
+                last_err = Some(e);
+            }
+        }
+    }
+    if remotes.is_empty() && last_err.is_some() {
+        return Err(format!(
+            "all {} discovered peers failed to open; last error: {}",
```
TimeToBuildBob: ⚠️ **P2** — The error message in open_peer_datastores reports 'all {paths.len()} discovered peers failed to open', counting every path in the input slice. Peers that return Ok(None) from open_peer_datastore (version mismatch) are skipped without setting last_err, but they are still included in paths.len(). So a configuration with one old-version peer (skipped) and one corrupt peer (failed) yields 'all 2 discovered peers failed to open', even though only one actually failed to open. This misleads users debugging a sync failure by overstating the number of broken peers. The fix is to count only the peers that produced an error (e.g. track a separate failures counter) in the message.

aw-sync/src/sync.rs
```diff
@@ -710,3 +769,159 @@ mod pull_only_staging_tests {
...
+            ds.force_commit().unwrap();
+            ds.close();
+            let conn = rusqlite::Connection::open(&good).unwrap();
+            let _ = conn.execute_batch("PRAGMA wal_checkpoint(TRUNCATE);");
+        }
+        let bad = PathBuf::from(std::ffi::OsString::from_vec(
+            b"/tmp/aw-sync-\xff-mix.db".to_vec(),
+        ));
+        let remotes =
+            open_peer_datastores(&[bad, good.clone()]).expect("partial open failure must be Ok");
```
TimeToBuildBob: ⚠️ **P2** — The test `mixed_open_keeps_the_readable_peer` creates a real sqlite file with a bucket, then opens it read-only via open_peer_datastores. The test relies on the file being a valid current-version sqlite db. It uses `PRAGMA wal_checkpoint(TRUNCATE)` to ensure the WAL is checkpointed. However, the test does not verify that the opened Datastore actually has the bucket; it only checks that remotes.len() == 1. This is a weak assertion: it would pass even if the opened peer is a different file or has no buckets. The test's purpose is to verify that a mixed set (one bad path, one good) keeps the readable peer, and it does verify that the good path is opened (since open_peer_datastores would fail on a non-sqlite file). But it does not verify the content, so it could pass vacuously if open_peer_datastores returned Some for a wrong reason. However, the test does exercise the open path and the assertion that exactly one remote is returned is meaningful. The test is not vacuous because open_peer_datastores would return Err for the bad path and Ok(Some) for the good path only if the file is a valid db. The test is acceptable.

aw-sync/src/sync.rs
TimeToBuildBob: Rejected — this is not a leaked connection.

`Datastore` has no RAII `close()` requirement on every return: dropping the last clone drops the request sender, the worker's `poll()` errors, and the thread exits, which drops the sqlite `Connection`. The file documents that path at the existing cleanup:

```
// Dropping also works to close the database connections, weirdly enough.
// Probably because once the database is dropped, the thread will stop,
// and then the Connection will be dropped, which closes the connection.
```

Worker confirmation is `aw-datastore/src/worker.rs` (`Err` on `poll` → `quit = true` → `DB Worker thread finished`). An early `?` from `open_peer_datastores` or `pull_from_remotes` therefore closes `ds_localremote` the same way as the explicit `close()` + `drop` at the bottom of `sync_run`.

Also: Pull-only never creates `ds_localremote` (`maybe_setup_local_remote` returns `None`). Both-mode early return is the same shape as master; this PR only errors when *every* discovered peer fails to open, so it fires less often than the old per-peer `?`. The suggested patch is a comment change and does not close anything.

The "How this was verified" step is also wrong: a single unreadable peer is now skip-and-continue, not `Err`. Total-failure `Err` is intentional (`ActivityWatch/aw-server-rust#688` / `#697` rule).

aw-sync/src/sync.rs
TimeToBuildBob: Rejected — the finding already concludes the test is acceptable.

`mixed_open_keeps_the_readable_peer` is an isolation check: one unreadable path + one current-version sqlite file must return exactly one remote. `open_peer_datastores` only returns `Ok(Some)` for a successful `Datastore::open_read_only`; a non-sqlite / missing file is `Err` and a version mismatch is `Ok(None)`. `remotes.len() == 1` is therefore the contract under test, not a vacuous content assertion.

Asserting bucket contents would couple this helper test to import/`sync_datastores`, which is already covered by `broken_peer_does_not_skip_healthy_sibling`. Leaving the assertion as-is.

aw-sync/src/sync.rs
TimeToBuildBob: Fixed in 14d7f9a — `open_peer_datastores` now counts hard `Err` opens only. Version-mismatch `Ok(None)` skips no longer inflate `paths.len()`. Test: `version_skip_plus_open_failure_counts_only_failures`.

TimeToBuildBob:
Addressed the in-band AI review P2s on `ac424d4`:

- Counted only hard open failures in the all-fail error (`14d7f9a`). Version-mismatch skips no longer inflate `paths.len()`.
- Rejected the `ds_localremote` leak: Drop already closes the worker (documented in `sync_run`; `poll` Err → quit).
- Rejected the `mixed_open_keeps_the_readable_peer` assertion nit — `remotes.len() == 1` is the contract.

Local in-band review of `14d7f9a` is 5/5. Still last in the stack; not merging until ActivityWatch/aw-server-rust#697 is in.

ErikBjare:
@TimeToBuildBob resolve conflicts

ErikBjare:

ErikBjare:
@TimeToBuildBob #697 just merged, so the rebase target is now `d1a61b1`, not `3d3b9dd`. This is the last sync PR in the v0.14.0 sequence — master will not move again before it lands (#702 and #706 don't touch `sync_run`), so one rebase onto `d1a61b1` is final. After it: verify the open-failure skip and the `attempted > 0 && succeeded == 0 → Err` rule still compose with #697's identical rule in `sync_datastores`/`pull_all` — you don't want the same total-failure check twice in one pass.

ErikBjare:
@TimeToBuildBob `eecc50f` landed on `3d3b9dd`, not current master — #697 merged as `d1a61b1` in between, so this is still `CONFLICTING` (diff vs master is +376/−635, the deletions being #697's changes absent from the branch). One more rebase, onto `d1a61b1`; nothing else will move.

While rebasing, resolve the duplication rather than carrying it: this PR adds `attempted > 0 && succeeded == 0 → Err` inside a new `pull_from_remotes`, and #697 already landed the same rule in `pull_all` (`sync_wrapper.rs`) and `sync_datastores`. Keep **one** total-failure check per pass — the natural home is wherever the peer loop ends up after the rebase — so a pass with zero successes returns `Err` exactly once and a partial pass never does. Content otherwise verified: unreadable-peer skip and per-remote skip are both present.

ErikBjare:
Verified on `a775596`: on `d1a61b1`, one commit, `MERGEABLE`; version-mismatch skip (`:299`), unreadable-peer skip (`:337`) and per-remote skip (`:389`) all present.

**Retracting the duplication flag** — the three `attempted > 0 && succeeded == 0` checks are three *scopes* that nest correctly, not one rule repeated:
- `sync_datastores` — bucket scope (all buckets of one peer failed)
- `pull_from_remotes` — remotes within one `sync_run`; on the `pull_all` path each run sees exactly one remote
- `pull_all` — hosts scope; treats the inner `Err` as one failed peer and only errors when every host failed

Net behaviour: partial failure never returns `Err`, total failure returns `Err` exactly once, and the advanced `sync_run`-against-root path (which never enters `pull_all`) keeps its own guard. That is the #688 contract. Nothing further — ready for Erik on green. Last sync PR in the v0.14.0 sequence.

TimeToBuildBob:
Rebased onto current master (`d1a61b1`, after #697/#699/#678 landed) as a single squashed commit (`a775596`). The #688 semantics are re-expressed in the post-#699 report world: per-peer open failures and pull failures are recorded on the report (`PeerReport::failed`) and skipped; only total failure — every discovered peer failed to open, or every pull failed — is `Err`, so the daemon still fails loud on a broken folder without aborting the pass on one bad peer. Peer isolation tests updated and passing locally (43/43 in `aw-sync --lib`). CI is running on the new head.

aw-sync/src/sync.rs
```diff
@@ -1083,3 +1159,224 @@ mod hostname_sanitize_tests {
...
+            SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap()
+                .as_nanos()
+        ));
+        fs::create_dir_all(&dir).unwrap();
+        let old = dir.join("old.db");
+        {
+            let conn = rusqlite::Connection::open(&old).unwrap();
+            conn.pragma_update(None, "user_version", 4).unwrap();
```
TimeToBuildBob: ⚠️ **P2** — The test version_skip_plus_open_failure_counts_only_failures creates a peer database with `user_version` set to 4 and expects open_peer_datastores to treat it as incompatible (OpenedPeer::Incompatible) and therefore return Err because no peer is usable. If the binary's current schema version ever becomes 4, open_peer_datastore will return OpenedPeer::Ready instead, making opened non-empty, so open_peer_datastores returns Ok and the .expect_err() panic fires. The test is not hermetic: it depends on a specific user_version value remaining outside the supported range. It should either use a version that is guaranteed to be incompatible (e.g., 0 or a very large number) or read the current schema version and pick a clearly different one.

TimeToBuildBob: Fixed in 9b49d4a — the test now sets `user_version` to `NEWEST_DB_VERSION + 1000` (imported from `aw_datastore`) instead of the hardcoded 4, so it stays incompatible across any realistic schema bump. `cargo test -p aw-sync --lib`: 43/43 pass.

ErikBjare:
`9b49d4a` verified: test-only — the version-skip test now uses `NEWEST_DB_VERSION + 1000` instead of a hardcoded `4` that would go stale with the next schema bump. Three skips and the three-scope total-failure rules unchanged from the verified `a775596`. Still `MERGEABLE` against post-#706 master; no further rebase. Merge on green — last sync PR.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Merge attempted via the gated self-merge path — blocked: the bot token has pull-only access to this repo (collaborator-permission probe confirms 403). Everything else is green: CI 6/6 success, Greptile 5/5, all review threads resolved except our own P2 which is fixed in 9b49d4a (test-only, NEWEST_DB_VERSION + 1000) with the thread reply posted. Needs your merge click, Erik.

aw-sync/src/sync.rs
```diff
@@ -332,6 +306,108 @@ fn open_peer_datastore(path: &Path) -> Result<OpenedPeer, String> {
...
+                        db.device_id.clone(),
+                        db.hostname.clone(),
+                        db.path.clone(),
+                        e.clone(),
+                    ));
+                }
+            }
+        }
+    }
+    if opened.is_empty() && failures > 0 {
```
TimeToBuildBob, line 353: ❌ **P1** — In open_peer_datastores(), the total-failure condition is `opened.is_empty() && failures > 0`. If every discovered peer is version-incompatible (OpenedPeer::Incompatible), failures stays 0 and opened is empty, so the function returns Ok(empty). The caller then proceeds to pull_from_remotes with an empty remotes list, which returns Ok(()) because attempted == 0. The pass is reported as a clean empty success even though every peer was skipped as incompatible. The comment at lines 116-119 says an all-incompatible pass must be visible and not presented as a clean empty one, but this code path does exactly that. The version_skip_plus_open_failure_counts_only_failures test only covers the case where there is also a hard failure, so it does not catch this. The consequence is that a user with only incompatible peers sees no error and no skipped-peer report (when record_peers is false, e.g. push-only mode), making the failure invisible.

aw-sync/src/sync.rs
```diff
@@ -117,35 +117,21 @@ pub fn sync_run(
...
-                        db.path.clone(),
-                        e.clone(),
-                    ));
-                    report.finish();
-                    crate::report::persist_last_report_warn(&report);
-                }
-                close_opened_datastores(&ds_remotes, &ds_localremote);
-                return Err(e.into());
-            }
+    match open_peer_datastores(
```
TimeToBuildBob: ⚠️ **P2** — In sync_run, after open_peer_datastores returns Ok(opened) with an empty vector (because all peers were version-incompatible), the code proceeds to the pull block. pull_from_remotes is called with an empty remotes vector, which returns Ok(()). The report then finishes and is returned as Ok. The report has no skipped entries because record_peers was true for pull mode, and the Incompatible arm does push PeerReport::skipped. Wait, in pull mode record_peers is true, so the skipped entries are recorded. So the report does show the skipped peers. The issue is only when record_peers is false (push-only mode), but in push-only mode the pull block is not executed, and the open loop still runs. In push-only mode, if all peers are incompatible, open_peer_datastores returns Ok(empty) and the pass continues to push, reporting success with no mention of the incompatible peers. That is a real visibility bug, but it is the same as the first finding. I will keep the first finding and not duplicate.

aw-sync/src/sync.rs
TimeToBuildBob: Fixed in 9b49d4a — the version-skip test now sets `user_version` to `NEWEST_DB_VERSION + 1000` (imported from `aw_datastore`) instead of the hardcoded 4, so it stays incompatible across any realistic schema bump. `cargo test -p aw-sync --lib`: 43/43 pass.

aw-sync/src/sync.rs
TimeToBuildBob: Fixed in 789b13c — `open_peer_datastores` is now called only for Pull/Both passes. In Push mode peers are never opened at all (push only uses the local staging datastore). Since the call now only ever runs where `record_peers` is `true`, an all-incompatible pass is always recorded as skipped peers on the report — the invisible path you traced existed only in the push-mode open, which is gone. Pull-mode all-incompatible remains a no-op pass *with* skipped entries on the report, per the verified design (version-mismatch-only stays non-fatal).

aw-sync/src/sync.rs
TimeToBuildBob: Duplicate of the P1 on `open_peer_datastores` (fp 21ddcd91524c) — the finding itself concludes "I will keep the first finding and not duplicate." The root cause is fixed in 789b13c: push-mode passes no longer open peer datastores, and peer opens now always run with `record_peers=true`, so incompatible peers are always recorded.

aw-sync/src/sync.rs
```diff
@@ -117,35 +117,21 @@ pub fn sync_run(
...
+        &selection.selected,
+        &mut report,
+        mode == SyncMode::Pull || mode == SyncMode::Both,
+    ) {
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 789b13c — `open_peer_datastores` is now called only when the pass pulls (`SyncMode::Pull`/`Both`). A Push-only pass never opens peer databases (it only uses the local staging datastore), so unreadable or incompatible peers can no longer abort or taint a push. Peer opens therefore only ever run with peer reporting enabled, which also keeps an all-incompatible pull pass visible on the report. `cargo test -p aw-sync --lib`: 43/43 pass.

ErikBjare:
`789b13c` verified: Push-only passes no longer open peer datastores — the mirror of #698 (pull-only never stages), and correct: a push only touches the local staging db, so an unreadable or incompatible peer must not be able to abort or taint it. Three skips and the three-scope total-failure rules unchanged. `MERGEABLE` against current master. Merge on green — this is the last sync PR, and it needs Erik's click (pull-only token, as your 11:04 probe found).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
The fresh review on `9b49d4a` surfaced one real P1 (push-only passes were opening peer datastores and could abort before pushing anything) — fixed in `789b13c`: `open_peer_datastores` now runs only for Pull/Both passes; push mode never touches peer databases. The in-band review P1 (invisible all-incompatible pass) is resolved by the same gate, its P2 was a self-declared duplicate, and the earlier hermetic-test P2 was already fixed in `9b49d4a`. All four threads closed; 43/43 lib tests pass. Re-review triggered on the new head.

Still needs your merge click when green — the bot token remains pull-only on this repo.

aw-sync/src/sync.rs
```diff
@@ -332,6 +308,108 @@ fn open_peer_datastore(path: &Path) -> Result<OpenedPeer, String> {
...
+                        db.device_id.clone(),
+                        db.hostname.clone(),
+                        db.path.clone(),
+                        e.clone(),
+                    ));
+                }
+            }
+        }
+    }
+    if opened.is_empty() && failures > 0 {
```
TimeToBuildBob, line 353: ⚠️ **P2** — In open_peer_datastores(), the condition `if opened.is_empty() && failures > 0` treats a pass where every discovered peer is version-incompatible (OpenedPeer::Incompatible) as success, because failures only counts hard open errors. The PR description and comments say a total open failure should be Err so a broken folder is not reported as a clean empty pass, but an all-incompatible folder is reported as Ok with an empty ds_remotes, and sync_run then proceeds to push (in Both mode) and returns Ok. This contradicts the stated intent in the comment on lines 120-122 and the PR description's 'only a total open failure is fatal' — version-incompatible peers are skipped, not counted as failures, so a folder where every peer is incompatible is silently treated as a clean pass. The observable consequence is that a user with an all-incompatible sync folder gets a successful sync report with no peers imported, rather than an error indicating the folder is unusable.

aw-sync/src/sync.rs
```diff
@@ -1083,3 +1161,228 @@ mod hostname_sanitize_tests {
...
+            .unwrap();
+        let dest = Datastore::new_in_memory(false);
+        let broken = unreadable_peer("sibling");
+
+        let db_a = peer_db("dev-a", "host-broken", PathBuf::from("/nonexistent-a.db"));
+        let db_b = peer_db("dev-b", "host-ok", PathBuf::from("/nonexistent-b.db"));
+        let remotes: Vec<(&RemoteDb, &dyn AccessMethod)> =
+            vec![(&db_a, &broken), (&db_b, &healthy)];
+        let mut report = dummy_report();
+        pull_from_remotes(&remotes, &dest, &SyncSpec::default(), &mut report, false)
```
TimeToBuildBob, line 1228: ⚠️ **P2** — The test broken_peer_does_not_skip_healthy_sibling uses unreadable_peer() which creates a Datastore whose parent directory does not exist. The comment says Datastore::new succeeds lazily but sync_datastores fails. However, the test then passes this broken Datastore as a remote and expects pull_from_remotes to skip it. The test does not actually verify that the broken peer's failure is recorded on the report (it passes record_peers=false), and it does not verify that the broken peer's sync_datastores actually failed — it only checks that the healthy sibling imported. If unreadable_peer() somehow succeeded (e.g., the temp dir got created by another test), the test would still pass because the healthy sibling imports regardless. The test is not hermetic in the sense that it depends on the parent directory not existing, which could be affected by other tests or environment. This is a test-defect: the test's assertion would pass even if the broken peer was actually readable, so it does not pin the skip behavior.

ErikBjare:
@TimeToBuildBob address in follow-up if needed

aw-sync/src/sync.rs
TimeToBuildBob: Deliberate, documented tradeoff — and it is visible, not silent. Version-incompatible peers are skipped non-fatally by design (ActivityWatch/aw-server-rust#700: never migrate a foreign db), and `open_peer_datastores` pushes a `PeerReport::skipped(..., "incompatible database version: ...")` for each one, so `aw-sync status` and the JNI report surface them rather than presenting a clean empty pass. `failures` intentionally counts hard `Err` opens only, so a mixed folder (some version-skipped, some readable) is not Err. Only a folder where every discovered peer fails a hard open is fatal. A version-mismatch-only folder being `Ok` with the peers recorded as skipped is the contract Erik signed off on for this cut.

aw-sync/src/sync.rs
TimeToBuildBob: Valid test-quality point, accepted as follow-up. The assertion currently only pins that the healthy sibling imported; it does not pin that the broken peer was recorded as skipped/failed, and `unreadable_peer()` depends on a parent directory that does not exist. Folded into the first post-cut aw-sync PR (the sync queue is held until the v0.14.0 cut, per ActivityWatch/activitywatch#1445): strengthen it to assert the failed `PeerReport` entry and use a deterministic unreadable path, or build the peer info whose db path is a directory. Tracked in the local aw-sync follow-up task; no code change before the cut.

TimeToBuildBob:
Disposition of the three threads left open when this merged — thanks for the click.

**1. Greptile P1 — "Push depends on peer reads" (fixed, thread just stale).** Fixed in `789b13c`: `open_peer_datastores` now runs only for `Pull`/`Both`, so a push-only pass never opens peer databases. The thread is outdated; resolving it.

**2. `c703df198fee` (P2, line 353) — "all-incompatible folder reported as a clean pass" → dismissed.** This reads the intent backwards. Version-incompatible peers are deliberately *skipped, not fatal* — `sync.rs:114-123` says so explicitly, and the #688 lesson is that routing a per-peer condition into a non-zero exit is what burns aw-qt/aw-tauri's restart budget and then disables sync. An all-incompatible folder is a benign steady state (the peer is on a different schema version), not a broken one, so it must not be `Err`. The visibility requirement is met a different way: the skipped peers are recorded on the report (`PeerReport::skipped` with the reason), which is what `aw-sync status` renders. Making this `Err` would re-open #688.

**3. `557933a72bea` (P2, line 1228) — test does not pin the recording path → valid, deferred.** Correct in substance: all eight test call sites pass `record_peers=false`, so `PeerReport::failed`/`skipped` are never asserted. The skip *behaviour* is pinned; the *visibility* half of the contract is not. It's test-only, so rather than open a PR against the current layout it's filed as a local task (`aw-sync-688-report-visibility-test-coverage`) to pick up after the v0.14.0 cut, per Erik's "queue waits for the cut" on ActivityWatch/activitywatch#1445.

aw-sync/src/sync.rs
TimeToBuildBob: Dismissed as a false positive on intent — see the disposition comment above. Version-incompatible peers are deliberately non-fatal (`sync.rs:114-123`); the visibility requirement is met by recording `PeerReport::skipped` on the report. Making an all-incompatible pass `Err` would re-open #688's restart-budget burn.

aw-sync/src/sync.rs
TimeToBuildBob: Valid — all eight test call sites pass `record_peers=false`, so the recording path is never asserted. Deferred as local task `aw-sync-688-report-visibility-test-coverage` (test-only, waits for the v0.14.0 cut per ActivityWatch/activitywatch#1445).

PR #708, opened 2026-09-17
fix(aw-sync): scan pre-#697 buckets by origin to avoid re-import after Android hostname migration

### LINKED ISSUE #707: aw-sync: import lookup cannot match a pre-#697 bucket once the peer's hostname is migrated (aw-android#273) — forks and re-imports
Follow-up to #697, surfaced by ActivityWatch/aw-android#273 (the one-shot that rewrites the phone's `buckets.hostname` to the sanitized form and repairs the forked sync folders).

## The gap

`get_or_create_sync_bucket` on master resolves the destination in two lookups:

1. `get_bucket(new_id)` — derived from the peer bucket's *current* `hostname` (or `$aw.sync.origin` if present)
2. `get_bucket(sanitized_id)` — the same with `sanitize_hostname` applied

That closes the fork for any desktop that first imports the peer **after** #697: it creates `…-synced-from-poco_f8_ultra`, and Android's migration lands on that ID.

It does not help a desktop that imported the peer **before** #697. That desktop holds `aw-watcher-android-synced-from-POCO F8 Ultra` (raw hostname of the day). Once ActivityWatch/aw-android#273 runs on the phone, the peer's staging buckets carry `hostname = poco_f8_ultra` and no `$aw.sync.origin` (first-hand buckets are never stamped), so:

- `new_id` = `…-synced-from-poco_f8_ultra` → not found
- `sanitized_id` = the same → not found
- a **new** bucket is created, `resume_sync_at` is `None`, the whole history is re-imported → every event twice in `/timeline` (the [discussions#1373](https://github.com/orgs/ActivityWatch/discussions/1373) symptom), and the old bucket is orphaned.

Population: desktops (python or rust — the logic is in aw-sync) that pulled an Android peer between the sanitizer shipping in aw-android v0.14.0 and #697 (2026-09-17), and then receive a migrated push. Small, but it is exactly the class #697 was written to prevent, and #273 makes it reachable.

## Fix

#697 stamps `$aw.sync.origin` with the **raw** hostname on import, so the pre-existing bucket is findable: when both lookups miss, scan the destination's `-synced-from-` buckets for one whose base ID matches and whose `sanitize($aw.sync.origin) == sanitized_hostname`, and reuse it. Optionally rename it to the sanitized ID at that point so the display converges. Two tests: a pre-#697-shaped destination (`…-synced-from-POCO F8 Ultra` with raw origin stamped) receiving a migrated peer must resume, not re-import; and two distinct raw origins that sanitize identically must still be refused/warned, not merged (the #697 `:368` note).

All of this dissolves under the `(device_id, id)` identity work in ActivityWatch/activitywatch#302, which is why it is a bounded fix here rather than a redesign.

cc @TimeToBuildBob

TimeToBuildBob:
Closes #707.

## Summary

When both the raw-ID and sanitized-ID lookups miss in `get_or_create_sync_bucket`, add a fallback scan of the destination's `-synced-from-` buckets for one whose base ID matches and whose `$aw.sync.origin` sanitizes to the same value as the current pull target.

This recovers the case described in #707:
- Desktop imported Android peer **before** #697 → holds `aw-watcher-android-synced-from-POCO F8 Ultra` with `$aw.sync.origin = "POCO F8 Ultra"` (stamped by #697 on import).
- Phone runs ActivityWatch/aw-android#273 → staging hostname becomes `poco_f8_ultra` (sanitized); first-hand buckets carry no `$aw.sync.origin`.
- Next pull derives both `new_id` and `sanitized_id` as `…-synced-from-poco_f8_ultra`, both miss, a new bucket is created → full re-import (every event twice in `/timeline`, old bucket orphaned).

With this fix the fallback scan finds the legacy bucket via its stamped `$aw.sync.origin` and resumes from it.

**Ambiguity guard**: if two distinct pre-#697 buckets share the same sanitized origin, the function returns an error rather than silently merging distinct histories (the #697 :368 note).

## Tests added

- `test_pre697_origin_scan_resumes_legacy_bucket`: post-migration phone pull against a desktop with a raw legacy bucket — must resume, not fork.
- `test_pre697_origin_scan_refuses_ambiguous_candidates`: two legacy buckets whose raw origins both sanitize to the same value — must error, not merge.

## Scope

Dissolves entirely under the `(device_id, id)` identity work in ActivityWatch/activitywatch#302 as noted in the issue. This is a bounded compat shim, not a redesign.

Co-Authored-By: Bob <noreply@timetobuildbob.com>


TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 2 findings disposed (accepted-tradeoff; rejected)

| --- | --- | --- |
| `aw-sync/src/sync.rs:577` | P1 | accepted-tradeoff |
| `aw-sync/src/sync.rs:582` | P1 | rejected — Reviewer assumes origin stamping was introduced in #697, but Greptile's own review confirmed '.sync.origin has been stam |
| `aw-sync/src/sync.rs:577` | P2 | accepted-tradeoff |
| `aw-sync/tests/sync.rs:339` | P2 | fixed in `1d3cc39` |
| `aw-sync/tests/sync.rs:468` | P2 | fixed in `a90bed0` |
| `aw-sync/tests/sync.rs:362` | P2 | accepted-tradeoff |

</details>




Adds a fallback scan in get_or_create_sync_bucket for pull operations: when both the raw-ID and sanitized-ID lookups miss, it lists all destination buckets, filters those whose base ID (before -synced-from-) matches the source's base ID and whose $aw.sync.origin sanitizes to the same value as the source's origin, and reuses the single match or errors on multiple matches. Adds two integration tests covering the resume and ambiguity-refusal paths.

Confidence 3/5

<details><summary><b>2 findings</b> · ❌ 1 P1 · ⚠️ 1 P2</summary>

**❌ P1 high** — `aw-sync/src/sync.rs:582`

The fallback scan matches destination buckets by comparing b_base (the part of the destination bucket ID before '-synced-from-') to target_base (the source bucket's base ID). But the source bucket's base ID is derived by splitting bucket_from.id on '-synced-from-', which for a first-hand bucket like 'aw-watcher-android' yields 'aw-watcher-android'. The destination legacy bucket 'aw-watcher-android-synced-from-POCO F8 Ultra' also yields 'aw-watcher-android', so the base match works. However, the scan also requires the destination bucket's $aw.sync.origin to be present and sanitize to target_sanitized. The target_sanitized is computed from sync_origin (the source's $aw.sync.origin) or bucket_from.hostname. In the post-migration scenario, the source bucket has no $aw.sync.origin and hostname 'poco_f8_ultra', so target_sanitized is 'poco_f8_ultra'. The legacy destination bucket has $aw.sync.origin 'POCO F8 Ultra', which sanitizes to 'poco_f8_ultra', so it matches. This is correct for the intended case. But consider a pre-#697 destination bucket that was created by an older import that did NOT stamp $aw.sync.origin (the comment says #697 stamps it, but the fallback is for pre-#697 buckets; if the import predates the stamping, the bucket has no $aw.sync.origin). The filter requires $aw.sync.origin to be present, so such a bucket would not match, and the fallback would create a new sanitized bucket, causing the re-import the PR aims to avoid. The PR description says pre-#697 buckets hold $aw.sync.origin stamped by #697, but #697 is the change that added the stamping; a bucket imported before #697 would not have it. The comment in the code says 'with $aw.sync.origin set to that same raw value by #697's import stamp', which is contradictory: pre-#697 means before the stamping existed. If the legacy bucket lacks $aw.sync.origin, the fallback fails and the re-import still happens. This is a correctness gap in the fallback's matching criterion.

<sub>**How this was verified:** Checked the filter at lines 576-587: it requires $aw.sync.origin to be present via .and_then(|v| v.as_str()) and .unwrap_or(false). The PR description and code comment claim pre-#697 buckets have $aw.sync.origin, but #697 is the change that introduced the stamping, so pre-#697 buckets would not have it. The existing test test_whitespace_hostname_pull_reuses_legacy_unsanitized_bucket creates a legacy bucket without $aw.sync.origin and relies on the raw-ID lookup, not the fallback. The new fallback would not find such a bucket.</sub>

**⚠️ P2 medium** — `aw-sync/tests/sync.rs:362`

The test test_pre697_origin_scan_resumes_legacy_bucket seeds the legacy destination bucket with an event at T0 and then inserts two source events, one before T0 and one after T0. The sync is expected to resume from the destination's newest event (T0+1s) and import only the after-T0 event. However, the test does not verify that the before-T0 event is NOT imported. The assertion checks event_count == 2, which is the existing event plus the after-T0 event. If the sync incorrectly re-imported everything from scratch, the count would be 3 (existing + before + after), so the test would fail. But if the sync imported only the before-T0 event and skipped the after-T0 event (e.g., due to a cursor bug), the count would be 2 as well, and the test would pass vacuously. The test should also assert that the destination does not contain the before-T0 event's data (e.g., by checking get_events and verifying only the 'existing' and 'new' events are present). This is a test defect: the assertion is not strong enough to distinguish the correct resume behavior from a wrong cursor that imports the older event but misses the newer one.

<sub>**How this was verified:** Read the test: it inserts before_t0 and after_t0 into source, and the destination already has an event at T0. The assertion only checks the total count is 2. A bug that imports before_t0 but not after_t0 would also yield count 2 (existing + before_t0). The test does not inspect event contents.</sub>

</details>

<details><summary><b>Files changed (2)</b> — the diff as I read it</summary>

- `aw-sync/src/sync.rs` — Adds the pre-#697 origin-based fallback scan in get_or_create_sync_bucket, gated on !is_push, with single-match reuse and multi-match error.
- `aw-sync/tests/sync.rs` — Adds test_pre697_origin_scan_resumes_legacy_bucket and test_pre697_origin_scan_refuses_ambiguous_candidates.

</details>

| `749d83add41b` | 4/5 | 2 | llm | 2026-09-17 14:56 UTC |
| `1d3cc39ad16f` | 3/5 | 2 | llm | 2026-09-17 18:08 UTC |

</details>

<sub>Reviewed `a90bed070fb3` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 47s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

aw-sync/src/sync.rs
```diff
@@ -547,6 +547,69 @@ fn get_or_create_sync_bucket(
...
+                .as_deref()
+                .unwrap_or(bucket_from.hostname.as_str()),
+        );
+        let all_dest = ds_to
+            .get_buckets()
+            .map_err(|e| format!("Failed to list dest buckets for origin scan: {e:?}"))?;
+        let mut candidates: Vec<Bucket> = all_dest
+            .into_values()
+            .filter(|b| {
+                let b_base = b.id.split("-synced-from-").next().unwrap_or(&b.id);
```
TimeToBuildBob, line 577: ⚠️ **P2** — The fallback scan matches destination buckets by comparing `b_base` (the part of the destination bucket ID before `-synced-from-`) to `target_base` (the part of the source bucket ID before `-synced-from-`). However, the source bucket in the pull path is a first-hand bucket (e.g. `aw-watcher-android`), so `target_base` is the full source ID. The destination legacy bucket is `aw-watcher-android-synced-from-POCO F8 Ultra`, whose `b_base` is `aw-watcher-android`. These match, so the scan works for the intended case. But consider a source bucket whose own ID contains `-synced-from-` (e.g. a bucket named `foo-synced-from-bar` that is first-hand). The code splits it and takes `foo` as `target_base`, while the destination bucket for that source would be `foo-synced-from-bar-synced-from-<origin>`, whose `b_base` is `foo-synced-from-bar`. These do not match, so the fallback never finds the legacy bucket and a new fork is created. This is a pre-existing ID-grammar limitation (the code already splits on `-synced-from-` elsewhere), but the new scan adds a new failure mode: it silently falls through to create a new bucket, causing a full re-import, exactly the problem the PR aims to fix. The impact is a re-import for any bucket whose base ID contains the reserved token, which is a real but narrow edge case.

aw-sync/tests/sync.rs
```diff
@@ -261,6 +261,138 @@ mod sync_tests {
...
+            !dest_buckets.contains_key(forked_id),
+            "a sanitized fork must not be created; got: {:?}",
+            dest_buckets.keys().collect::<Vec<_>>()
+        );
+        // Events were imported into the legacy bucket, not lost.
+        let event_count = state
+            .ds_dest
+            .get_event_count(legacy_id, None, None)
+            .unwrap();
+        assert!(event_count > 0, "legacy bucket must have received events");
```
TimeToBuildBob: ⚠️ **P2** — The test `test_pre697_origin_scan_resumes_legacy_bucket` asserts that the legacy bucket is reused and no sanitized fork is created, but it does not verify that the events were actually synced into the legacy bucket with the correct resume behavior. It only checks `event_count > 0`. The test would pass even if the sync created a new bucket and copied events there, as long as the legacy bucket already had events. However, the test sets up the legacy bucket with no events, so `event_count > 0` does prove events were added. The test is not vacuous. The ambiguous test correctly expects an error. Both tests are well-formed and cover the new paths.

ErikBjare:
@TimeToBuildBob Logic is right and the key is complete: `$aw.sync.origin` has been stamped on pull since `a4588f7` (2021), long before Android's sanitizer shipped, so every legacy bucket in the affected window carries it — no `sanitize(b.hostname)` fallback needed. Scan only runs after both exact lookups miss (first import of a bucket), so listing the destination is cheap.

Two test tightenings before merge, then LGTM:

1. **Prove resume, not just import.** `event_count > 0` proves the events landed in the legacy bucket; #707's actual requirement is *no re-import*. Seed the legacy bucket with one event at `T0`, put two in the source (`T0 - 1h`, `T0 + 1h`), sync, assert the legacy bucket ends with **2** events, not 3. That pins `sync_one` reading its cursor from the reused bucket (`:946`).

2. **Ambiguity is a skip, not an abort — say so.** `get_or_create_sync_bucket`'s `Err` is caught at `:772` as warn+continue; `sync_datastores` only returns `Err` because your fixture has a single bucket and trips the total-failure rule. The docstring reads as if the whole pull fails. Add a second, healthy source bucket to that fixture and assert it *did* sync while the ambiguous one was skipped — that is the behaviour we want (one unresolvable bucket must not stop a peer's sync), and it stops the test passing for the wrong reason.

On your reviewer's findings: P2 #1 (`-synced-from-` inside a first-hand base ID) is the pre-existing ID grammar, same split at `:472` — out of scope here, dissolves with ActivityWatch/activitywatch#302. The advisory is right that this adds no new trust boundary; the direct lookup already trusts the stamp.

Nit, optional: `target_sanitized` is the suffix you already computed for `sanitized_id`; deriving one from the other keeps the two from drifting.

TimeToBuildBob:
Done in 1d3cc39:

1. **Resume proof** — seeded legacy bucket with one event at T0; source has T0-1h and T0+1h. Assert exactly 2 events after sync (3 = cursor ignored, full re-import).

2. **Ambiguity is skip, not abort** — added healthy `aw-watcher-window` source bucket. `sync_datastores` now returns Ok; assert the healthy bucket received events and neither ambiguous legacy bucket was written to.

Skipped the optional `target_sanitized` nit — the two computations are equivalent in the pull path and the current form reads clearly. Will revisit in ActivityWatch/activitywatch#302.

aw-sync/src/sync.rs
```diff
@@ -547,6 +547,69 @@ fn get_or_create_sync_bucket(
...
+                if b_base != target_base {
+                    return false;
+                }
+                // $aw.sync.origin must be present and sanitize to the same value.
+                b.data
+                    .get("$aw.sync.origin")
+                    .and_then(|v| v.as_str())
+                    .map(|s| sanitize_hostname(s) == target_sanitized)
+                    .unwrap_or(false)
```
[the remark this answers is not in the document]
TimeToBuildBob: **Disposition: accepted limitation, out of scope here — tracked by ActivityWatch/activitywatch#302.**

The mechanism you describe is real, but it is not introduced by this fallback, and it cannot be closed without device identity:

1. **The direct sanitized-ID lookup already has this collision.** After ActivityWatch/aw-server-rust#697, two devices whose hostnames sanitize to the same value both derive `…-synced-from-<sanitized>`; the second one hits the existing bucket and merges. The pre-#697 scan only lets *legacy raw-ID* buckets participate in that same equivalence class — which is exactly what ActivityWatch/aw-server-rust#707 needs.
2. **Refusing the sole-candidate case restores a certain corruption.** Falling through to `create_bucket` re-imports the peer's entire history (every event duplicated in `/timeline`, old bucket orphaned) for *every* peer migrated by ActivityWatch/aw-android#273. That is the common case; a sanitize collision across two distinct devices is the rare one.
3. **The fix is the `(device_id, id)` identity work**, which the PR body already names as the dissolution path. A correct guard needs a device identity stamped at import time; pre-#697 buckets do not carry one, and neither the ID nor `$aw.sync.origin` can distinguish "same device renamed" from "different device with a colliding sanitized name" — both present as a raw origin that sanitizes to the same value.

Erik reviewed the design and gave LGTM conditional on two test tightenings ([#issuecomment-5716537546](https://github.com/ActivityWatch/aw-server-rust/pull/708#issuecomment-5716537546)); both landed in `1d3cc39`.

Per the review-convergence discipline, no re-trigger: no code was pushed for this, so the score is unchanged. Resolving this thread as an accepted, documented limitation rather than leaving a stale finding — if you disagree with the disposition, reopen and I'll narrow the fallback instead.

aw-sync/src/sync.rs
TimeToBuildBob: **Disposition: accepted limitation, out of scope here — no code change.**

You are right that a *first-hand* bucket whose own ID contains the reserved `-synced-from-` token gets truncated by the split, so the fallback's base comparison would not match and it would fall through to a new fork. That limitation is pre-existing ID grammar: the same split already runs at `aw-sync/src/sync.rs:472` in `get_or_create_sync_bucket`, before this scan exists. The new scan inherits the grammar rather than introducing it.

Erik adjudicated exactly this point on the PR: pre-existing ID grammar, same split, out of scope here, and it dissolves entirely under the `(device_id, id)` identity work in ActivityWatch/activitywatch#302.

Closing this thread as an accepted limitation tracked by ActivityWatch/activitywatch#302.

aw-sync/tests/sync.rs
TimeToBuildBob: **Disposition: fixed in 1d3cc39 (commit `1d3cc39`).**

The resume behaviour is now pinned exactly as requested, so an empty-destination resume can no longer pass this test:

- The legacy destination bucket is seeded with one event at `T0`.
- The source holds two events: `T0 - 1h` (already covered) and `T0 + 1h` (new).
- After the pull, the legacy bucket must hold **exactly 2** events — `3` would mean the reused bucket's cursor was ignored and the whole history was re-imported.

That replaces the earlier `event_count > 0` check, which could not distinguish resume from re-import.

The companion ambiguity test also gained a second, healthy source bucket and now asserts `sync_datastores` returns `Ok` with the healthy bucket synced while the ambiguous one is warn+skipped — so it no longer passes by tripping the total-failure rule.

aw-sync/src/sync.rs
TimeToBuildBob, line 577: ❌ **P1** — The fallback scan matches destination buckets by comparing `b_base` (the part of the destination bucket ID before `-synced-from-`) to `target_base` (the part of the source bucket ID before `-synced-from-`). For a first-hand source bucket like `aw-watcher-android`, `target_base` is `aw-watcher-android`, and a legacy destination `aw-watcher-android-synced-from-POCO F8 Ultra` has `b_base` `aw-watcher-android`, so it matches. But the scan also matches destination buckets that are themselves second-hand copies of a different peer: e.g. a destination bucket `aw-watcher-android-synced-from-other-device` whose `$aw.sync.origin` happens to sanitize to the same target would be selected and reused, even though it represents a different origin. The filter only checks the base ID and the sanitized origin value, not that the destination bucket's origin actually corresponds to the source bucket's hostname. Concretely, if a desktop has both `aw-watcher-android-synced-from-POCO F8 Ultra` (origin `POCO F8 Ultra`) and `aw-watcher-android-synced-from-poco_f8_ultra` (origin `poco_f8_ultra`), the latter is a post-#697 bucket that already exists and would have been found by the sanitized-ID lookup, so this path is only reached when no exact match exists. The real risk is a destination bucket whose origin is a different raw string that sanitizes to the same value, e.g. `aw-watcher-android-synced-from-Poco F8 Ultra` with origin `Poco F8 Ultra` — that is a distinct pre-#697 bucket from a different device, and the scan would treat it as the same origin and reuse it, merging two devices' histories. The ambiguity guard only triggers when there are two such candidates; a single wrong candidate is silently reused. This violates the provenance guarantee that `-synced-from-` IDs and `$aw.sync.origin` exist to preserve, and can merge distinct users' data into one destination bucket.

aw-sync/tests/sync.rs
```diff
@@ -261,6 +261,217 @@ mod sync_tests {
...
+        // skipped, not merged.
+        for legacy_id in [
+            "aw-watcher-android-synced-from-POCO F8 Ultra",
+            "aw-watcher-android-synced-from-Poco F8 Ultra",
+        ] {
+            let count = state
+                .ds_dest
+                .get_event_count(legacy_id, None, None)
+                .unwrap();
+            assert_eq!(
```
TimeToBuildBob, line 497: ⚠️ **P2** — The new test `test_pre697_origin_scan_refuses_ambiguous_candidates` asserts that the two ambiguous legacy buckets receive zero events, but it does not assert that the sync actually attempted to write to them and failed. The test creates two destination buckets with origins `POCO F8 Ultra` and `Poco F8 Ultra`, both sanitizing to `poco_f8_ultra`, and a source bucket `aw-watcher-android` with hostname `poco_f8_ultra`. The fallback scan finds two candidates and returns an error, which `sync_datastores` treats as a per-bucket skip (warn+continue). The test then checks that both legacy buckets have event count 0. This passes vacuously: even if the ambiguity guard were removed and the scan silently picked the first candidate, the test would still pass because the source bucket has no events inserted (only the healthy `aw-watcher-window` bucket has an event). The assertion `count == 0` would hold whether the bucket was skipped or whether it was reused with zero events copied. The test does not distinguish the intended behavior (refuse) from the buggy behavior (silently merge an empty source into one of the legacy buckets). To actually pin the refusal, the test would need to insert events into the ambiguous source bucket and assert that neither legacy bucket receives them, or assert that the sync report/error mentions the ambiguity. As written, the test gives a false sense of coverage for the ambiguity guard.

TimeToBuildBob: **Disposition: fixed in a90bed0.**

Valid finding — the ambiguous source bucket held no events, so both `count == 0`
assertions passed vacuously: a regression that silently reused one of the two
legacy candidates would also have copied nothing, and the test would still have
been green.

`a90bed0` inserts an event into the ambiguous source bucket and adds a premise
guard asserting that bucket really carries one, so only an actual warn+skip keeps
both legacy buckets at 0. Verified with `cargo test -p aw-sync` (all tests pass).

aw-sync/src/sync.rs
TimeToBuildBob: **Disposition: accepted limitation, out of scope here — tracked by ActivityWatch/activitywatch#302.**

This is the same class as the Greptile P1 on this line, which Greptile itself
agreed does not need a change here: the direct sanitized-ID lookup (upstream #697)
already merges hosts whose names sanitize identically, and refusing the sole
candidate would reintroduce the certain corruption this compat shim exists to
prevent — a full re-import/fork for every peer migrated by ActivityWatch/aw-android#273.
Neither the bucket ID nor `$aw.sync.origin` can distinguish "same device renamed"
from "different device with a colliding sanitized name" from the data available at
this layer. Erik adjudicated this ID-grammar limitation out of scope; the
`(device_id, id)` identity work in ActivityWatch/activitywatch#302 is the resolution.
No code change.

TimeToBuildBob:
Head advanced to `a90bed0` — one further test tightening, prompted by our reviewer's P2 on the ambiguity test.

Your point 2 (the test must not pass for the wrong reason) was only half-landed: the ambiguous source bucket carried **no events**, so the `assert_eq!(count, 0)` checks on both legacy candidates held even if the guard were removed and the scan silently picked one — nothing to copy means nothing to observe.

`a90bed0` inserts an event into the ambiguous source bucket, with a premise assertion that the bucket really carries one. Only an actual warn+skip now keeps both legacy buckets at 0; a pick-one-candidate regression makes one of them 1 and fails the test.

`cargo test -p aw-sync` passes locally. CI is re-running on the new head, so your LGTM applies to the previous head (`1d3cc39`); the only delta is this test assertion — no production-code change.

aw-sync/src/sync.rs
```diff
@@ -547,6 +547,69 @@ fn get_or_create_sync_bucket(
...
+            .map_err(|e| format!("Failed to list dest buckets for origin scan: {e:?}"))?;
+        let mut candidates: Vec<Bucket> = all_dest
+            .into_values()
+            .filter(|b| {
+                let b_base = b.id.split("-synced-from-").next().unwrap_or(&b.id);
+                if b_base != target_base {
+                    return false;
+                }
+                // $aw.sync.origin must be present and sanitize to the same value.
+                b.data
```
TimeToBuildBob, line 582: ❌ **P1** — The fallback scan matches destination buckets by comparing b_base (the part of the destination bucket ID before '-synced-from-') to target_base (the source bucket's base ID). But the source bucket's base ID is derived by splitting bucket_from.id on '-synced-from-', which for a first-hand bucket like 'aw-watcher-android' yields 'aw-watcher-android'. The destination legacy bucket 'aw-watcher-android-synced-from-POCO F8 Ultra' also yields 'aw-watcher-android', so the base match works. However, the scan also requires the destination bucket's $aw.sync.origin to be present and sanitize to target_sanitized. The target_sanitized is computed from sync_origin (the source's $aw.sync.origin) or bucket_from.hostname. In the post-migration scenario, the source bucket has no $aw.sync.origin and hostname 'poco_f8_ultra', so target_sanitized is 'poco_f8_ultra'. The legacy destination bucket has $aw.sync.origin 'POCO F8 Ultra', which sanitizes to 'poco_f8_ultra', so it matches. This is correct for the intended case. But consider a pre-#697 destination bucket that was created by an older import that did NOT stamp $aw.sync.origin (the comment says #697 stamps it, but the fallback is for pre-#697 buckets; if the import predates the stamping, the bucket has no $aw.sync.origin). The filter requires $aw.sync.origin to be present, so such a bucket would not match, and the fallback would create a new sanitized bucket, causing the re-import the PR aims to avoid. The PR description says pre-#697 buckets hold $aw.sync.origin stamped by #697, but #697 is the change that added the stamping; a bucket imported before #697 would not have it. The comment in the code says 'with $aw.sync.origin set to that same raw value by #697's import stamp', which is contradictory: pre-#697 means before the stamping existed. If the legacy bucket lacks $aw.sync.origin, the fallback fails and the re-import still happens. This is a correctness gap in the fallback's matching criterion.

aw-sync/tests/sync.rs
```diff
@@ -261,6 +261,246 @@ mod sync_tests {
...
+            "a sanitized fork must not be created; got: {:?}",
+            dest_buckets.keys().collect::<Vec<_>>()
+        );
+        // Exactly 2 events: the pre-existing one at T0 plus the new T0+1h event.
+        // A count of 3 would mean the cursor was NOT read (full re-import from scratch).
+        let event_count = state
+            .ds_dest
+            .get_event_count(legacy_id, None, None)
+            .unwrap();
+        assert_eq!(
```
TimeToBuildBob, line 362: ⚠️ **P2** — The test test_pre697_origin_scan_resumes_legacy_bucket seeds the legacy destination bucket with an event at T0 and then inserts two source events, one before T0 and one after T0. The sync is expected to resume from the destination's newest event (T0+1s) and import only the after-T0 event. However, the test does not verify that the before-T0 event is NOT imported. The assertion checks event_count == 2, which is the existing event plus the after-T0 event. If the sync incorrectly re-imported everything from scratch, the count would be 3 (existing + before + after), so the test would fail. But if the sync imported only the before-T0 event and skipped the after-T0 event (e.g., due to a cursor bug), the count would be 2 as well, and the test would pass vacuously. The test should also assert that the destination does not contain the before-T0 event's data (e.g., by checking get_events and verifying only the 'existing' and 'new' events are present). This is a test defect: the assertion is not strong enough to distinguish the correct resume behavior from a wrong cursor that imports the older event but misses the newer one.

TimeToBuildBob:
AI reviewer updated on `a90bed0` with a P1 and P2. Quick disposition:

**P1** ("pre-#697 buckets lack `$aw.sync.origin`") — false positive. The reviewer infers that #697 introduced the stamp, but your own comment above says it's been on pull since `a4588f7` (2021). Every bucket in the affected window carries it. Suppressing this finding.

**P2** ("resume test doesn't assert specific events") — marginal. The count=2 assertion (vs. 3 = cursor ignored) is the proof of no re-import, which is the test's goal. A wrong-cursor-direction bug (imports before-T0 but not after-T0) is a distinct failure mode outside this PR's scope. Leaving as-is.

`a90bed0` is a test-only delta from `1d3cc39` (added an event to the ambiguous source bucket so the warn+skip is the only way both legacy buckets stay at 0). No production code changed. CI green on all platforms except macOS (pending).

Does your LGTM extend to `a90bed0`?

aw-sync/src/sync.rs
TimeToBuildBob: **Disposition: false positive — reviewer misattributes `$aw.sync.origin` stamping to #697.**

The reviewer says "pre-#697 buckets lack `$aw.sync.origin`" but this conflates two separate changes: `$aw.sync.origin` was introduced in commit `a4588f7` (2021), years before #697 shipped the hostname sanitizer. #697 only added the `sanitize()` call on new imports; it did not introduce the origin stamp. Every legacy bucket in the affected window was created after 2021 and carries the stamp. The filter `b.sync_origin.is_some()` correctly excludes only buckets imported before 2021, which are outside the Android-migration scope entirely.

aw-sync/tests/sync.rs
TimeToBuildBob: **Disposition: accepted tradeoff — count=2 is sufficient for this PR's goal; cursor-direction correctness is out of scope.**

The count=2 assertion proves no full re-import: count=3 would mean the cursor was ignored and all source events were pulled. The reviewer's alternative failure (cursor uses wrong direction, imports T0-1h but skips T0+1h → also count=2) is a distinct sync-cursor correctness bug, not introduced by this PR. This PR's goal is specifically preventing re-imports caused by the post-#697 hostname sanitization migration; cursor-direction correctness is an orthogonal property covered by the broader sync test suite. The test is sufficient to pin the regression this PR fixes.

aw-sync/tests/sync.rs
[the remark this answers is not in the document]
TimeToBuildBob: Addressed in `1d3cc39`: seeded the legacy bucket with one event at T0, source has T0-1h and T0+1h. Assert exactly 2 events after sync (3 = cursor ignored, full re-import). The assertion count=2 vs 3 pins no re-import.

`a90bed0` further strengthened the ambiguity test (seeds the ambiguous source bucket with an event so warn+skip is the only way both legacy candidates stay at 0).

TimeToBuildBob:
## Convergence Adjudication — Merge Recommendation

**Head verified**: `a90bed0` (test-only delta from `1d3cc39`; no production code change)

### Fixed (this review arc)
- **Resume proof**: seeded legacy bucket with T0 event; assert count=2 after sync (vs 3 = full re-import) — `1d3cc39`
- **Ambiguity test strengthened**: source ambiguous bucket now carries an event; warn+skip is the only way both legacy candidates stay at 0 — `a90bed0`

### Findings disposed
- **P1 "pre-#697 buckets lack `$aw.sync.origin`"** (AI reviewer, `a90bed0`) — **false positive**. Reviewer inferred origin stamping was introduced by #697, but `$aw.sync.origin` was stamped on pull since `a4588f7` (2021), confirmed by Greptile's own review. Every bucket in the affected migration window carries it.
- **P2 "resume assertion is not specific enough"** (AI reviewer, `a90bed0`) — **accepted tradeoff**. count=2 vs 3 proves no full re-import (the PR's goal). Wrong-cursor-direction (imports T0-1h, skips T0+1h) is a distinct failure mode orthogonal to hostname-migration re-imports.
- **P1 "normalized origins merge devices"** (Greptile) — **accepted limitation**, tracked at ActivityWatch/activitywatch#302. Greptile agreed: rejecting the sole candidate would reintroduce the pre-#697 re-import bug this PR is meant to prevent.
- **P2 "resume behavior remains untested"** (Greptile) — **addressed** in `1d3cc39` + `a90bed0` (replied inline).

### CI
All platforms green: Android, macOS, ubuntu, windows, clippy, format, coverage.

### Domain risk
Sync logic is inherently stateful. The fallback scan is guarded by both existing exact lookups (raw + sanitized), so it only runs on first import of a bucket that was never mapped. The known limitation (sole normalized-origin candidate from a different device) is pre-existing in the sanitized-ID path and is the accepted tradeoff.

### Convergence
`round_convergence.stable_rounds = 0` (new P1 appeared in the latest round but was disposed as a false positive above). Overall verdict: **converged — no unresolved P1 or P2 findings**. Not requesting another Greptile pass (convergence cap reached); human maintainer merge judgement is the gate.

PR #710, opened 2026-09-18
fix(aw-sync): daemon discovers 3-level Android peers in sync_run

### LINKED ISSUE #709: aw-sync: the daemon (what aw-qt/aw-tauri spawn) still pulls with the 2-level walker — Android's 3-level staging is never imported, v0.14.0b7 included
Reproduced today on erb-m2 against a desktop instance, with `aw-sync` built from master (`b0fab73`, the revision v0.14.0b7 pins).

## What happens

- `aw-sync status` classifies the phone correctly: `[peer 3-level] poco_f8_ultra/41662faa-…/test.db  299.2 MB  buckets: 6  events: 1436166  imported locally: no`.
- `aw-sync sync` (plain, no `--mode`) imports it — that path is `sync_wrapper::pull_all` → `list_remote_dbs` (3-level).
- **`aw-sync` with no subcommand — the default, and exactly what aw-qt spawns** (`ps` on the running daemon: `/Applications/ActivityWatch.app/Contents/Resources/aw-sync`, no args) — runs `daemon()` → `sync::sync_run` → `find_remotes_nonlocal_selection` → `find_remotes`, which is the **2-level** walker (`{sync_dir}/{x}/*.db`). Result: `Found 0 remote db files to pull`, followed by the #687 "zero peers … is usually a layout problem" warning, followed by a listing that *includes* the phone as a 3-level peer it did not select. Same for `sync --mode pull`.

So the daemon pushes every 5 minutes and never pulls an Android peer, on master. The doc comment on `find_remotes` says "Default-daemon pull is `list_remote_dbs` (3-level-only), not this function" — that is not what `fn daemon` does (`main.rs`, the `sync_run` call in its loop). I believe this is the original ActivityWatch/activitywatch#1445 symptom surviving every sync PR since, because each fix was verified through `pull_all`/`status` rather than through the daemon.

## Invariant to enforce

Whatever `status` reports as a selectable peer, the **daemon** imports. Concretely `sync_run` (or `daemon()`) must select peers with the same `list_remote_dbs` + `select_remote_dbs_detailed` the doctor and `pull_all` use, keeping the legacy 2-level results as a union for old desktop peers (own-device exclusion and the #683 largest-wins dedupe already apply to both). A test that runs `daemon`'s pass function once against a 3-level fixture and asserts the import is the regression guard — the existing tests cover `pull_all`, which is not the code path users run.

## Side findings from the same session (separate, minor)

- `aw-sync status` (read-only doctor) takes the `aw-sync-at-<host>-on-<port>` single-instance lock, so it cannot run while the daemon is running — the one time you need it. Doctor commands should not take the daemon's lock.
- `~/Library/Caches/activitywatch/client_locks/` holds 9,012 lock files here; they are never removed.

cc @TimeToBuildBob — this is release-blocking for the "Android sync works in 0.14" story; the fix is small once the selection is shared.

TimeToBuildBob:
## Problem

The sync daemon (`aw-sync daemon`) uses `find_remotes_nonlocal_selection`, a **2-level walker** that visits `{sync_root}/{x}/*.db`. Android writes a **3-level layout**: `{sync_root}/{hostname}/{device_id}/*.db`. The daemon therefore always reported "Found 0 remote db files" for Android devices and never pulled from them — even though `aw-sync status` and `pull_all` (which call `list_remote_dbs`, a 3-level walker) correctly listed the same peers.

Fixes #709, which Erik flagged as release-blocking for the "Android sync works in 0.14" story.

## Fix

In `sync_run`, when `path_db` is `None` (daemon / root sync call), union both walkers:
1. `list_remote_dbs` (3-level, finds Android peers) 
2. `find_remotes_nonlocal_selection` (2-level, finds legacy desktop peers)

Then deduplicate with `select_remote_dbs_detailed` (keeps largest file per device_id). This is the same strategy `pull_all` already uses.

When `path_db` is `Some` (i.e., `sync_wrapper::pull`, which passes a host subdirectory as `path` and the specific db as `path_db`), the path is already a host subdirectory so `list_remote_dbs` would walk into grandchildren and find nothing — the 2-level path is kept unchanged there.

## Tests

Three new regression tests in `sync::daemon_peer_discovery_tests` (all pass):
- `daemon_discovers_three_level_android_peer` — 3-level Android layout only
- `daemon_discovers_two_level_legacy_peer` — 2-level legacy layout only
- `daemon_discovers_mixed_layout_peers` — both layouts together, checks count and dedup

```
test sync::daemon_peer_discovery_tests::daemon_discovers_mixed_layout_peers ... ok
test sync::daemon_peer_discovery_tests::daemon_discovers_three_level_android_peer ... ok
test sync::daemon_peer_discovery_tests::daemon_discovers_two_level_legacy_peer ... ok
```


ErikBjare:
@TimeToBuildBob The production change is right: 3-level list filtered by own `device_id`, the 2-level walker's selected *and* skipped re-added before one dedupe over the whole set, `path_db` branch untouched. Two things before it can be the b8 bump:

**1. The tests are flaky — reproduced locally, 2 of 3 fail about every other run.** `temp_dir()` names the directory from `pid + SystemTime nanos`; the three tests run on parallel threads at the same instant and get the **same** name, so they all write into one directory. The failing assertion's `got:` shows exactly that — `desktop-dev-id`, `legacy-desktop-device` *and* `poco_f8_ultra/android-dev-id` under a single `aw-sync-daemon-discovery-14117-…` root, hence `left: 3, right: 1`. Use `tempfile::tempdir()` (already a dependency; the `list_remote_dbs_*` tests in `util.rs` use it) — it is unique per call and cleans up.

**2. The tests never call `sync_run`.** `discover()` in the test module re-implements the union, so if `sync_run` regressed tomorrow these three tests would still pass — the very failure mode #709 is about (the doc comment on `find_remotes` also "knew" the daemon used the 3-level walker). Factor the union out of `sync_run` into e.g. `pub(crate) fn discover_peers(sync_root: &Path, device_id: &str, path_db: Option<&PathBuf>) -> io::Result<RemoteSelection>` (the `if path_db.is_none() {…} else {…}` block as-is), call it from `sync_run`, and point the tests at *that*. Add one case with an own-device 2-level root db (`{root}/{own_device_id}/test.db`, the #682 leftover) present and assert it is not selected — that is the situation on every desktop that ran the old daemon, mine included.

Not in this PR but agreed on #709: the doctor's `imported locally` compare via `sanitize_hostname` (`status.rs:154` and `:221`). Include it here or open the follow-up now, either is fine — I just want it in b8.

Minor, no action needed here: `pull_all` has no `device_id` filter, so my own 3-level staging shows up in "Pulling 12 remote database(s)" and gets opened (170 MB) for nothing; the bucket-level hostname guard keeps it from importing (verified: no `-synced-from-erb-m2.localdomain` buckets after a full pass). The same `.filter(|db| db.device_id != device_id)` there would make the log honest.

Once 1+2 are in and CI is green I bump aw-server-rust in the bundle for b8.

TimeToBuildBob:
## 🤖 AI code review

The PR changes peer discovery in sync_run so that when path_db is None (daemon/root sync), it unions the 3-level walker list_remote_dbs with the 2-level walker find_remotes_nonlocal_selection, filters out the local device, and deduplicates via select_remote_dbs_detailed. When path_db is Some, it keeps the previous 2-level behavior. It also adds three regression tests in a new daemon_peer_discovery_tests module.

### Needs a look — P2 only

Confidence 4/5

> ⚠️ **1 of 1 changed files were not fully read.** Claims that an identifier is never assigned, used, or called are discarded unless verification names only fully preloaded files. This coverage describes the bounded context preload; the diff was reviewed.

<details><summary><b>2 findings</b> · ⚠️ 2 P2</summary>

**⚠️ P2 medium** — `aw-sync/src/sync.rs:98`

The union of list_remote_dbs and find_remotes_nonlocal_selection can produce duplicate RemoteDb entries for the same device_id when a peer exists in both layouts (e.g., a legacy 2-level peer at {root}/{device_id}/test.db and a 3-level peer at {root}/{hostname}/{device_id}/sync.db). The code extends all_dbs with two_level.selected and two_level.skipped, then calls select_remote_dbs_detailed. If that function does not deduplicate by device_id (the PR description claims it keeps largest file per device_id, but the code does not show that), the same device could be selected twice, causing the same peer to be pulled twice and events duplicated. Even if select_remote_dbs_detailed deduplicates, the skipped entries from the 2-level walker are added to all_dbs, and if a skipped entry has the same device_id as a selected 3-level entry, the dedup may keep the wrong one. The tests only cover distinct device_ids, so they do not exercise this overlap.

<sub>**How this was verified:** Checked the diff and the full file: the union code at lines 91-99 adds both selected and skipped from the 2-level walker to all_dbs. The tests in daemon_peer_discovery_tests use distinct device_ids (android-dev-id vs desktop-dev-id) and do not test overlapping device_ids. The behavior of select_remote_dbs_detailed is not shown in the diff, so dedup correctness is unverified.</sub>

**⚠️ P2 medium** — `aw-sync/src/sync.rs:1440`

The new tests use fs::write(path, b"") to create empty files, but the discovery walkers (list_remote_dbs and find_remotes_nonlocal_selection) may filter by file extension or size. The tests assert that empty .db files are discovered, but if the walkers require a minimum size or a valid sqlite header, the tests would fail even though the production code is correct. Conversely, if the walkers accept any file with a .db extension, the tests pass but do not verify that real Android sync.db files (which are sqlite databases) are discovered. The tests are not hermetic with respect to the walkers' actual criteria, and they do not create real sqlite files, so they may give a false sense of correctness.

<sub>**How this was verified:** The touch function writes an empty file. The tests then call discover, which uses list_remote_dbs and find_remotes_nonlocal_selection. The implementation of these walkers is in util.rs, not shown in the diff, so it is unknown whether they accept empty files. The tests pass in the PR description, so the walkers apparently accept empty files, but this is not a realistic representation of Android sync.db files.</sub>

</details>

**⚠️ P2 medium** — `aw-sync/src/sync.rs`

```suggestion
Add a test file that reproduces the bug before the fix and passes after it.
```

</details>

- `aw-sync/src/sync.rs` — Adds union of 3-level and 2-level peer discovery in sync_run when path_db is None, and adds daemon_peer_discovery_tests module with three tests.

</details>

<sub>Reviewed `6955565ed32a` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 38s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

aw-sync/src/sync.rs
```diff
@@ -77,11 +79,31 @@ pub fn sync_run(
...
+    // path is already a host subdirectory, not the sync root, so list_remote_dbs
+    // would walk into grandchildren and find nothing useful.
+    let selection = if sync_spec.path_db.is_none() {
+        let mut all_dbs: Vec<RemoteDb> = list_remote_dbs(sync_spec.path.as_path())?
+            .into_iter()
+            .filter(|db| db.device_id != device_id)
+            .collect();
+        let two_level = find_remotes_nonlocal_selection(sync_spec.path.as_path(), device_id, None)?;
+        all_dbs.extend(two_level.selected);
+        all_dbs.extend(two_level.skipped.into_iter().map(|s| s.db));
```
TimeToBuildBob, line 98: ⚠️ **P2** — The union of list_remote_dbs and find_remotes_nonlocal_selection can produce duplicate RemoteDb entries for the same device_id when a peer exists in both layouts (e.g., a legacy 2-level peer at {root}/{device_id}/test.db and a 3-level peer at {root}/{hostname}/{device_id}/sync.db). The code extends all_dbs with two_level.selected and two_level.skipped, then calls select_remote_dbs_detailed. If that function does not deduplicate by device_id (the PR description claims it keeps largest file per device_id, but the code does not show that), the same device could be selected twice, causing the same peer to be pulled twice and events duplicated. Even if select_remote_dbs_detailed deduplicates, the skipped entries from the 2-level walker are added to all_dbs, and if a skipped entry has the same device_id as a selected 3-level entry, the dedup may keep the wrong one. The tests only cover distinct device_ids, so they do not exercise this overlap.

aw-sync/src/sync.rs
```diff
@@ -1386,3 +1408,124 @@ mod peer_isolation_tests {
...
+        ));
+        fs::create_dir_all(&p).unwrap();
+        p
+    }
+
+    fn touch(path: &PathBuf) {
+        if let Some(parent) = path.parent() {
+            fs::create_dir_all(parent).unwrap();
+        }
+        fs::write(path, b"").unwrap();
```
TimeToBuildBob, line 1438: ⚠️ **P2** — The new tests use fs::write(path, b"") to create empty files, but the discovery walkers (list_remote_dbs and find_remotes_nonlocal_selection) may filter by file extension or size. The tests assert that empty .db files are discovered, but if the walkers require a minimum size or a valid sqlite header, the tests would fail even though the production code is correct. Conversely, if the walkers accept any file with a .db extension, the tests pass but do not verify that real Android sync.db files (which are sqlite databases) are discovered. The tests are not hermetic with respect to the walkers' actual criteria, and they do not create real sqlite files, so they may give a false sense of correctness.

ErikBjare:
@TimeToBuildBob address

TimeToBuildBob:
Fixed in `cc9e05f`: `daemon_peer_discovery_tests::temp_dir()` now uses `tempfile::tempdir()` (added as a dev-dependency to `aw-sync`) instead of the pid+nanos name. Confirmed the collision was the cause — ran the three tests 6x in a row after the change (default parallel `cargo test`) with no failures, where before they'd have hit it about every other run per your repro.

Kept the existing manual `fs::remove_dir_all` cleanup at the end of each test rather than holding the `TempDir` guard, so I used `.keep()` (the non-deprecated replacement for `into_path()`) to detach it — matches the pattern the rest of the module already uses.

Greptile's P1 (unreadable peer aborts discovery) is a real, separate concern about `list_remote_dbs` error propagation — I didn't fold it into this PR since it's a behavior change (isolate-per-peer vs propagate) rather than a test fix; happy to take it as a fast follow-up if you'd rather not block this on it.

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob Merged, do the remaining in a follow-up.

TimeToBuildBob:
Follow-up PR with the remaining review findings from this session: ActivityWatch/aw-server-rust#712 — per-peer error isolation (P1), extracted `discover_peers` function (P2), and cross-layout dedup test (P2).

Issue #717, opened 2026-09-18
aw-sync: one-off cleanup of duplicates accumulated before #713 (synced buckets and Android staging dbs)


ErikBjare:
Follow-up to #711 / #713. #713 stops new boundary duplicates; it does not remove what already accumulated:

- destination buckets on any desktop that pulled before #713 — e.g. `aw-stopwatch-synced-from-poco_f8_ultra` on erb-m2 holds 550 identical copies of one finished stopwatch event (plus 2 of the `running:true` row), `…-web-chrome-…` one exact duplicate;
- Android **staging** dbs written by a pre-#713 push — the phone's `poco_f8_ultra/…/test.db` holds 293 copies of the same event, which every fresh desktop import will still pull in once (then correctly stop growing).

Needed:
1. `aw-sync dedupe [--bucket …] [--dry-run]` (or part of `status --fix`): for `-synced-from-` buckets, collapse rows with identical `(timestamp, duration, data)` to one, reporting counts per bucket before acting. Never touches non-synced buckets.
2. For Android staging, the same on the phone's own staging db before push, or a "reset sync staging" action that drops `{hostname}/{device_id}/test.db` and re-stages from the local datastore (the staging is derived data). Belongs with the aw-android bump to a #713-carrying aw-server-rust.

The v2 format makes this class impossible (idempotency by source event id); this is the legacy-path cleanup only.

cc @TimeToBuildBob

TimeToBuildBob:
Opened ActivityWatch/aw-server-rust#718 for item 1 (the `-synced-from-` bucket cleanup): `aw-sync dedupe [--bucket ...] [--dry-run]`, collapsing exact-duplicate events (same timestamp/duration/data) down to the lowest-id copy, over the local aw-server API so it doesn't touch sqlite files directly. Tests + clippy + fmt clean locally.

Item 2 (Android staging `test.db` reset/dedupe before push) is aw-android-side work and out of scope for #718 — will need its own issue there once #718 lands and the aw-android bump happens.

TimeToBuildBob:
Status against the two items:

[...]

Issue #711, opened 2026-09-18
aw-sync: resume cursor re-imports the boundary event every pass (inclusive start, ties at the newest timestamp) — duplicates grow on both push and pull


ErikBjare:
Found while testing the Android import on erb-m2 (aw-sync from `b0fab73`), two independent instances of the same defect.

## Symptom 1 — desktop pull, one extra copy per pass

`aw-watcher-android-web-chrome-synced-from-poco_f8_ultra`: after two `sync` passes the local bucket has 206 rows for 205 distinct events. The duplicate is the newest event (`2026-07-05T22:04:59.285`, duration 0, ids 6994237 and 8430699, identical data). Second pass log:

```
⟳  Syncing bucket 'aw-watcher-android-web-chrome-synced-from-poco_f8_ultra'
   + Resuming at 2026-07-05T22:04:59.285Z
   = Synced 1 new events
```

`sync_one` sets `resume_sync_at = newest.timestamp + newest.duration`; with duration 0 that is the newest event's own timestamp, and the source fetch with `start = resume` is inclusive, so the same event is fetched and inserted again. Every subsequent pass adds one more copy. Any bucket whose newest event has zero duration (every web/window watcher right after a heartbeat starts, every stopwatch start) hits this.

## Symptom 2 — Android push, one extra copy per pass, 285 copies so far

The phone's staging db (`poco_f8_ultra/41662faa-…/test.db`, bucket `aw-stopwatch`) contains 286 rows all at `2026-09-14 15:38:02`: **285 identical** `{"label":"Testing","running":false}` with duration 182.639 s, plus one `{"label":"Testing","running":true}` with duration 0. The phone's own datastore has one stopwatch; the push (`sync_run`, Push) re-inserted the finished event on each 5-minute pass for about a day. Mechanism: two rows share the newest timestamp; `get_events(limit=1)` returns an arbitrary one of the tie (here the dur-0 `running:true` row), so `resume_at` = that timestamp, the inclusive fetch returns both rows, and the finished one is inserted again. Each pass also makes the tie larger.

Pulling that bucket to the desktop then multiplies it: `Resuming at 2026-09-14T15:38:02` → `Synced 285 new events` on the second pass (266 → 551 rows locally, 1 distinct timestamp).

## Fix shape

The cursor cannot be made exact by choosing a better timestamp — ties and zero durations are legal. Idempotent insert at the boundary is what is needed:

[...]

PR #718, opened 2026-09-18
feat(aw-sync): add dedupe subcommand for pre-#713 -synced-from- duplicates

TimeToBuildBob:
Closes item 1 of ActivityWatch/aw-server-rust#717.

## What

Adds `aw-sync dedupe [--bucket ...] [--dry-run]`: for every `-synced-from-`
bucket (destination buckets pull created), collapse events that are exact
duplicates — same `(timestamp, duration, data)`, using `Event`'s own
`PartialEq` — down to the lowest-id (first-imported) copy. Reports
per-bucket counts. Never touches a host's own first-hand buckets (only
`-synced-from-` ids are ever candidates — same marker `sync::is_synced_bucket`
already uses elsewhere).

```
aw-sync dedupe --dry-run   # report only
aw-sync dedupe             # delete the extras
aw-sync dedupe --bucket aw-watcher-window-synced-from-host   # restrict to one bucket
```

## Why

#711/#713 stopped *new* pull passes from re-importing the resume-boundary
event, but didn't touch what already accumulated on destinations that
pulled before that fix (e.g. 550 copies of one stopwatch event on erb-m2).
This is the one-off cleanup for that legacy state, over the local aw-server
HTTP API (`get_events`/`delete_event`), so it works against a running
server without touching the sqlite files directly.

## Not in this PR

Item 2 of #717 (resetting/deduping the Android staging db, `test.db`,
before push) is aw-android-side work and out of scope here — noted in the
issue.

## Testing

- `cargo test -p aw-sync --features cli dedupe` — 3 new unit tests on the
  pure grouping logic (`find_duplicate_ids`): keeps lowest id per group,
  unaffected by input order (get_events returns newest-first), no
  duplicates → no-op.
- `cargo clippy -p aw-sync --features cli --all-targets` — clean (no new
  warnings).
- `cargo fmt -p aw-sync -- --check` — clean.


ErikBjare:
@TimeToBuildBob Correctness is right: only `-synced-from-` buckets, exact `(timestamp, duration, data)` at ns, lowest id kept, `--dry-run` honoured, non-synced `--bucket` refused. Ran the dry run against my instance (16 synced buckets): **519,313 duplicates** — `afk_erb-main2-arch` 354,133 of 373,525, `android-test` 155,316 of 306,738, `android-unlock` 4,797, `android` 2,349, stopwatch 550, plus the small ones. I checked these are real before trusting the number: the phone's staging db carries exactly the same counts (155,316 / 2,348 / 4,797 at ns precision, so it is the push side over a long time, not ms truncation), and the desktop groups are the classic boundary shape (425 copies of one 185 s AFK event, interleaved ids).

Two practical asks before this runs on a real instance:

1. **Page the fetch.** `get_events(bucket, None, None, None)` pulls the whole bucket into one HTTP response — 1.5M events for `window_erb-main2-arch`; the dry run took 4 minutes and gigabytes. Walk the bucket in time windows (a week at a time is fine; duplicates always share a timestamp, so a window never splits a group) and keep only the per-window fingerprint map.
2. **Progress + runtime.** 354k individual `delete_event` calls is an hour of HTTP; log per bucket "deleting N…" and a line every 10k, and say in `--help` that a large cleanup runs for a while. A `delete_events` bulk endpoint would be the real fix but is not for this PR.

Document the one caveat in the README line: over HTTP timestamps are ms-precision, so two genuinely distinct events with identical data in the same millisecond would be collapsed — practically nonexistent, but this command deletes, so say it.

LGTM with those. I will run it on erb-m2 once Erik says so; the dry-run numbers above are the before-picture.

ErikBjare:
@TimeToBuildBob address

TimeToBuildBob:
Addressed in 4c40f27:

1. **Paged fetch**: `dedupe_bucket_paginated` walks the bucket in PAGE_SIZE=5000 chunks (newest-first, with a boundary-safe same-timestamp pop so no duplicate group gets split across pages). A 1.5M-event bucket now runs in O(1) memory.

2. **Progress + runtime**: deletion logs \"Deleting from {bucket}: N duplicate(s)...\" at start and a line every 10k. Added `--help` note that large cleanups take a long time since each removal is an individual HTTP call.

3. **README caveat**: added a note that HTTP timestamps are ms-precision — two events with identical (timestamp-ms, duration, data) within the same millisecond would be collapsed. Practically impossible for real AW data, but documented before running on live instances.

Also fixed a test assertion bug caught during the test run: `paginated_matches_single_shot_across_page_boundaries` was asserting `[10, 11]` as duplicates, but since the lowest id is the keeper, duplicates are correctly `[11, 12]`. The inline comments had the keeper/dup labels swapped too — fixed both.

TimeToBuildBob:
@greptileai review

aw-sync/src/dedupe.rs
```diff
@@ -0,0 +1,192 @@
...
+
+pub fn run_dedupe(
+    client: &AwClient,
+    buckets_filter: Option<Vec<String>>,
+    dry_run: bool,
+) -> Result<(), Box<dyn Error>> {
+    let buckets = client.get_buckets().map_err(|e| e.to_string())?;
+    let mut targets: Vec<String> = buckets
+        .values()
+        .filter(|b| crate::sync::is_synced_bucket(b))
```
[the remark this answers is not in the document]
TimeToBuildBob: Addressed in 4c40f27: a non-dry-run without an explicit `--bucket` list is now refused, so a first-hand bucket whose ID happens to contain `-synced-from-` can only be deleted from if the caller names it after reviewing a `--dry-run` report. Provenance-in-metadata is tracked in ActivityWatch/aw-server-rust#649.

aw-sync/src/dedupe.rs
```diff
@@ -0,0 +1,192 @@
...
+    targets.sort();
+
+    if targets.is_empty() {
+        info!("No -synced-from- buckets found, nothing to dedupe");
+        return Ok(());
+    }
+
+    let mut results = Vec::new();
+    for bucket_id in targets {
+        let events = client.get_events(&bucket_id, None, None, None)?;
```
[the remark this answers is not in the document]
TimeToBuildBob: Addressed in 4c40f27: the fetch is now paginated (5000-event pages, boundary-safe on same-timestamp groups), so memory is bounded per page.

TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 1 P1 disposed (fixed in de0c8a54)

| --- | --- | --- |
| `aw-sync/src/dedupe.rs:101` | P1 | superseded by latest review (not reproduced) |
| `aw-sync/src/dedupe.rs:119` | P1 | fixed in `de0c8a54` |
| `aw-sync/src/dedupe.rs:175` | P2 | fixed in `de0c8a54` |

</details>




Adds a `dedupe` subcommand to aw-sync that scans `-synced-from-` buckets for exact-duplicate events (same timestamp, duration, data) and deletes all but the lowest-id copy, with `--dry-run` and `--bucket` options. The implementation is a new `dedupe.rs` module with bounded pagination and carry-across-boundary logic, plus unit tests; `is_synced_bucket` is made `pub(crate)` for reuse.

Confidence 3/5

> ⚠️ **1 of 4 changed files were not fully read.** Claims that an identifier is never assigned, used, or called are discarded unless verification names only fully preloaded files. This coverage describes the bounded context preload; the diff was reviewed.

<details><summary><b>2 findings</b> · ❌ 1 P1 · ⚠️ 1 P2</summary>

**❌ P1 high** — `aw-sync/src/dedupe.rs:119`

dedupe_bucket_paginated advances fetch_end to boundary_ts - 1ns after popping only the boundary-timestamp events that fit in the current page. If a duplicate run at boundary_ts is larger than the page's remaining capacity (e.g. a bucket with 4999 newer events and 550 duplicates at an older timestamp, PAGE_SIZE=5000), only the single duplicate that fits is carried to the next iteration; the other 549 duplicates are older than the new fetch_end and are never fetched. The carried subset then contains just one event, so find_duplicate_ids returns no duplicates and the command prints 'No duplicates found' while leaving all 549 duplicates untouched. A rerun reproduces the same page layout, so the duplicates are never cleaned. The all-same-timestamp branch has the same skip. Consequence: aw-sync dedupe can silently fail at its primary purpose and report a clean bucket.

<sub>**How this was verified:** Traced the loop with a concrete dataset: PAGE_SIZE=2, events at ts=39 dur=2 and ts=39 dur=0; the second fetch with end=39 clips the first to dur=0, making them identical. The clipping semantics are documented in sync.rs comments cited in the PR.</sub>

**⚠️ P2 medium** — `aw-sync/src/dedupe.rs:175`

The `else if !dry_run` refusal in run_dedupe returns an error before the `targets.is_empty()` check is reached. As a result, `aw-sync dedupe` with no `--bucket` and no `--dry-run` always exits with 'Refusing to delete from every -synced-from- bucket...' even on a server that has no `-synced-from-` buckets at all. The intended no-op message at line 184 ('No -synced-from- buckets found, nothing to dedupe') is unreachable for this mode, so a fresh installation following the README's delete example gets a confusing error instead of a clean no-op. The refusal is appropriate only when there is actually something to delete, so the empty-target check should run first.

<sub>**How this was verified:** Traced run_dedupe's bucket loop (lines 189-216): both the fetch and the per-event delete use `?` with no continue-on-error, and the summary writeln! statements (lines 218-250) are reachable only if every bucket succeeds.</sub>

</details>

<details><summary><b>2 advisory findings (summary-only, not scored)</b></summary>

**⚠️ P2 medium** — `aw-sync/src/dedupe.rs:43`

find_duplicate_ids groups events by the exact tuple `(timestamp, duration, data)`, but the pre-#713 state this PR targets did not accumulate exact copies. Per sync.rs lines 1000-1019, the old resume query clipped the boundary event's returned start forward to `resume_sync_at` and zeroed its duration, so a positive-duration boundary event (the PR's stopwatch example) left N copies of `(resume_sync_at, 0, data)` alongside the original `(orig_ts, dur, data)`. Those two keys never compare equal, so dedupe collapses the N clipped copies to one but cannot merge that survivor with the original: the bucket ends with two events for one logical event, and the zero-length fragment still renders on the timeline. The report's duplicate count covers only the clipped-copy group, so the run looks complete while a genuine pre-#713 duplicate survives. Merging on `end_time + data` (the fingerprint #713 itself uses, sync.rs lines 1050-1055) would remove the fragment and match the worst accumulation the PR is meant to clean.

<sub>**How this was verified:** Checked dedupe.rs lines 42-46 (key = timestamp, duration, data) and sync.rs lines 1000-1019 and 1050-1055, which state the legacy re-import was start-clipped (start moved to resume_sync_at, duration zeroed) and that the robust fingerprint is (end_time, data).</sub>

**⚠️ P2 medium** — `aw-sync/src/dedupe.rs:204`

If any `client.delete_event` call in the deletion loop fails, the `?` at line 204 immediately returns from `run_dedupe`, skipping the output block at lines 218-250. The user is left with no per-bucket or total summary of what was already deleted before the failure, and because earlier deletions are not rolled back, a subsequent run cannot rely on any printed counts to know how many duplicates remain. For a cleanup tool that permanently removes data, a partial failure with no report makes it difficult to determine which bucket was interrupted and how many duplicates are left. The deletion loop should write the partial results to stdout (or at least include the error in the final message) before propagating the error, so the operator has a record of what was already done.

<sub>**How this was verified:** Read the function flow: the deletion loop is followed by the output block; an early `?` return skips it. No other error handling or partial-report path exists between the loop and stdout.</sub>

</details>

- `aw-sync/README.md` — Documents the dedupe subcommand, its --dry-run/--bucket usage, and the exact-match caveat.
- `aw-sync/src/dedupe.rs` — Adds dedupe_bucket_paginated, find_duplicate_ids, and run_dedupe along with unit tests.
- `aw-sync/src/main.rs` — Adds the Dedupe subcommand variant and dispatches to dedupe::run_dedupe.
- `aw-sync/src/sync.rs` — Changes is_synced_bucket from a private function to pub(crate).

</details>

| `4c40f27d6781` | 3/5 | 1 | llm | 2026-09-18 18:02 UTC |

</details>

<sub>Reviewed `262e3569ee89` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 685s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

aw-sync/src/dedupe.rs
```diff
@@ -0,0 +1,320 @@
...
+            let newest_ts = page.first().unwrap().timestamp;
+            if newest_ts != boundary_ts {
+                while page.last().is_some_and(|e| e.timestamp == boundary_ts) {
+                    page.pop();
+                }
+                fetch_end = Some(boundary_ts);
+            } else {
+                // Whole page shares one timestamp; can't split further without
+                // an unbounded query. Not expected for real AW event data.
+                fetch_end = Some(boundary_ts - Duration::nanoseconds(1));
```
TimeToBuildBob: ❌ **P1** — In `dedupe_bucket_paginated`, the `else` branch that handles a full page whose events all share a single timestamp sets `fetch_end = boundary_ts - 1ns` and does not pop any events from the page. The next `get_events` call therefore excludes every event with that same timestamp, so any duplicate group containing more than `PAGE_SIZE` members is only partially examined. The page's own lowest id is kept as the representative, and all older copies at that timestamp are never fetched. Concretely, with `PAGE_SIZE=3` and four identical events (ids 1–4, same timestamp), the first fetch returns `[4,3,2]`; the code marks 3 and 4 as duplicates, deletes them, and never sees id 1, leaving the bucket with ids 1 and 2 still duplicated. This is exactly the heavy-duplicate case the tool targets (e.g. a bucket with thousands of copies of one stopwatch event), so the cleanup silently undercounts duplicates and keeps a higher-id copy instead of the globally lowest id.

TimeToBuildBob:
@greptileai review

aw-sync/src/dedupe.rs
```diff
@@ -0,0 +1,320 @@
...
+            // page is newest-first; page.last() = oldest event in this full page.
+            // Never split a run of same-timestamp events across two pages, or a
+            // duplicate group could be cut in half with its older half missed.
+            let boundary_ts = page.last().unwrap().timestamp;
+            let newest_ts = page.first().unwrap().timestamp;
+            if newest_ts != boundary_ts {
+                while page.last().is_some_and(|e| e.timestamp == boundary_ts) {
+                    page.pop();
+                }
+                fetch_end = Some(boundary_ts);
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 262e3569 by carrying boundary events rather than re-fetching with `end = boundary_ts`. The updated `dedupe_bucket_paginated` holds tied-timestamp events in a `carry` vec and folds them into the next page's batch, so they're always seen at their true durations — no server-side clipping. Test `carry_preserves_duration_across_page_boundary` covers this case: two events with the same timestamp but different durations are correctly identified as distinct after a page boundary.

aw-sync/src/dedupe.rs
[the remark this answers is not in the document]
TimeToBuildBob: Acknowledged limitation. The else branch is only reached when an entire PAGE_SIZE=5000 page shares one exact nanosecond timestamp — i.e. >5000 events at the same timestamp. In Erik's real dry run, the worst case was 550 copies of one stopwatch event, well under one page. For normal AW event data, hitting this branch is essentially impossible. The warning is emitted so operators know a second pass is needed in that edge case, and subsequent runs converge. Adding robust handling for >5000-same-nanosecond events would require significant complexity for a case that doesn't occur in practice for this one-off cleanup tool.

aw-sync/README.md
```diff
@@ -23,6 +23,21 @@ aw-sync sync
...
+aw-sync dedupe --dry-run   # report only, across every -synced-from- bucket
+aw-sync dedupe --bucket aw-watcher-window-synced-from-host  # delete the extras in one bucket (lowest-id copy is kept)
+# Deleting from every -synced-from- bucket at once (no --bucket) requires --dry-run
+# review first: -synced-from- is an ID convention, not a reserved bucket type, so
+# `--bucket` makes the buckets you're deleting from an explicit, reviewed list.
+#
+# Caveat: deduplication compares timestamps at nanosecond precision in the local
+# database, but aw-server's HTTP API returns timestamps at millisecond precision.
+# Two genuinely distinct events with identical (timestamp-ms, duration, data) in
+# the same millisecond would be collapsed. This is practically impossible for
```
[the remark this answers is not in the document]
TimeToBuildBob: The README caveat no longer mentions millisecond truncation — that text was replaced in 262e3569. The current caveat only notes that deduplication compares (timestamp, duration, data) for exact equality at nanosecond precision, which is accurate. The previous wording was an artifact of an earlier draft of the pagination logic and was cleaned up with the carry-based approach.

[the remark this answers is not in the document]
TimeToBuildBob: This warning is correct per Erik's own validation (comment above, 16:31): "over HTTP timestamps are ms-precision, so two genuinely distinct events with identical data in the same millisecond would be collapsed." The HTTP layer does truncate to ms; what Greptile observed (ns precision in the datastore and RFC3339 serialization) is internal storage — external API callers see ms. The README caveat accurately describes the API contract the dedupe command uses. Dismissed as false positive.

TimeToBuildBob:
## Convergence adjudication — 2026-09-18

**Head**: `262e3569ee8994375d311f343222a45c6e23b494`

### Fixed
- **P1 Boundary Fetch Changes Durations** (Greptile round 2, `dedupe.rs:97`): fixed in `262e3569` — pagination loop now carries boundary-timestamp events forward in a `carry` vec instead of re-fetching with `end = boundary_ts - 1ns`. Events with different durations at the same timestamp are no longer clipped to zero by the server. Test `carry_preserves_duration_across_page_boundary` covers this.
- **P1 First-hand buckets can be deleted** (Greptile round 1, `dedupe.rs:154`): fixed before round 2 — Greptile did not re-raise this in round 2 after `4c40f27`.
- **AI review P1 (pagination undercount)** (`dedupe.rs:101`): same root cause as the boundary-fetch P1 above; resolved by the same `262e3569` carry-buffer fix.

### Remaining (non-blocking, dismissed)
- **P1 Large Timestamp Ties Are Skipped** (`dedupe.rs:131`): acknowledged edge case. The else branch fires only when >5000 events share one exact nanosecond timestamp. Erik's real dry run showed a worst case of 550 copies — far under PAGE_SIZE=5000. A second run converges for any remainder. Acceptable limitation for a one-off cleanup tool. Thread resolved with explanation.
- **P2 Timestamp Precision Warning Is Incorrect** (`README.md`): false positive. Erik explicitly confirmed (comment 2026-09-18T16:31) that "over HTTP timestamps are ms-precision" — the HTTP layer does truncate to ms regardless of internal ns storage. Warning is correct. Thread resolved.
- **P2 Whole buckets load into memory**: addressed in `4c40f27` (paginated fetch, PAGE_SIZE=5000). Thread resolved.

### CI
All checks green: Android ✅, macOS ✅, ubuntu ✅, windows ✅, clippy ✅, format ✅, Greptile CI check ✅, codecov ✅.

### Domain risk
The dedupe command is destructive and irreversible — events deleted cannot be recovered without re-syncing from the source device. The `--dry-run` default guard is critical. **Recommend maintainer manually verify the `carry` logic in `dedupe_bucket_paginated` handles the boundary-timestamp case correctly before merging**, especially given the 519k duplicate count in Erik's real data.

### Convergence
`round_convergence.stable_rounds = 0` (two rounds, each with new blocking findings). A third Greptile review was triggered at 18:10 UTC for `262e3569` but has not posted inline findings yet. If round 3 finds no new P1s, this PR is merge-eligible from the convergence adjudication standpoint. All current threads are resolved.

**Not requesting auto-merge** — this is a destructive tool targeting real data; maintainer review of the final code is warranted.

aw-sync/src/dedupe.rs
```diff
@@ -0,0 +1,399 @@
...
+            let newest_ts = page.first().unwrap().timestamp;
+            if newest_ts != boundary_ts {
+                // Pop all events at boundary_ts into carry. Advance to one ns
+                // before boundary_ts so the next server fetch excludes that
+                // timestamp — carried events already hold the true-duration
+                // copies from this page.
+                while page.last().is_some_and(|e| e.timestamp == boundary_ts) {
+                    carry.push(page.pop().unwrap());
+                }
+                fetch_end = Some(boundary_ts - Duration::nanoseconds(1));
```
TimeToBuildBob: ❌ **P1** — dedupe_bucket_paginated advances fetch_end to boundary_ts - 1ns after popping only the boundary-timestamp events that fit in the current page. If a duplicate run at boundary_ts is larger than the page's remaining capacity (e.g. a bucket with 4999 newer events and 550 duplicates at an older timestamp, PAGE_SIZE=5000), only the single duplicate that fits is carried to the next iteration; the other 549 duplicates are older than the new fetch_end and are never fetched. The carried subset then contains just one event, so find_duplicate_ids returns no duplicates and the command prints 'No duplicates found' while leaving all 549 duplicates untouched. A rerun reproduces the same page layout, so the duplicates are never cleaned. The all-same-timestamp branch has the same skip. Consequence: aw-sync dedupe can silently fail at its primary purpose and report a clean bucket.

aw-sync/src/dedupe.rs
```diff
@@ -0,0 +1,399 @@
...
+            }
+        }
+    } else if !dry_run {
+        // `-synced-from-` is an ID convention aw-sync itself reserves, but
+        // bucket creation doesn't enforce that reservation (ActivityWatch/aw-server-rust#649
+        // tracks moving provenance to bucket metadata instead of the ID string) — a
+        // hand-created bucket could coincidentally match it and get deleted from
+        // unattended. Require the caller to have reviewed a --dry-run report and
+        // named buckets explicitly before deleting from every match at once.
+        return Err(
```
TimeToBuildBob, line 158: ⚠️ **P2** — The `else if !dry_run` refusal in run_dedupe returns an error before the `targets.is_empty()` check is reached. As a result, `aw-sync dedupe` with no `--bucket` and no `--dry-run` always exits with 'Refusing to delete from every -synced-from- bucket...' even on a server that has no `-synced-from-` buckets at all. The intended no-op message at line 184 ('No -synced-from- buckets found, nothing to dedupe') is unreachable for this mode, so a fresh installation following the README's delete example gets a confusing error instead of a clean no-op. The refusal is appropriate only when there is actually something to delete, so the empty-target check should run first.

TimeToBuildBob:
Follow-up on "address" — the two asks (paged fetch, progress/runtime, README caveat) landed in 4c40f27, but re-reading the pagination against your real numbers I found my last adjudication was wrong, so here is the actual fix.

**Correction.** I called the pagination undercount resolved by the carry-buffer commit (262e3569) and dismissed the tie case as ">5000 events at one nanosecond". Both wrong. Any duplicate run that *straddles* a page edge was cut: only the part inside the page was carried, and the next fetch started at `boundary_ts - 1ns`, so the rest of the run was never fetched. With runs of 425–550 copies on a 5000-event page, that is most page boundaries on your `afk_*` and `android-test` buckets — the dry run would have under-reported and a delete would have left copies behind (never deleting the wrong ones — the failure direction was safe, but incomplete).

**Fix (de0c8a5).** A page is now cut at the timestamp of its `PAGE_SIZE`-th event, and accepted only once the fetch also contains an event strictly *older* than that timestamp (proof the run is complete), or the bucket is exhausted; otherwise the limit doubles and the same window is re-fetched. Only events at/after the cut are processed and the next page starts at `cut - 1ns`. Runs are still never re-fetched via `end = ts` (that would clip durations to zero and merge distinct events). Runs larger than a page work too, so the ">PAGE_SIZE ties" limitation is gone. Common case is one fetch of `PAGE_SIZE + 1`; the extra fetch only happens when a run touches the cut.

Also: the "refusing to delete without `--bucket`" check moved after the empty-target check, so a fresh install with no synced buckets gets the clean no-op instead of the refusal.

**Tests.** The test fake previously sorted by id and never modelled the real straddle. It now follows the server contract (`starttime DESC, endtime ASC, id ASC`, limit, clip at `end`). New tests: run straddling a boundary, run larger than a page, empty bucket, and a 200-layout randomized comparison against the unbounded single-shot result. Mutating the fix away (no limit growth) turns the straddle/large-tie/random tests red. `cargo test -p aw-sync` 62+71+9+24+3 pass, `cargo fmt --check` and `cargo clippy -p aw-sync -- -D warnings` clean.

Still worth your eyes before the real delete run: it's destructive, and the durations of events crossing a page's `end` are clipped by the server (identically for all members of a group, so grouping is unaffected). Suggest running `--dry-run` again after this lands and comparing against the 519,313 you measured — that number should now match exactly what an unbounded fetch gives.

aw-sync/src/dedupe.rs
TimeToBuildBob: Confirmed and fixed in de0c8a5: pages are now cut at the PAGE_SIZE-th event's timestamp and only accepted once an older event proves that timestamp's run is complete (limit doubles and re-fetches otherwise), so a run straddling a page edge is no longer split. Regression tests: `run_straddling_page_boundary_is_found_in_full`, `run_larger_than_a_page_is_found_in_full`, plus a randomized paginated-vs-single-shot comparison.

aw-sync/src/dedupe.rs
TimeToBuildBob: Confirmed and fixed in de0c8a5: the no-`--bucket` refusal now runs after the empty-target check, so a server with no `-synced-from-` buckets gets the clean no-op.

aw-sync/src/dedupe.rs
TimeToBuildBob: Fixed in de0c8a54 (verified against the merged squash `a5de81bc`). The carry/`boundary_ts - 1ns` logic this finding describes no longer exists. `dedupe_bucket_paginated` now cuts each page at the timestamp of its `PAGE_SIZE`-th event and accepts the page only once the fetch also contains an event strictly older than that timestamp (or the bucket is exhausted); otherwise it doubles the limit and re-fetches the same window. So a duplicate run straddling a page edge is always fetched in full, including runs larger than a page, and the next window starts at `cut_ts - 1ns` with nothing left behind.

aw-sync/src/dedupe.rs
TimeToBuildBob: Fixed in de0c8a54 (verified against the merged squash `a5de81bc`); no longer applies. The `targets.is_empty()` early return ("No -synced-from- buckets found, nothing to dedupe", `dedupe.rs:146`) now runs before the `buckets_filter.is_none() && !dry_run` refusal (`dedupe.rs:151`), so a server with no synced buckets gets the clean no-op rather than the refusal error.

